PR #19, opened 2019-02-28
add URL parameters for /metrics, make /metrics load faster

youngj:
/metrics was taking too long to load 3 days of data from graphql


hathix:
thanks! much better

PR #26, opened 2019-03-04
Refactored backend functions for getting stops and wait times

jtanquil:
`get_stops` can now either query graphql (with the same parameters as before) or can take raw json data from a previous query. Moved functions for computing the wait times to another file.


hathix:
Looks great

PR #36, opened 2019-03-10
Fetch route config from Nextbus public JSON API

youngj:
* fetch route config from Nextbus public JSON API directly instead of restbus.info which is not currently working
* cache route config locally to reduce number of queries to Nextbus API and improve performance
* update Dockerfile to use python:3.7.2-slim-stretch as base image (can now use python instead of python3)


hathix:
lgtm

PR #41, opened 2019-03-14
Try making Heroku deployment work (and without Yarn, either)

hathix:
I'm almost at making the Heroku deployment work (local deployment already works).


hathix:
Yarn deployment works!

PR #45, opened 2019-03-21
Use yarn, set up css grid

sravanti:
A few things in this PR: 
- Remove package-lock.json -- we should only be using yarn for package management. 
- Introduce CSS grid -- visually this still looks rough, but the pieces are in place to follow the mockup Samantha made.

Mockup:
Open Transit Wireframes

Current status:
Screen Shot 2019-03-20 at 8 42 53 PM

I also did some code cleanup, removing unnecessary divs and using Emotion for our CSS-in-JS lib. I also moved the big address object into its own constants file.

@hathix lmk who from front-end I should add to this review for a heads up. I'll also merge latest master to avoid merge conflicts.

PR #49, opened 2019-03-24
Connect frontend and backend and get everything working in Docker, Gitpod, and Heroku

youngj:
This PR updates the React frontend to actually fetch metrics from the Flask backend, and updates the build scripts for Docker, Gitpod, and Heroku so that everything works in the various supported development and production environments.

Running example here: https://opentransit-youngj.herokuapp.com/

### Updated the React frontend to fetch metrics and routes/directions/stops from the Flask backend

Updated the ControlPanel to fetch route, direction, and stop information for the dropdowns from the backend. Instead of using restbus.info, I added some endpoints on the backend so that the frontend and backend are both using the same data source. By avoiding using restbus.info, we avoid the risk of the entire app failing if restbus.info goes down or if Nextbus blocks its IP address (again?). It also reduces load on Nextbus's API since the Flask backend caches route configuration.

Shows an histogram of the minutes between each bus as an example of fetching data from the backend. This is intended to be a placeholder that could be replaced with the actual data we want to display.

The environment variable `REACT_APP_METRICS_BASE_URL` can be set for the React app to fetch metrics and route configuration from a different domain (which is necessary when running in development). The Flask app now supports CORS to make this possible.

### Combined Dockerfile for react/flask

When running in production, flask serves the static frontend files generated by `yarn build`. The easiest way to do this seems to be to generate one Docker container that has python/flask as well as node/react, which runs `yarn build` to build the static frontend files during the build process. 

When running in development via docker-compose, there are still two different containers for react and flask, but now both of them run from the same image (with different commands).

### Added message when viewing root URL of API server if production assets haven't been generated

When running in development via docker-compose or Gitpod, by default the production frontend assets haven't been built yet, so you would get a 404 error when accessing the root of the API server. http://localhost:5000 now shows a more helpful error explaining the situation.

(When running in development via docker-compose, the frontend assets generated by `yarn build` in the Dockerfile are not visible due to the `.:/app` volume mount.)

### Updated Heroku deploys to use Docker

By using Docker to deploy to Heroku instead of the old Procfile system, we can reduce the number of different environments and build processes that need to be supported. We just give Heroku the same Dockerfile used in development by docker-compose.  Since Heroku only allows serving one "web" container per app, it works well to have one docker container that serves both the production frontend and backend.

### Updated Gitpod scripts 

Gitpod scripts now support hot reloading for both frontend and backend, and use yarn instead of npm to match the Docker build process. Removed virtualenv since Gitpod already provides an isolated development environment. Removing virtualenv allows python scripts to be run directly from any Gitpod terminal.

### Updated Readme

Explained how to set up a local development environment with Docker, and removed explanation for how to set up a non-Docker local development environment (which people could still do, but probably no need to encourage that). 

Explained how anyone can deploy their local code to their own Heroku account.

Explained how to compute arrival times and other command line scripts from previous commits.


.gitpod.yml
```diff
@@ -1,15 +1,14 @@
...
-    virtualenv -p python3 venv &&
-    source venv/bin/activate &&
     pip install -r requirements.txt &&
-    FLASK_APP=metrics-api.py flask run --host 0.0.0.0
+    FLASK_DEBUG=1 FLASK_APP=metrics-api.py flask run --host=0.0.0.0
 - command: |
-    npm install --prefix frontend &&
-    npm run-script build
-
+    (cd frontend && yarn install) &&
```
hathix: Are we going with Yarn as our package manager, or using straight-up NPM? I think we've been debating this in recent weeks.

README.md
```diff
@@ -2,48 +2,105 @@
...
+If you need to install some new dependencies in the Docker images, you can rebuild them via `docker-compose build`.
+
 ### Option 2: Cloud IDE
 
-If you don't want to handle configuring your system, installing packages,
-wrangling Python versions, and stuff, just **[use this cloud editor](http://gitpod.io#https://github.com/trynmaps/metrics-mvp)**.
-This will automatically install requirements on a cloud machine
-and get the app running for you instantly.
+If you don't want to handle installing anything locally, just **[use this cloud editor](http://gitpod.io#https://github.com/trynmaps/metrics-mvp)**.
+This will automatically install requirements on a cloud machine and get the app running for you instantly.
```
hathix: Does GitPod play nicely with Docker? I think many people won't want to try getting Docker set up on their machines so support for GitPod is key.

README.md
```diff
@@ -2,48 +2,105 @@
...
 
-To make changes, make sure you've been added to the trynmaps organization
-on GitHub.
+## Computing Arrival Times
 
+The API and command-line scripts generate statistics based on Muni arrival times that are precomputed from the raw GPS data.
+The pre-computed arrival times are stored in S3 at http://opentransit-stop-arrivals.s3.amazonaws.com/?prefix=v2 with a separate JSON file
+for each route for each day.
```
hathix: Are these cached anywhere? We should call it out, if so

hathix:
Does this PR still auto-deploy things from the master branch to Heroku?

hathix:
Overall, your architecture changes look good from my end; Docker will simplify our process. Can't comment on the React code because I'm not too familiar with it. 

It would be great to see a demo / walkthrough on Wednesday, as well.

Thanks for doing this! This will be a big step forward.

README.md
youngj: The Dockerfile in this repo currently isn't used in Gitpod. Gitpod has its own Docker image with various tools preinstalled, and we're still using that for now. However, Gitpod's Docker image has different versions of python and node than those in the metrics-mvp Dockerfile. If that becomes a problem, we could create another Docker image for Gitpod later on 
 (https://docs.gitpod.io/42_Config_Docker.html#configure-a-custom-docker-image)

.gitpod.yml
youngj: Based on https://github.com/trynmaps/metrics-mvp/pull/45 I just updated it to use yarn everywhere for consistency. I don't really have an preference for whether to use npm or yarn.

youngj:
How does auto-deploying the master branch to Heroku work?

I guess that it would just require running `heroku stack:set container` once (as mentioned in the readme) before deploying this branch. 

If people would rather revert to using Procfile on Heroku instead of Docker, that works too. I originally didn't realize the Procfile system would compile the production frontend assets if you run `heroku buildpacks:add heroku/nodejs` .

Also if anyone feels strongly about having non-Docker local installation instructions in the Readme I could add those back too.

README.md
youngj: Yes, I'll add a sentence about how the JSON files are cached in the data/ directory.

PR #57, opened 2019-03-30
wait times command line utility, updated /metrics endpoint

jtanquil:
- updated `get_wait_times` to work with the updated ArrivalHistory class
- created a utility script, `waits.py`, to display summary statistics for waiting times at a particular stop
- added examples of waiting times (at a stop) and trip lengths (between two stops) to the /metrics endpoint in `metrics-api.py`
- refactored computation logic from `trips.py` and `metrics-api.py` into `metrics.py`
- created `timetable.py` to load Riley's scraped timetables into a DataFrame


models/wait_times.py
```diff
@@ -1,77 +1,50 @@
...
+    # truncate time interval to minute before first bus/minute after last bus leave
+    if start_time is None or \
+        (start_time is not None and (time.fromisoformat(start_time) < first_bus)):
+        start_time = first_bus.isoformat()[:-3]
+
+    if end_time is None or \
+        (end_time is not None and (time.fromisoformat(end_time) > last_bus)):
+        end_time = last_bus.isoformat()[:-3]
+
+    start_timestamp = datetime.strptime(f"{d} {start_time} -0800", "%Y-%m-%d %H:%M %z").timestamp()
```
youngj: Offset isn't always -0800. With tz set to `pytz.timezone('US/Pacific')`, I think `tz.localize(datetime.combine(d, start_time)).timestamp()` should work

waits.py
```diff
@@ -0,0 +1,92 @@
...
+import numpy as np
+import pandas as pd
+
+from models import nextbus, arrival_history, util, metrics, wait_times
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description = 'Compute wait times (in minutes) at a given stop in a given direction on a route, for one or more dates, optionally at particular times of day')
+    parser.add_argument('--route', required = True, help = 'Route id')
+    parser.add_argument('--stop', required = True, help = 'Stop id')
+    parser.add_argument('--direction', required = True, help = 'Stop direction (I for inbound, O for outbound)')
```
youngj: Would be easier for the user if the direction is inferred from the route and stop IDs rather than requiring the user to input it.

waits.py
```diff
@@ -0,0 +1,92 @@
...
+
+    route_config = nextbus.get_route_config(agency, route_id)
+    
+    stop_info = route_config.get_stop_info(stop)
+    stop_dirs = route_config.get_directions_for_stop(stop)
+    if len(stop_dirs) == 0 or stop_info is None:
+        raise Exception(f"invalid stop id {stop}")
+        
+    stop_dir = [d for d in stop_dirs if direction in d]    
+    if direction not in ["I", "O"] or len(stop_dir) == 0:
```
youngj: Would be clearer to rename `direction` as `timetable_direction` since it is not actually a Nextbus direction ID, which looks like 1____O_F00 (some routes have multiple unique direction IDs that contain "O" or "I")

models/timetable.py
```diff
@@ -0,0 +1,22 @@
...
+    # TODO: make this work wherever this is being called from
+    cwd = os.getcwd().split("\\")[-1]
+
+    if cwd == "metrics-mvp":
+      path = f"./data/timetables/route_{route_id}_timetables_data.csv"
+    elif cwd == "models":
+      path = f"./data/timetables/route_{route_id}_timetables_data.csv"
+    
+    df = pd.read_csv(path)
+    df['TIME'] = df['TIME'].apply(lambda x: datetime.strptime(f"{x} -0800", "%Y-%m-%d %H:%M:%S %z"))
```
youngj: Use `tz.localize(dt)` since offset isn't always -0800. Probably clearer if the column storing datetime objects is named DATE_TIME instead of TIME.

models/wait_times.py
```diff
@@ -1,77 +1,50 @@
...
+# assumes we have a df from one route, one stop, one direction
+def get_waits(df: pd.DataFrame, d: str, route: str, start_time = None, end_time = None):
+    # for each arrival time, rounding down to the nearest minute gives us the
+    # corresponding minute for which that arrival is first arrival
+    pst = timezone(timedelta(hours = -8))
+    waits = df['TIME'].copy(deep = True)
+    waits = pd.DataFrame({"ARRIVAL" : waits,
+                          "TIME_FLOOR" : waits - (waits % 60)})
+    # get corresponding timetable to determine the time interval for which to compute wait times
+    stop_id = int(("1" + df["SID"]).unique().squeeze())
```
youngj: Eventually we should look up the actual timetable stop ID via the route config since it may not always be "1" + Nextbus stop tag. Not necessary in this PR

models/timetable.py
```diff
@@ -0,0 +1,22 @@
...
+    day = date.fromisoformat(d).weekday()
+    timetable_type = "Weekday" if day < 5 else ("Saturday" if day == 5 else "Sunday")
+    
+    return df[(df['stop_id'] == stop_id) & (df['timetable'] == timetable_type)]
```
youngj: Eventually it would be better to save the timetables as JSON (instead of CSV) to make the relevant data more easily accessible. Not necessary for this PR.

README.md
```diff
@@ -72,6 +72,11 @@ Show stops visited by a particular vehicle:
 python vehicle.py --date=2019-02-01 --route=1 --vid=5737
 ```
 
+Show summary statistics of waiting times at a particular stop:
+```
+python waits.py --date=2019-02-01 --route=12 --stop=3476 --direction=I
```
youngj: When I run waits.py with these arguments, it says `longest wait time   = 144.8 min` , but  
`python headways.py --date=2019-02-01 --route=12 --stop=3476` says `longest headway    = 39.8 min`. 

I guess the 144.8 min is referring to the time between 3 AM and the first bus at 05:24:46, but the timetable shows the first expected bus is at 5:30. It seems times before 5:30 shouldn't count towards the max wait time.

models/timetable.py
hathix: also, if you are going to use -0800 you should put it into a named constant.

models/timetable.py
hathix: #59

metrics-api.py
```diff
@@ -59,7 +59,13 @@ def metrics_page():
         route_id = '12'
     stop_id = request.args.get('stop_id')
     if stop_id is None:
-        stop_id = '4970'
+        stop_id = '3476'
+    start_stop = request.args.get('start_stop')
+    if start_stop is None:
```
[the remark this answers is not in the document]
jtanquil: Removed `stop_id` and changed the names of `start_stop` and `end_stop` to `start_stop_id` and `end_stop_id`.

models/metrics.py
```diff
@@ -1,30 +1,101 @@
...
+    bin_size = math.ceil(2 * iqr/(len(df) ** (1/3)))
+
+    return bin_size
+
+def get_histogram(df: pd.Series):
+    bin_size = get_bin_size(df)
+    percentiles = range(0, 101, bin_size)
+    percentile_values = np.percentile(df, percentiles)
+
+    bin_min = math.floor(percentile_values[0] / bin_size) * bin_size
```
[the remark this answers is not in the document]
jtanquil: This was caused by passing in series with `NaN` values; changed the calling functions by dropping these values first.

README.md
jtanquil: I'll fix this in a later commit updating how the timetables data is read.

models/util.py
```diff
@@ -25,4 +27,15 @@ def get_dates_in_range(start_date_str, end_date_str, max_dates=1000):
         if len(res) > max_dates:
             raise Exception(f'too many dates between {start_date_str} and {end_date_str}')
 
-    return res
\ No newline at end of file
+    return res
+
+def get_data_dir():
+    return os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
```
youngj: get_data_dir() is not actually returning the data dir -- append `/data` to this

models/wait_times.py
```diff
@@ -1,77 +1,48 @@
...
+from datetime import datetime, timedelta, time 
+
+import pandas as pd
+import numpy as np
+
+from . import timetable, util
+
+
+# assumes we have a df from one route, one stop, one direction
+def get_waits(df: pd.DataFrame, d: str, tz: pytz.timezone, route: str, start_time = None, end_time = None):
```
youngj: I've been using `d` as a variable name for a date even though it's not ideal, because `date` often is imported as the class `datetime.date` (aside: maybe `date_obj` would be better? or just do `import datetime` instead of `from datetime import date`? Python's lowercase names for built-in classes are problematic). If a date variable is actually a string like "2019-04-04", I've been using the variable name `date_str`. In this case `d`, is what I would call `date_str`. Should either:
* rename the `d` variable and any other `d` str variables to `date_str`
* change `d` to actually be a date object instead of a str (Better in this case because `get_localized_datetime` needs to parse the date string again.... would be simpler if you just had a date object to begin with)

Also would be better to rename start_time/end_time to start_time_str/end_time_str to be consistent with other places in the code, so it is clear that it is a string like "04:00" and not a timestamp

models/wait_times.py
```diff
@@ -1,77 +1,48 @@
...
+    # for each arrival time, rounding down to the nearest minute gives us the
+    # corresponding minute for which that arrival is first arrival
+    waits = df['TIME'].copy(deep = True)
+    waits = pd.DataFrame({"ARRIVAL" : waits,
+                          "TIME_FLOOR" : waits - (waits % 60)})
+    # get corresponding timetable to determine the time interval for which to compute wait times
+    # TODO: get stop_id from route_config (enable route_config to do this)
+    stop_id = "1" + df["SID"][0]
```
youngj: This line throws an exception ` IndexError: index out of bounds` if df is empty, try http://localhost:5000/metrics?route_id=12&direction_id=12___O_F00&stop_id=7862&date=2019-02-01 . 

I'd suggest selecting various routes/stops from the frontend at http://localhost:3000 to make sure /metrics doesn't throw exceptions

metrics-api.py
```diff
@@ -121,46 +131,40 @@ def metrics_page():
...
+        waits.append(wait_times.get_waits(df, str(d), tz, route_id, start_time_str, end_time_str))
+        
+        if end_stop_id:
+            trips = trip_times.get_trip_times(df, history, tz, start_stop_id, end_stop_id)
 
         headway_min = df.headway_min[df.headway_min.notnull()] # remove NaN row (first bus of the day)
-        headway_min_arr.append(headway_min)
+        headway_min_arr.append(df.headway_min)
+
+        completed_trips.append(trips.trip_min[trips.trip_min.notnull()])
```
youngj: Trips may not be defined, move inside `if end_stop_id` block.

Also, throws `AttributeError: 'DataFrame' object has no attribute 'trip_min'` exception if df is empty, because trips.trip_min is not defined. Test with http://localhost:5000/metrics?route_id=12&direction_id=12___O_F00&stop_id=7862&date=2019-02-01

metrics-api.py
```diff
@@ -57,9 +57,12 @@ def metrics_page():
     route_id = request.args.get('route_id')
     if route_id is None:
         route_id = '12'
-    stop_id = request.args.get('stop_id')
-    if stop_id is None:
-        stop_id = '4970'
+    start_stop_id = request.args.get('start_stop_id')
```
youngj: If you rename stop_id to start_stop_id, will also need to update the frontend to use start_stop_id instead of stop_id.

metrics-api.py
```diff
@@ -57,9 +57,12 @@ def metrics_page():
...
+    if start_stop_id is None:
+        start_stop_id = '3476'
+    end_stop_id = request.args.get('end_stop_id')
+    if end_stop_id is None:
+        end_stop_id = '4973'
```
youngj: Better not to set a default value for end_stop_id because the front end may not always provide two stops and we don't need to compute trip times in that case. Also, the default end_stop_id may not correspond to the route_id . May also be better not to set a default value for start_stop_id for the same reason (though since it was that way already could leave it for this PR)

models/util.py
youngj: get_data_dir isn't actually returning the data dir, need to append /data

models/trip_times.py
```diff
@@ -0,0 +1,27 @@
...
+import pandas as pd
+import numpy as np
+
+from . import arrival_history
+
+def get_trip_times(df: pd.DataFrame, history: arrival_history.ArrivalHistory, tz: pytz.timezone, s1: str, s2: str):
+    s1_df = df.copy(deep = True)
+
+    if s1_df.empty:
+        return s1_df
```
youngj: In this case it doesn't add the trip_min series which is causing the metrics-api.py to throw an exception

models/wait_times.py
youngj: There is an error here if df is empty, e.g. http://localhost:5000/metrics?route_id=38&stop_id=7854 . 

In general, try selecting stops at http://localhost:3000 and make sure it doesn't throw errors.

models/wait_times.py
youngj: To be consistent with the rest of the codebase, variables named d should be date objects (avoiding the name `date` to avoid conflicting with the date class sometimes imported from the datetime module), and date strings should be named date_str. In this case it would make the code simpler if d was just a date object (since the caller of this function already has a date object), and then it wouldn't be necessary to parse the date string back to a date/datetime again.

Time strings like "04:00" should be start_time_str and end_time_str to distinguish them from timestamps.

PR #60, opened 2019-04-04
Pretty frontend + Regenerate build files

akgupta89:
**Summary**
Made the website responsive using bootstrap components and made docker-compose rebuild the production web page. You have to manually run "npm run build" to regenerate the production website normally.

**Preview**
Screen Shot 2019-04-03 at 8 27 30 PM


docker-compose.yml
```diff
@@ -5,10 +5,11 @@ services:
     build:
       dockerfile: Dockerfile
       context: .
-    ports:
-      - "5000:5000"
+    command: ["npm", "run", "build"]
```
youngj: This creates an optimized production build and exits without running the flask backend.

I think we shouldn't create an optimized production build here because:
- it's slow, which adds extra time waiting for your development environment to be ready
- it's usually not needed because people will be making changes to the front end at http://localhost:3000 which has live reloading 
- it's likely to confuse developers who look at http://localhost:5000 , try to make changes to the front end code and get confused because none of their changes are having any effect.

I'd like to keep the previous functionality and not create an optimized production build when running `docker-compose up`.

frontend/package-lock.json
```diff
@@ -0,0 +1,17372 @@
+{
```
youngj: package-lock.json should be deleted -- we are already using yarn.lock and `yarn build` / `yarn install`, and shouldn't run npm commands that generate package-lock.json (unless we make a decision to stop using yarn, but if we do that should be in a separate PR that removes all instances of yarn).

frontend/src/App.jsx
```diff
@@ -23,18 +23,8 @@ class App extends Component {
   render() {
     const { graphData, graphError, routes } = this.props;
     return (
-      <div className={css`
-        display: grid;
```
youngj: This removes the grid layout that was added last week in https://github.com/trynmaps/metrics-mvp/pull/45 . There are probably differing ideas about what CSS layout we're heading towards. The one column layout adapts well to mobile, though, which is nice. An advantage of using a grid is that it makes more efficient use of the screen space when the screen is wider (compared to using the entire screen width for the control panel). We could use flex box as a more responsive alternative to a grid layout: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/Relationship_of_Grid_Layout

frontend/package-lock.json
sravanti: Ah I thought we had decided on npm and not yarn—@youngj is that not correct? In any case, agreed that we just need to use one package manager.

frontend/src/App.jsx
sravanti: +1 here—with CSS grid, we can specify a one-column layout specifically for mobile-width screens, so let's investigate that first before removing CSS grid.

frontend/src/components/ControlPanel.jsx
```diff
@@ -113,43 +113,42 @@ class ControlPanel extends Component {
...
-          grid-column: col1-start / col3-start;
-           grid-row: row2-start ;
-      `
-      }
-        >
-          <DatePicker value={date} onChange={this.setDate} />
-          <DropdownControl title="Route" name='route' value={routeId}
+      <div className="row">
+        <div className="col-12 form-group"><DatePicker value={date} onChange={this.setDate} className={css`
+          display: block !important;
```
sravanti: Is there a reason why you're using `!important` here? It's typically best practice to avoid using it since it does disrupt the CSS specificity rules and should only be used in extreme edge cases (see this article: https://j11y.io/css/dont-use-important/).

docker-compose.yml
hathix: Agree; let's not do an optimized production build here.

akgupta89:
I think @sravanti made my PR obsolete, but we created two tasks from this:

1. yarn vs npm
2. update the documentation and provide the command to regenerate production files

hathix:
Have made Trello cards for these. Thanks Arun!

PR #64, opened 2019-04-11
Cleanup frontend

sravanti:
Cleaned up the intro panel a bit—still need to work on the d3 graphs.

Screen Shot 2019-04-10 at 8 48 04 PM


frontend/src/App.jsx
```diff
@@ -27,10 +27,11 @@ class App extends Component {
         display: grid;
         grid-gap: 4px;
         grid-template-columns: [col1-start] 300px [col2-start] 400px  [col3-start] auto [col3-end];
-        grid-template-rows: [row1-start] 200px [row2-start] 400px [row2-end];
+        grid-template-rows: [row1-start] 100px [row2-start] 400px [row2-end];
```
ckrajewski: nitpick, but should we get into the habit of using % instead of px for column widths?

frontend/src/components/Info.jsx
```diff
@@ -48,38 +17,37 @@ class Info extends Component {
...
-      <code>{graphError ? graphError : ''}{ graphData ? JSON.stringify(graphData) : null}</code>
+              xAxis={{label: "minutes"}}
+              barPadding={0.3}
+              style={{fill: 'red'}}
+              yAxis={{innerTickSize: 10, label: "number"}}
+            />
+          ) : null }
+        <code>
+          {graphError || ''}
+          { graphData ? <div className={css`color: white`}>JSON.stringify(graphData) </div> : null}
```
ckrajewski: @sravanti 
please get rid of this line. 
We are already printing out the JSON in console.log :) I added a middleware in redux that is showing the JSON file
image

ckrajewski: So we don't need to show/hide this on the page

youngj: just removed this line

PR #65, opened 2019-04-11
Added nextbus stop_id to StopInfo class

jtanquil:
Added the Nextbus stopId as the `nextbus_id` attribute to the StopInfo class and updated the function calls in `waits.py` and `metrics-api.py` accordingly. Also removed a debug statement from `waits.py`.


youngj:
In terms of naming, I think the stop "tag" (what we have been calling the stop ID) could be better described as a "Nextbus ID" than the "stopId", which originally comes from Muni rather than Nextbus. 

Maybe to be clear without needing to rename all of our previous references to stop IDs, it might be better to call the "stopId" something like location_id (instead of nextbus_id). Then we could continue using the term "stop" to refer to a location reached on a specific route in a specific direction, and use the new term ("location" or synonym) as a place that has one or more stops on one or more routes. 

possibilities: location_id, place_id, point_id, station_id, spot_id, site_id

jtanquil:
That's a good idea, I changed `nextbus_id` to `location_id` wherever it was referenced.

PR #71, opened 2019-04-18
Updated documentation to regenerate production

akgupta89:
Clearing this task:

https://trello.com/c/vpEJ46hF/130-update-documentation-to-tell-people-to-use-npm-run-build-to-make-a-prod-build


youngj:
This is currently in the section specific to Docker but also applies to GitPod, so it would probably be clearer to add it to a new section or the Command line scripts section.

akgupta89:
> This is currently in the section specific to Docker but also applies to GitPod, so it would probably be clearer to add it to a new section or the Command line scripts section.

Good point, fixed!

PR #73, opened 2019-04-18
[WIP] Replaced yarn with NPM

akgupta89:
Tested NPM vs Yarn
_Please do not merge this yet_

using NPM and disabling package auditing, the service starts up in ~45s (checks for deprecated/audited packages which needs to be skipped) 
using yarn, the service starts up in ~40s

There are a lot of pro NPM articles if you search for NPM vs Yarn 2019, but I don't know if there is any valuable metrics behind this. I think we should discuss this more before deciding on one or the other and possibly merging this PR

Other factors to consider:
* Package install time
* Future of tooling: 
  * Yarn: https://github.com/yarnpkg/yarn/issues/6953

Addressing task: https://trello.com/c/96kHqsAa/139-remove-all-yarn-usage-make-it-npm-instead


hathix:
naming this as a WIP so we don't accidentally merge

hathix:
Let us know once you're ready for review Arun!

sravanti:
Thanks for doing this benchmarking! It seems like from your analysis that NPM is faster, which IMO is fine to switch to, since I don't believe there are huge differences between NPM and Yarn. Looking through the future plans for yarn, I don't see anything that would significantly improve our workflow. For a MVP, I think you have the right idea that we should optimize for speed/developer experience.

One note of caution: I wouldn't suggest skipping package auditing—we don't want to download malicious versions of packages, which auditing would prevent.

akgupta89:
Updated my original post. It seems that they perform relatively the same, both in start time and downloading packages time so essentially its up in the air in regards to personal choice

akgupta89:
> Thanks for doing this benchmarking! It seems like from your analysis that NPM is faster, which IMO is fine to switch to, since I don't believe there are huge differences between NPM and Yarn. Looking through the future plans for yarn, I don't see anything that would significantly improve our workflow. For a MVP, I think you have the right idea that we should optimize for speed/developer experience.
> 
> One note of caution: I wouldn't suggest skipping package auditing—we don't want to download malicious versions of packages, which auditing would prevent.

Correct me if I'm wrong, but yarn doesn't audit every time it starts up, this check has to be run manually. That was the decision behind disabling auditing is to get a fair comparison between npm and yarn. I feel like that is a command that should be done manually, especially if yarn isn't doing the same and it adds significant delay to the start up time.

hathix:
I'm fine with just going with NPM if it's a roughly equivalent experience.

hathix:
Arun, is this ready to merge?

akgupta89:
> Looks good to me as long as you can confirm that this builds and runs properly.

Works fine, just tested. Ready to be merged.

PR #74, opened 2019-04-18
only show wait times between first and last actual arrival

youngj:
Using the timetable data for first/last bus times was causing the maximum wait time for most routes to be very large, which threw off the statistics, so I just changed it to use the first/last actual arrival for now.

image


hathix:
does this catch all the problems or are there any remaining ones that you can identify?

hathix:
data looks pretty reasonable to me

PR #77, opened 2019-04-25
Add front-end routing with redux-first-router

jimleeisme:
https://github.com/faceyspacey/redux-first-router


frontend/src/components/About.js
```diff
@@ -0,0 +1,20 @@
...
+
+  render() {
+    return (
+      <h3>About Us</h3>
+    );
+  }
+}
+
+const mapStateToProps = (state) => {
+  return state;
```
sravanti: Do we need mapStateToProps here if we're just copying over the whole state? It would make better sense to use mapStateToProps in the components that need only certain parts of state (otherwise, what this is saying is to rerender the About page any time _any_ part of state changes instead of only rerendering when state relevant to the About page changes).

frontend/src/App.jsx
```diff
@@ -1,68 +1,22 @@
...
+    About: <About />,
+    NotFound: <NotFound />
+  };
+  return (
+    <div>{components[page]}</div>
+  );
 };
 
-export default connect(mapStateToProps, mapDispatchToProps)(App);
+const mapStateToProps = ({ page }) => ({ page });
```
sravanti: What is the purpose of mapStateToProps here? I'm not sure I understand what's being mapped / if this is necessary at all.

frontend/src/components/About.js
```diff
@@ -0,0 +1,20 @@
+import React, { Component } from 'react';
+import { connect } from 'react-redux';
+
+class About extends Component {
+  constructor(props) {
```
sravanti: (Nit) you don't need to define a constructor if it's just calling `super(props)` (I recognize this is code from earlier and may/may not be yours)

frontend/src/components/Home.js
```diff
@@ -0,0 +1,72 @@
...
+import Intro from './Intro';
+import {
+  fetchGraphData,
+  fetchRoutes,
+  fetchRouteConfig,
+  resetGraphData,
+} from '../actions';
+
+class Home extends Component {
+  constructor(props) {
```
sravanti: Ditto re: constructor

frontend/src/store.js
```diff
@@ -0,0 +1,24 @@
...
+import { connectRoutes } from 'redux-first-router';
+import thunk from 'redux-thunk';
+
+import routesMap from './routesMap';
+import * as reducers from './reducers';
+import page from './reducers/page';
+import * as actionCreators from './actions';
+
+const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__
+  ? window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__({ actionCreators })
```
sravanti: Awesome, thank you for adding dev tools

frontend/src/components/About.js
jimleeisme: Fixed. Thanks!

frontend/src/components/About.js
jimleeisme: Good to know - thanks

frontend/src/App.jsx
jimleeisme: Here we're getting the `page` so we know which page component (home, about, etc.) to render, if that makes sense.

frontend/src/components/About.js
```diff
@@ -0,0 +1,15 @@
+import React, { Component, Fragment } from 'react';
```
youngj: Suggest using .jsx file extension for consistency with other components that use JSX (same for Home and NotFound)

jimleeisme: Fixed.

jimleeisme:
> With redux-first-router, how would we update the URL when the control panel inputs change? And then when the page is loaded, how would we initialize the control panel based on the URL? Could be implemented in a separate PR.

@youngj We'll handle the control panel state with redux and the corresponding actions will update the state and URL (and vice versa)

hathix:
Changes LGTM

PR #80, opened 2019-04-30
Bus grade

exxonvaldez:
I've added an implementation of route grading.  There are subscores for 1. average wait, 2. long wait probability, 3. travel variability, and 4. travel speed: https://docs.google.com/document/d/19eMnShDf8f6Qitq175uYil7StwzSkuVXSNDUfyeVCMw/edit?usp=sharing

To calculate travel distance, I added the "graph parameters"  to the Redux state. This allows other React components to know what the graphed route and direction are.  I also exposed lat/lon for stops in the route config.  The distance is calculated using straight line distance between stops.


frontend/src/components/ControlPanel.jsx
```diff
@@ -80,12 +80,25 @@ class ControlPanel extends Component {
...
+    const { directionId, secondStopId } = this.state;
     const selectedRoute = { ...this.getSelectedRouteInfo() };
     const secondStopInfo = this.getStopsInfoInGivenDirection(selectedRoute, directionId);
     const secondStopListIndex = secondStopInfo.stops.indexOf(stopId);
     const secondStopList = secondStopInfo.stops.slice(secondStopListIndex + 1);
-    this.setState({ firstStopId: stopId, secondStopList }, this.selectedStopChanged);
+    
+    let newSecondStopId = secondStopId;
+    
+    // If the "to stop" is not set or is not valid for the current "from stop",
```
sravanti: Thank you for adding this comment! Really helpful to parse the logic here.

exxonvaldez: Thanks for the review, Sravanti.  Agree that having d3 libraries saves a lot of wheel reinventing.  Found the import statement for d3 among differing examples on Google. It was kind of tricky because we're using an older version of d3 (v3).  I'll wait for another review.

-- Terence

frontend/src/components/Info.jsx
```diff
@@ -1,19 +1,182 @@
...
+  }
+  
+  /**
+   * This is approximate straight line distance between two lat/lon's,
+   * assuming a fixed distance per degree of longitude, and that the
+   * distance per degree of latitude varies by cosine of latitude.
+   *
+   * http://jonisalonen.com/2014/computing-distance-between-coordinates-can-be-simple-and-fast/
+   */
+  milesBetween(p1, p2) {
```
youngj: The backend computes distances between lat/lon using the haversine formula, and returns distances in meters (haver_distance in eclipses.py). It'd probably be more consistent to have a function that uses the haversine formula and returns meters on the front end, then the caller could convert from meters to miles if necessary. Using meters will also be helpful for mapping since leaflet.js uses meters as its unit of distance.

frontend/src/components/Info.jsx
```diff
@@ -1,19 +1,182 @@
...
+    .domain([5, 7.5, 10])
+    .range(["A", "B", "C", "D"]);
+    
+    //
+    // grade and score for long wait probability
+    // 
+    // where probability of 20 min wait is sum of histogram buckets #4 and higher divided by count
+    //
+    
+    const TWENTY_INDEX = 4;
```
youngj: As an alternative to hardcoding the index, could modify the data returned by get_histogram in metrics.py to return the minimum/maximum values of each bin as a property of the bin. Or, could update get_wait_times_stats in metrics.py to return the probability of a 20 minute wait directly.

frontend/src/components/Info.jsx
```diff
@@ -25,6 +188,31 @@ class Info extends Component {
...
+            90% of waits under { Math.round(waitTimes.percentiles[18].value) } minutes
+            </td><td>{grades.averageWaitGrade}</td><td> {grades.averageWaitScore} </td>
+            </tr>
+            <tr>
+            <td>20 min wait probability</td><td> {Math.round(grades.longWaitProbability * 100)}% { grades.longWaitProbability > 0 ? "(1 time out of " + Math.round(1/grades.longWaitProbability) + ")" : ""} <br/>
+            </td><td> { grades.longWaitGrade } </td><td> { grades.longWaitScore } </td>
+            </tr>
+            <tr>
+            <td>Travel time</td>
+            <td>Average time {Math.round(tripTimes.avg)} minutes ({ speed } mph)
```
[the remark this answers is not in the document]
exxonvaldez: Done.

frontend/src/components/Info.jsx
```diff
@@ -1,19 +1,182 @@
...
+        
+    //
+    // grade score for travel time variability
+    //
+    // where variance is 90th percentile time minus average time
+    //
+    
+    let travelVarianceTime = 0;
+    if (tripTimes) {
+        travelVarianceTime = tripTimes.percentiles[18].value - tripTimes.avg;
```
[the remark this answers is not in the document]
exxonvaldez: Okay, added a getPercentile function.  Also, the percentile objects in the array are already labeled with "percentile" number like 0, 5, 10, etc., so using that.

frontend/src/components/Info.jsx
exxonvaldez: Put in a javascript haversine function that uses meters.

frontend/src/components/Info.jsx
exxonvaldez: Modifed metrics.py so each bin has a bin_start and bin_end value (not to be confused with bin_min/bin_max, which refer to the range across all bins).

exxonvaldez:
Good feedback, Jesse, thanks.  All incorporated to the pull request now.

hathix:
Feel free to merge

PR #82, opened 2019-04-30
edit the chart

fruit13ok:
styled the chart and changed y-axis amount of ticks, tick format, label.

Screen Shot 2019-04-30 at 2 14 10 PM


hathix:
This is good as long as you can confirm this builds correctly (with the yarn.lock change).

PR #88, opened 2019-05-02
Show front page without error

vicky11z:
Frontend was loading this error:
Screen Shot 2019-05-01 at 9 19 11 PM

@jimleeisme suggested this one line change that fixes it


hathix:
Do we know why the error is getting created in the first place?

jimleeisme:
Oh, oops. I should have done some more digging before making the change. You should still be able to investigate the `RECEIVED_ROUTES_ERROR` action in redux devtools. I would do it myself, but I can't replicate the error.

PR #89, opened 2019-05-07
Compare scheduled timetable to actual arrival times

jtanquil:
This pull request adds modules for scraping GTFS data and working with the scheduled timetables obtained from that data.

`gtfs.py` contains the code that works directly with the GTFS data. For now, the purpose of the instance methods is just to extract the timetables for each route into csv files, but given the wealth of information in the GTFS data it would make sense to add additional functionality to extract that data in the future. Currently, the code requires that the GTFS data be available locally in a given directory. Given that this data seems to be updated by SFMTA fairly regularly, automating the process of scraping these files at regular intervals is another possible future feature. For now, you'll have to get the GTFS files from SFMTA (https://www.sfmta.com/reports/gtfs-transit-data) and run `gtfs_scraper.py` to load the timetable data locally.

`timetable.py` contains the code that takes data from the CSVs scraped from GTFS and returns a timetable for a particular stop as a Pandas DataFrame. The function `compare_timetable_to_actual` has been added to `metrics.py` to compare the scheduled arrival times with the actual arrival data for a given stop. At the moment this function works with dummy data by changing the dates of the arrival data passed in, but this can be removed once we have arrival data and scheduled data for some dates.

`gtfs_scraper.py` is a command line script that extracts timetables from GTFS data in a given directory. `timetables.py` is a script that returns the timetable for a given stop, and can also compare that to actual arrival data from the stop.


hathix:
@youngj let's get this reviewed

.gitignore
```diff
@@ -343,5 +343,8 @@ typings/
 .idea
 # theia
 .theia
-# test notebook
-/models/Untitled.ipynb
\ No newline at end of file
+# test notebooks
+/models/Untitled.ipynb
```
hathix: You need to fix the merge conflict in gitignore. i made some changes on master to gitignore. you can pull in those changes and add your custom stuff at the bottom of the file.

.gitignore
```diff
@@ -343,5 +343,8 @@ typings/
...
+/*.ipynb
```
hathix: why this? do you have experimental ipynbs in the home folder? i would avoid doing that because it might confuse someone who *does* put a real ipynb in the home folder later and doesn't notice it's in the gitignore.

README.md
```diff
@@ -85,6 +85,16 @@ Show summary statistics of waiting times at a particular stop:
 python waits.py --date=2019-02-01 --route=12 --stop=3476
 ```
 
+Show scheduled timetable for a particular stop:
+```
+python timetables.py --route=12 --stops=3476 --date=2019-04-12
```
hathix: thanks for putting instructions in the readme!

models/gtfs.py
```diff
@@ -0,0 +1,239 @@
...
+        if len(gtfs_route_id) > 0:
+            return gtfs_route_id[0]
+        else:
+            raise Exception(f"No gtfs route id found for route {route}.")
+
+    def get_gtfs_route_ids(self):
+        feed = ptg.load_geo_feed(self.inpath, {})
+        routes = feed.routes        
+        nextbus_route_ids = [route.id for route in nextbus.get_route_list("sf-muni")]
+        # convert nextbus naming convention for OWL routes to gtfs naming convention
```
hathix: can you give an example of what the OWL route's conventions are, and what they should be turned into?

models/gtfs.py
```diff
@@ -0,0 +1,239 @@
...
+            stop_times["nextbus_id"] = stop_times["stop_id"].apply(lambda x: get_nextbus_stop_id(x, gtfs_direction, route_config))
+
+            return stop_times[["trip_id", "arrival_time", "departure_time", "stop_id", "nextbus_id", "stop_sequence", "direction"]].sort_values("arrival_time")
+        else:
+            print(f"Stop times df for route {route_id} on {date} going {direction} is empty")
+            return pd.DataFrame()
+
+    def get_excluded_stops(self, route_id, stop_times_df, route_config, direction, d):
+        stop_dirs = {
+            stop.id:(1 if "I" in route_config.get_directions_for_stop(stop.id)[0]
```
hathix: can you clarify in a comment that 1 = inbound, 0 = outbound?

models/metrics.py
```diff
@@ -90,6 +92,63 @@ def filter_by_time_of_day(df: pd.DataFrame, start_time_str, end_time_str) -> pd.
...
     return df
 
+def compare_timetable_to_actual(tt: timetable.Timetable, df: pd.DataFrame, direction: str):
+    try:
+        stops_df = df.copy(deep = True)
+        stop_id = stops_df["SID"].unique().squeeze()
+        timetable = tt.get_data_frame(stop_id, direction)
+
+        #use dummy data for now
+        delta = tt.date - date(2019, 4, 8)
```
hathix: i feel like we use 4/8 as a hardcoded value all around the codebase. can we put it in a constants file?

timetables.py
```diff
@@ -0,0 +1,119 @@
...
+        comparison = False
+    elif args.comparison.lower() == 'true':
+        comparison = True
+    else:
+        raise Exception("comparison must be true or false")
+
+    threshold = 5 if args.threshold is None else int(args.threshold)
+
+    inpath = "data"
+    agency = "sf-muni"
```
hathix: can these be named constants in another file? that way other people could more easily borrow our code for other transit agencies

timetables.py
```diff
@@ -0,0 +1,119 @@
...
+                            print(f"Actual Arrival | Actual Headway | Closest Scheduled Headway | {'Delta'.ljust(9)} | First After Arrival Headway | {'Delta'.ljust(9)}")
+                            for row in stops_df.itertuples():
+                                actual_arrival = row[1].time().isoformat()
+                                actual_headway = f"{round(row[6], 1)} min"
+                                closest_scheduled_headway = f"{round(row[7], 1)} min"
+                                closest_headway_delta = f"{round(row[8], 1)} min"
+                                first_after_headway = f"{round(row[9], 1)} min"
+                                first_after_headway_delta = f"{round(row[10], 1)} min"
+
+                                print(f"{actual_arrival.ljust(14)} | {actual_headway.ljust(14)} | {closest_scheduled_headway.ljust(25)} | {closest_headway_delta.ljust(9)} | {first_after_headway.ljust(28)} | {first_after_headway_delta.ljust(9)}")
```
hathix: is there a cleaner way to print this information than handcrafting this printed table? i feel like this way is pretty fragile, and the `14`s and all are confusing.

.gitignore
```diff
@@ -343,5 +343,8 @@ typings/
...
+# timetable data (~148 MB of csv files)
+/data/*.csv
```
hathix: make sure this doesn't have any unintended consequences -- maybe people will put other CSVs they want to track in the /data folder. Maybe put the timetable data in a subfolder and gitignore that.

models/gtfs.py
```diff
@@ -0,0 +1,239 @@
...
+from . import nextbus, util
+
+class GtfsScraper:
+    def __init__(self, inpath):
+        self.inpath = inpath
+
+    def get_gtfs_route_id(self, route):
+        # convert nextbus naming convention for OWL routes to gtfs naming convention
+        route_view = {
+            "routes.txt": {"route_short_name": f"{route.replace('_', '-')}"}
```
hathix: is routes.txt a file we have locally? can the filename be passed as a parameter to this function, for robustness's sake?

models/gtfs.py
```diff
@@ -0,0 +1,239 @@
...
+            added_services = changed_service_ids[changed_service_ids.exception_type == 1].service_id
+            removed_services = changed_service_ids[changed_service_ids.exception_type == 2].service_id
+
+            # add service_ids that have been added that day (exception_type == 1)
+            service_ids = service_ids.append(added_services)
+            # remove service_ids that have been removed that day (exception_type == 2)
+            service_ids = service_ids[~(service_ids.isin(removed_services))]
+
+        trip_view = {
+            "trips.txt": {
```
hathix: again, where is this trips.txt file?

models/gtfs.py
```diff
@@ -0,0 +1,239 @@
...
+        # get time string from the timestamps
+        def get_time_isoformat(t: float):
+            second = "{:02d}".format(int(t) % 60)
+            minute = "{:02d}".format((int(t) // 60) % 60)
+            hour = "{:02d}".format((int(t) // 3600) % 24)
+
+            return f"{hour}:{minute}:{second}"
+
+        # stop times past midnight happen on the next day
+        def get_date(t: float):
```
hathix: love that you're describing the types of the inputs in the function headers!

models/gtfs.py
```diff
@@ -0,0 +1,239 @@
...
+            return d if t < 86400 else d.replace(day = d.day + 1)
```
hathix: can you just do 60 * 60 * 24 instead of 86400 so it's more clear what the number stands for?

models/gtfs.py
hathix: or can we have some helper function that does this conversion? we do the inbound/outbound code conversion a lot

models/gtfs.py
```diff
@@ -0,0 +1,239 @@
...
+            opposite_dir_stop = [stop_id for stop_id in nextbus_stop_ids if gtfs_stop_id in stop_id]
+
+            try:
+                return opposite_dir_stop[0]
+            except Exception as err:
+                raise Exception(f"Could not find nextbus stop id for {gtfs_stop_id}: {err}")
+    else:
+        # if there's no direction on nextbus...
+        # for now just return nan (so it'll be effectively ignored)
+        # TODO: deal with stops w/o direction (on nextbus)
```
hathix: is this related to the issue we had raised earlier?

exxonvaldez:
Neat to see schedule data coming into the project.  Frequently changing schedule data seems like a challenge.  Obviously unfair to compare actual performance vs a schedule that was not in effect at the time.

In addition to the "late by 5 minutes" (gap) percentage, SFMTA has a "very late" bucket (10 minutes or more) for its on-time scorecard:  https://sfgov.org/scorecards/transportation/transit-time-performance .  Would definitely like to have the gap percentage available in the metrics API, and also early/on-time/late/very late percentages (where I guess gap is just sum of late + very late).

Also I think it'd be great to have the "schedule versions" of the graph data in the metrics API (headways, wait times, travel times), so that they can be compared with actual.  Histograms, number of arrivals, averages, percentiles, etc.  As long as that's not too difficult.

models/metrics.py
```diff
@@ -90,6 +92,63 @@ def filter_by_time_of_day(df: pd.DataFrame, start_time_str, end_time_str) -> pd.
...
+        
+        stops_df["closest_headway_delta"] = stops_df["headway"] - stops_df["closest_scheduled_headway"]
+        stops_df["first_headway_delta"] = stops_df["headway"] - stops_df["closest_first_after_headway"]
+
+        stops_df = stops_df.rename(mapper = {"TIME": "actual_arrival_time"}, axis = "columns")
+        
+        return stops_df[["actual_arrival_time", "closest_scheduled_arrival", "closest_delta", "first_scheduled_after_arrival", "first_after_delta", "headway", "closest_scheduled_headway", "closest_headway_delta",  "closest_first_after_headway", "first_headway_delta"]]
+    except Exception as err:
+        print(f"Error occurred while comparing timetable and actual data: {repr(err)}")
+        return pd.DataFrame()
```
youngj: Swallowing exceptions makes debugging difficult, could it just let the exception propagate to the caller?

models/metrics.py
```diff
@@ -90,6 +92,63 @@ def filter_by_time_of_day(df: pd.DataFrame, start_time_str, end_time_str) -> pd.
...
+            return timetable.iloc[nonnegative_deltas.idxmin()].arrival_time if len(nonnegative_deltas) > 0 else np.nan
+
+        def get_closest_timetable_time(dt: datetime):
+            deltas = (timetable["arrival_time"] - dt).apply(lambda x: abs(x.total_seconds()))
+            return timetable.iloc[deltas.idxmin()].arrival_time if ~np.isnan(deltas.idxmin()) else np.nan
+
+        def get_corresponding_scheduled_headway(s: pd.Series):
+            return pd.Series({
+                "closest_scheduled_headway": timetable[timetable["arrival_time"] == s.closest_scheduled_arrival].arrival_headway.values[0],
+                "closest_first_after_headway": timetable[timetable["arrival_time"] == s.first_scheduled_after_arrival].arrival_headway.values[0]
```
youngj: Throws IndexError if no rows match the condition

.gitignore
jtanquil: good point, i'll move mine to another branch

models/gtfs.py
jtanquil: yeah, almost all of these stops have no data so they should be fine to ignore for now

hathix:
Looks like this needs a bit more attention to address comments from Jesse (and a few of my old comments). Let's get those addressed then we can merge this!

jtanquil:
The latest commit stores the files in a s3 bucket I created, and will pull timetables from that and cache them locally when they're needed. I will add the hardcoded values throughout the changes in `constants.py` and work on @exxonvaldez 's request for timetable metrics in future pull requests. Thank you for all the feedback!

models/timetable.py
```diff
@@ -1,40 +1,120 @@
...
+            data = f.read() 
+    except FileNotFoundError as err:
+        s3_bucket = get_s3_bucket()
+        data = requests.get(f"http://{s3_bucket}.s3.amazonaws.com/{date_range_str}/{filename}").text
+
+        with open(f"{path}/{filename}", "w") as f:
+            f.write(data)
+
+    timetable = csv_string_to_df(data)
+    timetable[["arrival_time", "departure_time"]] = timetable[["arrival_time", "departure_time"]].applymap(lambda x: datetime.strptime(x, "%Y-%m-%d %H:%M:%S%z") + delta)
```
youngj: If storing int offsets in CSV instead of datetime, will be simpler to create datetime here with date d like `midnight + timedelta(seconds=t)`.

models/timetable.py
```diff
@@ -1,40 +1,120 @@
...
 
-    # mock data for stops that don't have a timetable yet
-    timetable.index = range(len(timetable))
-    if len(timetable) == 0:
-      timetable = pd.DataFrame({
-        "DATE_TIME" : [util.get_localized_datetime(d, t) for t in ["00:00", "23:59"]],
-        "stop_id" : stop_id,
-        "timetable" : timetable_type
-        })
+        return list(map(lambda x: datetime.combine(x, time(), tzinfo = pytz.timezone('US/Pacific')), [date_range[0], date_range[-1]]))
```
youngj: Could move duplicate return line after if/else statement so it is only needed once.

However it seems like this method doesn't actually need to return datetime objects since the caller get_timetable_from_csv only looks at the date objects. Seems like it could return the two date objects.

models/timetable.py
```diff
@@ -1,40 +1,120 @@
...
+
+def get_s3_bucket():
+    return "opentransit-muni-schedules"
+
+def csv_string_to_df(s: str):
+    # remove empty rows
+    elements = [list(map(lambda x: x.strip(), x.split(","))) for x in s.split("\n") if len(x) > 0]
+    return pd.DataFrame(elements[1:], columns = elements[0])
+
+def get_timetable_from_csv(path: str, agency: str, route_id: str, d: date):
```
youngj: Does `path` need to be a parameter? Could it always be `util.get_data_dir()`?

gtfs_scraper.py
```diff
@@ -0,0 +1,36 @@
...
+from datetime import datetime, date, time
+
+import pandas as pd
+
+from models import gtfs, util
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description = "Get timetables from GTFS data")
+    parser.add_argument("--inpath", required = True, help = "Path to directory containing GTFS data")
+    parser.add_argument("--outpath", help = "Destination directory for timetable CSVs")
```
youngj: Does `outpath` need to be a parameter? Could it always write to `util.get_data_dir()`?

models/gtfs.py
```diff
@@ -0,0 +1,275 @@
...
+    def save_date_ranges(self, outpath, s3 = False):
+        df = self.get_date_ranges()
+        filepath = f"{outpath}/date_ranges.csv"
+
+        if s3:
+            s3_path = f"date_ranges.csv"
+            self.upload_to_s3(s3_path, df)
+        
+        if Path(filepath).is_file():
+            print(f"{datetime.now().time().isoformat()}: Date ranges have already been cached locally.")
```
youngj: Over time there will be new gtfs files with new date ranges, that don't include historical date ranges. Would we want to add new date ranges to the existing file? Could be something for a later PR.

models/gtfs.py
```diff
@@ -0,0 +1,239 @@
...
+            for route in [route.replace("_", "-") for route in nextbus_route_ids]
+            if len(routes[routes.route_short_name == f"{route}"].route_id.values) > 0
+        ]
+
+        return {
+            nextbus_route_id:gtfs_route_id for nextbus_route_id, gtfs_route_id in zip(nextbus_route_ids, gtfs_route_ids)
+        }
+
+    def get_route_trips_by_date(self, route_id, d):
+        feed = ptg.load_geo_feed(self.inpath, {})
```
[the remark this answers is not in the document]
youngj: GtfsScraper is much faster now, thanks for making this change!

models/timetable.py
```diff
@@ -1,40 +1,120 @@
...
+                }, axis = "columns"))
+        else:
+            print(f"No timetable found for {stop_id} on route {self.route_id} on {self.date} going {self.get_stop_direction(stop_id)}.")
+
+def get_s3_bucket():
+    return "opentransit-muni-schedules"
+
+def csv_string_to_df(s: str):
+    # remove empty rows
+    elements = [list(map(lambda x: x.strip(), x.split(","))) for x in s.split("\n") if len(x) > 0]
```
youngj: I'm not sure why there would be empty rows in the CSV files that are created by GtfsScraper. Could GtfsScraper avoid writing empty rows so that timetable.py doesn't need to remove them?

I'd expect pd.read_csv to be faster than parsing in python (and also able to handle quoted fields).  The dtype parameter to pd.read_csv can be used read numeric fields as strings.

models/timetable.py
```diff
@@ -1,40 +1,120 @@
...
+
+def get_timetable_from_csv(path: str, agency: str, route_id: str, d: date):
+    date_period = get_date_period(path, d)
+    delta = datetime.combine(d, time(), tzinfo = pytz.timezone('US/Pacific')) - date_period[0]
+    date_range_str = f"{date_period[0].date().isoformat()}_to_{date_period[-1].date().isoformat()}"
+    filename = f"route_{route_id}_{date_range_str}_timetable.csv"
+    
+    # checks for a local file; if it doesn't exist, pull it from the s3 bucket and cache it locally
+    try:
+        with open(f"{path}/{filename}", "r") as f:
```
youngj: It looks like GtfsScraper is writing the files to a subdirectory (which seems like a good idea to keep the data directory more organized) but this is reading from the root data/ directory

models/timetable.py
```diff
@@ -1,40 +1,120 @@
...
+            data = f.read() 
+    except FileNotFoundError as err:
+        s3_bucket = get_s3_bucket()
+        data = requests.get(f"http://{s3_bucket}.s3.amazonaws.com/{date_range_str}/{filename}").text
+
+        with open(f"{path}/{filename}", "w") as f:
+            f.write(data)
```
youngj: May need to create subdirectory if it doesn't exist

models/timetable.py
```diff
@@ -1,40 +1,120 @@
...
+                    "arrival_time": "Arrival Time",
+                    "departure_time": "Departure Time",
+                    "arrival_headway": "Arrival Headway",
+                    "departure_headway": "Departure Headway"
+                }, axis = "columns"))
+        else:
+            print(f"No timetable found for {stop_id} on route {self.route_id} on {self.date} going {self.get_stop_direction(stop_id)}.")
+
+def get_s3_bucket():
+    return "opentransit-muni-schedules"
```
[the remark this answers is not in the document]
jtanquil: that's a bucket I made on my free tier AWS account, which will be around until sometime in november - I figured this would be fine for testing purposes and we can move it when necessary

models/gtfs.py
jtanquil: Yeah, periodically updating date ranges and uploading new timetables will be something that needs to be done in the future - ideally we would automate the process of regularly updating the files whenever SFMTA updates the GTFS information.

jtanquil:
Apologies for the wait, I thought I pushed these changes on Wednesday but somehow I managed to lose some of them in the process. But this latest commit has all the changes I made earlier this week.

models/gtfs.py
```diff
@@ -0,0 +1,263 @@
...
+                filename = f"{self.agency}_route_{nextbus_route_id}_{date_range_string}_timetable_{self.version}.csv"
+                filepath = f"{subdirectory}/{filename}"
+                csv_path = f"{outpath}/{filepath}"
+                local_file_exists = Path(csv_path).is_file()
+                stops = []
+
+                if local_file_exists:
+                    print(f"{datetime.now().time().isoformat()}: The file {filepath} already exists, skipping.")
+                else:
+                    rc = nextbus.get_route_config(self.agency nextbus_route_id)
```
youngj: `SyntaxError: invalid syntax` - missing comma

models/gtfs.py
```diff
@@ -0,0 +1,263 @@
...
+            ContentEncoding = 'gzip',
+            ContentType = 'text/csv',
+            ACL = 'public-read'
+        )
+        
+        print(f"{datetime.now().time().isoformat()}: Uploaded {s3_path} to s3 bucket.")
+
+    def save_date_ranges(self, outpath, s3 = False):
+        df = self.get_date_ranges()
+        filepath = f"{outpath}/date_ranges.csv"
```
youngj: paths written are different from paths read by timetable.py (which eventually causes timetable.py to raise `AttributeError: ("'Series' object has no attribute 'start_date'", 'occurred at index 0')`). would be good to have functions that both places could call to get the paths

models/metrics.py
```diff
@@ -90,6 +128,63 @@ def filter_by_time_of_day(df: pd.DataFrame, start_time_str, end_time_str) -> pd.
         df = df[df.TIME_STR < end_time_str]
     return df
 
+def compare_timetable_to_actual(tt: timetable.Timetable, df: pd.DataFrame, direction: str):
+    stops_df = df.copy(deep = True)
+    stop_id = stops_df["SID"].unique().squeeze()
+    timetable = tt.get_data_frame(stop_id, direction)
+
+    #use dummy data for now
+    delta = tt.date - date(2019, 4, 8)
```
youngj: this variable seems unused, remove?

timetables.py
```diff
@@ -0,0 +1,121 @@
...
+        nextbus_dir = rc.get_directions_for_stop(stop)
+        if len(nextbus_dir) == 0:
+            print(f"Stop {stop} has no directions.")
+        else:
+            for direction in nextbus_dir:
+                tt.pretty_print(stop, direction)
+                
+                if comparison:
+                    # get dummy data for now
+                    ah = arrival_history.get_by_date(agency, route, date(2019, 4, 8))
```
youngj: This causes errors when the --date argument is not 2019-04-08, but seems to work when replacing the hardcoded date with `d` (for 2019-04-08, 2019-05-08, 2019-05-11, 2019-05-24)

models/gtfs.py
jtanquil: I added `get_s3_bucket` and `get_schedule_dir` to `gtfs.py`, felt like it made more sense to keep those functions there.

models/util.py
```diff
@@ -36,6 +36,12 @@ def render_dwell_time(seconds):
 def get_data_dir():
     return f"{os.path.dirname(os.path.dirname(os.path.realpath(__file__)))}/data"
 
+def get_schedule_s3_bucket(agency: str):
```
youngj: these functions seem to be duplicates of functions in gtfs.py, remove?

models/gtfs.py
```diff
@@ -0,0 +1,267 @@
...
+            except Exception as err:
+                raise Exception(f"Could not find nextbus stop id for {gtfs_stop_id}: {err}")
+    else:
+        # if there's no direction on nextbus...
+        # for now just return nan (so it'll be effectively ignored)
+        # TODO: deal with stops w/o direction (on nextbus)
+        return np.nan
+
+def get_s3_bucket(agency: str):
+    return f"opentransit-{agency}-schedules"
```
youngj: It's possible that a bucket for each agency may not be available since S3 bucket names are a global namespace. It would be preferable to include the agency ID as part of the S3 path and have a fixed bucket name that is not specific to an agency.

The S3 bucket should be in the same AWS account as the other AWS buckets, otherwise it's going to be difficult to manage access (so that you aren't the only one who can write schedules to S3). @EddyIonescu could probably help with that.

youngj:
Could you change models/timetable.py to use Unix line endings? It would be nice to eventually get all of the files to use the same line endings, if only so that `git diff` doesn't show ^M at the end of modified lines.

jtanquil:
Line endings in `models/timetable.py` should be fixed, and I changed my vscode settings to use unix line endings.

youngj:
Looks like models/timetable.py was deleted in the last commit?

jtanquil:
Oops, reverted it and fixed the line endings, `^M` no longer shows up on my diff so it should be alright.

models/util.py
youngj: these duplicate functions are back in https://github.com/trynmaps/metrics-mvp/pull/89/commits/675d6c9a5a8a1ed3cfa59faee0368b660fa44764

jtanquil: Thanks, I'm going to work on updating some of the comparison features to use `RouteMetrics` so I'll fix this in that commit.

jtanquil:
Added some functions to the `RouteMetrics` object to compute summary statistics for timetable headways and compare scheduled arrival times to arrival data.

models/metrics.py
```diff
@@ -72,6 +72,139 @@ def get_data_frame(self, d, stop_id=None, direction_id=None):
         self.data_frames[key] = df
         return df
 
+    def get_route_timetable(self, d):
+        try:
+            self.timetables
+        except AttributeError:
+            self.timetables = {}
```
youngj: Probably simpler to do `self.timetables = {}` in the constructor

models/metrics.py
```diff
@@ -72,6 +72,139 @@ def get_data_frame(self, d, stop_id=None, direction_id=None):
...
+        stop_timetable = self.get_stop_timetable(d, stop_id, direction_id)
+        stop_arrivals = self.get_data_frame(d, stop_id, direction_id)
+
+        # first headway is always nan
+        arrival_headways = np.insert(compute_headway_minutes(stop_arrivals['TIME'].to_numpy()), 0, np.nan)
+        stop_arrivals['headway'] = arrival_headways
+
+        # for each scheduled arrival time, get the next actual arrival and the corresponding headway
+        def get_next_arrival_time(scheduled_arrivals, arrivals, arrival_headways):
+            def get_filter_index(min_scheduled, arrivals):
```
youngj: Is this the same as `np.searchsorted(arrivals, min_scheduled)` ?

models/metrics.py
```diff
@@ -72,6 +72,139 @@ def get_data_frame(self, d, stop_id=None, direction_id=None):
...
+
+                # arrivals are sorted, so this returns the index of the first arrival after the first scheduled arrival
+                while arrivals[first_index] < min_scheduled:
+                    first_index += 1
+
+                return first_index
+            
+            # scheduled arrivals are sorted, so the first element is the min
+            filtered_arrivals = arrivals[get_filter_index(scheduled_arrivals[0], arrivals):]
+            real_headways = arrival_headways[get_filter_index(scheduled_arrivals[0], arrivals):]
```
youngj: Add variable to reuse result of same computation from previous line

models/metrics.py
```diff
@@ -72,6 +72,139 @@ def get_data_frame(self, d, stop_id=None, direction_id=None):
...
+            next_arrival_time = []
+            next_headway = []
+            current_index = 0
+
+            for index in range(len(filtered_arrivals)):
```
youngj: this looks like `np.searchsorted(filtered_arrivals, scheduled_arrivals)` - searchsorted will return an array of indexes for each scheduled arrival, then `filtered_arrivals[arrival_indexes]` and `real_headways[arrival_indexes]` would get the arrival times and headways (may need to pad arrays with nan at the end)

models/metrics.py
```diff
@@ -72,6 +72,139 @@ def get_data_frame(self, d, stop_id=None, direction_id=None):
...
+        stop_timetable['next_arrival_headway'] = next_headways
+
+        # for each scheduled arrival time, get the closest actual arrival (earlier or later) and the corresponding headway
+        def get_closest_arrival_time(scheduled_arrivals, arrivals, arrival_headways):
+            closest_arrivals = []
+            closest_headway = []
+            current_schedule_index = 0
+            current_arrival_index = 0
+            previous_arrival_index = -1
+
```
youngj: I think this could be simplified to remove loops, roughly:
- use np.searchsorted to get the indexes of the next arrival times
- subtract 1 from those indexes to get the indexes of the previous arrival times
- create two arrays containing the absolute value difference between the schedule and next arrival time, and between the schedule and previous arrival time
- create a boolean array like (absolute differences to previous arrival time) < (absolute differences to next arrival time)
- use np.where to construct an array containing either the previous arrival time or the next arrival time depending on whether each value in the boolean array is true or false

jtanquil: Thanks, this makes the method a lot more concise!

jtanquil:
> When I run `python timetables.py --route=12 --stops=3476 --date=2019-04-12` it raises `AttributeError: ("'Series' object has no attribute 'start_date'", 'occurred at index 0')` on models/timetable.py line 96

I believe this was an issue with the path to the s3 bucket returned by `get_s3_bucket`, since I changed that to the bucket that Eddy created it should work now.

models/metrics.py
```diff
@@ -72,6 +73,84 @@ def get_data_frame(self, d, stop_id=None, direction_id=None):
...
+        stop_timetable['next_arrival_headway'] = next_headways
+
+        # for each scheduled arrival time, get the closest actual arrival (earlier or later) and the corresponding headway
+        def get_closest_arrival_time(scheduled_arrivals, arrivals, arrival_headways):
+            next_arrival_times = [np.searchsorted(arrivals, scheduled_arrival) for scheduled_arrival in scheduled_arrivals]
+            next_arrivals = [arrivals[next_arrival_times[i]] if next_arrival_times[i] < len(arrivals) else np.nan for i in range(len(next_arrival_times))]
+            next_headways = [arrival_headways[next_arrival_times[i]] if next_arrival_times[i] < len(arrival_headways) else np.nan for i in range(len(next_arrival_times))]
+            next_abs = [abs(scheduled_arrivals[i] - next_arrivals[i]) for i in range(len(scheduled_arrivals))]
+
+            previous_arrival_times = [index - 1 for index in next_arrival_times]
```
youngj: variable name should probably be previous_arrival_indexes. You can add/subtract a constant from each element of a numpy array without an explicit loop, like `next_arrival_indexes - 1` (assuming next_arrival_times renamed to next_arrival_indexes)

models/metrics.py
```diff
@@ -72,6 +73,84 @@ def get_data_frame(self, d, stop_id=None, direction_id=None):
...
+        next_arrival_deltas = next_arrivals - stop_timetable['arrival_time'].values
+        stop_timetable['next_arrival'] = next_arrivals
+        stop_timetable['next_arrival_delta'] = next_arrival_deltas
+        stop_timetable['next_arrival_headway'] = next_headways
+
+        # for each scheduled arrival time, get the closest actual arrival (earlier or later) and the corresponding headway
+        def get_closest_arrival_time(scheduled_arrivals, arrivals, arrival_headways):
+            next_arrival_times = [np.searchsorted(arrivals, scheduled_arrival) for scheduled_arrival in scheduled_arrivals]
+            next_arrivals = [arrivals[next_arrival_times[i]] if next_arrival_times[i] < len(arrivals) else np.nan for i in range(len(next_arrival_times))]
+            next_headways = [arrival_headways[next_arrival_times[i]] if next_arrival_times[i] < len(arrival_headways) else np.nan for i in range(len(next_arrival_times))]
```
youngj: next_arrivals and next_headways appear to be the same values returned from get_next_arrival_time, so get_next_arrival_time could be removed and you could return these values (since this version is simpler without needing loops)

models/metrics.py
```diff
@@ -72,6 +73,84 @@ def get_data_frame(self, d, stop_id=None, direction_id=None):
...
+            next_abs = [abs(scheduled_arrivals[i] - next_arrivals[i]) for i in range(len(scheduled_arrivals))]
+
+            previous_arrival_times = [index - 1 for index in next_arrival_times]
+            previous_arrivals = [arrivals[previous_arrival_times[i]] if previous_arrival_times[i] < len(arrivals) else np.nan for i in range(len(previous_arrival_times))]
+            previous_headways = [arrival_headways[previous_arrival_times[i]] if previous_arrival_times[i] < len(arrival_headways) else np.nan for i in range(len(previous_arrival_times))]
+            previous_abs = [abs(scheduled_arrivals[i] - previous_arrivals[i]) for i in range(len(scheduled_arrivals))]
```
youngj: Should be the same as `np.absolute(scheduled_arrivals - previous_arrivals)`

models/metrics.py
```diff
@@ -72,6 +73,84 @@ def get_data_frame(self, d, stop_id=None, direction_id=None):
...
+        next_arrivals, next_headways = get_next_arrival_time(stop_timetable['arrival_time'].values, stop_arrivals['TIME'].values, arrival_headways)
+        next_arrival_deltas = next_arrivals - stop_timetable['arrival_time'].values
+        stop_timetable['next_arrival'] = next_arrivals
+        stop_timetable['next_arrival_delta'] = next_arrival_deltas
+        stop_timetable['next_arrival_headway'] = next_headways
+
+        # for each scheduled arrival time, get the closest actual arrival (earlier or later) and the corresponding headway
+        def get_closest_arrival_time(scheduled_arrivals, arrivals, arrival_headways):
+            next_arrival_times = [np.searchsorted(arrivals, scheduled_arrival) for scheduled_arrival in scheduled_arrivals]
+            next_arrivals = [arrivals[next_arrival_times[i]] if next_arrival_times[i] < len(arrivals) else np.nan for i in range(len(next_arrival_times))]
```
youngj: To avoid loop and if condition, could do something like this:
```
arrivals_padded = np.r_[arrivals, np.nan]
next_arrivals = arrivals_padded[next_arrival_indexes]
```

models/metrics.py
```diff
@@ -72,6 +73,84 @@ def get_data_frame(self, d, stop_id=None, direction_id=None):
...
+            next_headways = [arrival_headways[next_arrival_times[i]] if next_arrival_times[i] < len(arrival_headways) else np.nan for i in range(len(next_arrival_times))]
+            next_abs = [abs(scheduled_arrivals[i] - next_arrivals[i]) for i in range(len(scheduled_arrivals))]
+
+            previous_arrival_times = [index - 1 for index in next_arrival_times]
+            previous_arrivals = [arrivals[previous_arrival_times[i]] if previous_arrival_times[i] < len(arrivals) else np.nan for i in range(len(previous_arrival_times))]
```
youngj: I think the previous index could be -1 which would return the wrong value when indexing into arrivals.
I think something like `np.r_[np.nan, arrivals][next_arrival_indexes]` would work without needing a list comprehension or if condition.

jtanquil:
Thanks for the feedback! Still getting used to using numpy but the code's a lot cleaner now.

jtanquil:
Thanks for the feedback! I fixed the remaining errors.

models/gtfs.py
```diff
@@ -167,9 +167,9 @@ def upload_to_s3(self, s3_path: str, df: pd.DataFrame):
...
         print(f"{datetime.now().time().isoformat()}: Uploaded {s3_path} to s3 bucket.")
 
-    def save_date_ranges(self, s3 = False):
+    def save_date_ranges(self, s3 = False, ver = "v1"):
         df = self.get_date_ranges()
-        filepath = f"{get_schedule_dir()}/date_ranges.csv"
+        filepath = f"{get_schedule_dir()}/date_ranges_{ver}.csv"
 
         if s3:
             s3_path = f"date_ranges.csv"
```
youngj: The S3 path also needs to add `_{ver}`

PR #96, opened 2019-05-09
Add metrics interval api

vicky11z:
Need an API that displays `metrics` by interval. If `start_time` and `end_time` are passed in, renders hourly metrics. Without any times attached, it shows times by the default 5 intervals that are shown on the front-end.

This code is really repetitive, still a WIP refactoring but wanted to get this API out there first.

### DIFF
--- a/constants.py
+++ b/constants.py
@@ -1,23 +1,27 @@
 from datetime import datetime
-DEFAULT_TIME_INTERVALS = [
+from pytz import timezone
+
+DEFAULT_TIME_STR_INTERVALS = [
   {
-    'start_time': datetime.strptime('03:00', '%H:%M'),
-    'end_time': datetime.strptime('07:00', '%H:%M')
+    'start_time': '03:00',
+    'end_time': '07:00'
   },
   {
-    'start_time': datetime.strptime('07:00', '%H:%M'),
-    'end_time': datetime.strptime('10:00', '%H:%M')
+    'start_time': '07:00',
+    'end_time': '10:00'
   },
   {
-    'start_time': datetime.strptime('10:00', '%H:%M'),
-    'end_time': datetime.strptime('16:00', '%H:%M')
+    'start_time': '10:00',
+    'end_time': '16:00'
   },
   {
-    'start_time': datetime.strptime('16:00', '%H:%M'),
-    'end_time': datetime.strptime('19:00', '%H:%M')
+    'start_time': '16:00',
+    'end_time': '19:00'
   },
   {
-    'start_time': datetime.strptime('19:00', '%H:%M'),
-    'end_time': datetime.strptime('03:00', '%H:%M')
+    'start_time': '19:00',
+    'end_time': '03:00+1'
   }
 ]
+
+PACIFIC_TIMEZONE = timezone('US/Pacific')
--- a/constants.py
+++ b/constants.py
@@ -1,0 +1,23 @@
+from datetime import datetime
+DEFAULT_TIME_INTERVALS = [
+  {
+    'start_time': datetime.strptime('03:00', '%H:%M'),
+    'end_time': datetime.strptime('07:00', '%H:%M')
+  },
+  {
+    'start_time': datetime.strptime('07:00', '%H:%M'),
+    'end_time': datetime.strptime('10:00', '%H:%M')
+  },
+  {
+    'start_time': datetime.strptime('10:00', '%H:%M'),
+    'end_time': datetime.strptime('16:00', '%H:%M')
+  },
+  {
+    'start_time': datetime.strptime('16:00', '%H:%M'),
+    'end_time': datetime.strptime('19:00', '%H:%M')
+  },
+  {
+    'start_time': datetime.strptime('19:00', '%H:%M'),
+    'end_time': datetime.strptime('03:00', '%H:%M')
+  }
+]
--- a/metrics-api.py
+++ b/metrics-api.py
@@ -125,7 +125,6 @@ def calc_metrics(request):
     waits = []
     if end_stop_id:
         completed_trips  = []
-
     for d in dates:
         try:
             history = arrival_history.get_by_date('sf-muni', route_id, d)
@@ -148,14 +147,14 @@ def calc_metrics(request):
             raise Exception(f"Arrival history not found for route {route_id} on {d.isoformat()}")
         except IndexError as ex:
             raise Exception(f"No arrivals found for stop {start_stop_id} on route {route_id} in direction {direction_id} on {d.isoformat()}")
-
+    
     headway_min = pd.concat(headway_min_arr)
     waits = pd.concat(waits)
     if end_stop_id and both_stops_same_dir:
         completed_trips = pd.concat(completed_trips)
-
     if headway_min.empty:
-        raise Exception(f"No arrivals for stop {start_stop_id} on route {route_id}")
+        # TODO: need a better error type than this
+        raise BufferError(f"No arrivals for stop {start_stop_id} on route {route_id}")
 
     data = {
         'params': params,
@@ -205,12 +204,15 @@ def add_intervals_to_data(arr, time_intervals):
     for time_interval in time_intervals:
         start_time = time_interval['start_time']
         end_time = time_interval['end_time']
+        req = create_calc_metrics_request(request, start_time, end_time)
         try:
-            req = create_calc_metrics_request(request, start_time, end_time)
+            curr_data = calc_metrics(req)
+        except BufferError:
+            # no trains for the current time interval
+            continue
         except Exception as ex:
             raise(Exception(ex))
 
-        curr_data = calc_metrics(req)
         arr.append({
             'start_time': start_time.strftime('%H:%M'),
             'end_time': end_time.strftime('%H:%M'),
@@ -267,7 +269,7 @@ def metrics_by_interval():
         end_time.replace(microsecond=0, second=0, minute=0) - timedelta(hours=1)
 
         hourly_time_intervals = []
-        while start_time != end_time:
+        while start_time.hour != end_time.hour:
             curr_interval = {}
             curr_interval['start_time'] = start_time
             curr_interval['end_time'] = start_time + timedelta(seconds=3600)
@@ -287,7 +289,6 @@ def metrics_by_interval():
                 'params': params,
                 'error': 'Need both a start and end time'
             }, indent=2), status=404, mimetype='application/json')
-        print('here')
         try:
             add_intervals_to_data(data['intervals'], constants.DEFAULT_TIME_INTERVALS)
         except Exception as ex:
--- a/metrics-api.py
+++ b/metrics-api.py
@@ -209,7 +209,7 @@ def metrics_by_interval():
 
     return Response(json.dumps(data, indent=2), mimetype='application/json')
 
-def calc_metrics(args, route_config):
+def calc_metrics(args: dict, route_config: nextbus.RouteConfig) -> dict:
     route_id = args['route_id']
     start_stop_id = args['start_stop_id']
     if start_stop_id is None:
@@ -324,7 +324,7 @@ def calc_metrics(args, route_config):
     }
     return data
 
-def create_intervals_list(time_intervals, params, route_config):
+def create_intervals_list(time_intervals: dict, params: dict, route_config: nextbus.RouteConfig) -> list:
     intervals_arr = []
     for time_interval in time_intervals:
         params['start_time_str'] = time_interval['start_time']
--- a/metrics-api.py
+++ b/metrics-api.py
@@ -56,11 +56,10 @@ def route_config():
 @app.route('/metrics', methods=['GET'])
 def metrics_page():
     metrics_start = time.time()
-
     calc_metrics_args = {
         'start_stop_id': request.args.get('start_stop_id'),
         'end_stop_id': request.args.get('end_stop_id'),
-        'route_id': request.args.get('route_id'),
+        'route_id': request.args.get('route_id') or '12',
         'direction_id': request.args.get('direction_id'),
         'start_date_str': request.args.get('start_date'),
         'end_date_str': request.args.get('end_date'),
@@ -68,8 +67,9 @@ def metrics_page():
         'start_time_str': request.args.get('start_time'),
         'end_time_str': request.args.get('end_time'),
     }
+    route_config = nextbus.get_route_config('sf-muni', calc_metrics_args['route_id'])
     try:
-        data = calc_metrics(calc_metrics_args)
+        data = calc_metrics(calc_metrics_args, route_config)
     except FileNotFoundError as ex:
         return Response(json.dumps({
             'params': calc_metrics_args,
@@ -134,6 +134,8 @@ def metrics_by_interval():
         'end_time_str': end_time_str,
     }
 
+    route_config = nextbus.get_route_config('sf-muni', params['route_id'])
+
     data = {'intervals': []}
 
     if start_time_str is not None and end_time_str is not None:
@@ -158,7 +160,7 @@ def metrics_by_interval():
         
         hourly_time_intervals = constants.DEFAULT_TIME_STR_INTERVALS
     try:
-        add_intervals_to_data(data['intervals'], hourly_time_intervals, params)
+        add_intervals_to_data(data['intervals'], hourly_time_intervals, params, route_config)
     except FileNotFoundError as ex:
         return Response(json.dumps({
             'params': params,
@@ -180,8 +182,6 @@ def metrics_by_interval():
             'error': str(ex),
         }, indent=2), status=400, mimetype='application/json')
 
-
-    route_config = nextbus.get_route_config('sf-muni', route_id)
     start_stop_info = route_config.get_stop_info(start_stop_id)
     end_stop_info = route_config.get_stop_info(end_stop_id) if end_stop_id else None
 
@@ -208,10 +208,8 @@ def metrics_by_interval():
 
     return Response(json.dumps(data, indent=2), mimetype='application/json')
 
-def calc_metrics(args):
+def calc_metrics(args, route_config):
     route_id = args['route_id']
-    if route_id is None:
-        route_id = '12'
     start_stop_id = args['start_stop_id']
     if start_stop_id is None:
         start_stop_id = '3476'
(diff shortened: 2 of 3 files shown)


metrics-api.py
```diff
@@ -153,26 +144,17 @@ def metrics_page():
...
     if end_stop_id and both_stops_same_dir:
         completed_trips = pd.concat(completed_trips)
-
     if headway_min.empty:
-        return Response(json.dumps({
-            'params': params,
-            'error': f"No arrivals for stop {start_stop_id} on route {route_id}",
-        }, indent=2), status=404, mimetype='application/json')
+        # TODO: need a better error type than this
+        raise BufferError(f"No arrivals for stop {start_stop_id} on route {route_id}")
```
vicky11z: So we also reach this error if there are no buses running for a given timezone - I don't want to cancel the entire request in these cases (since the buses will always stop running at a certain time) so I'm raising a specific type of error. Not sure what the best one to throw was here so just raised a `BufferError` but definitely open to suggestions.

metrics-api.py
```diff
@@ -92,10 +90,7 @@ def metrics_page():
     try:
         dates = util.get_dates_in_range(start_date_str, end_date_str)
     except Exception as ex:
```
tylerhou: I think this is a bit dangerous -- you might end up catching exceptions you didn't actually expect. You're also changing the API now because all exceptions return 400 instead of sometimes 404.

A solution which would fix both of these issues is to create your own exception classes. Then you can catch those specific exceptions and switch between 400/404 in this handler.

I think Python style guide suggests naming them "___Error", but I'm not 100% sure.

metrics-api.py
```diff
@@ -185,12 +167,165 @@ def metrics_page():
         'trip_times': metrics.get_trip_times_stats(completed_trips, start_stop_id, end_stop_id)
             if end_stop_id and both_stops_same_dir else None,
     }
+    return data
+
+def create_calc_metrics_request(request, start_time=None, end_time=None):
```
tylerhou: `serialize_request`? `create_request` sounds like you're creating a request to send off to another server.

metrics-api.py
```diff
@@ -185,12 +167,165 @@ def metrics_page():
...
+def metrics_page():
+    metrics_start = time.time()
+
+    req = create_calc_metrics_request(request)
+    try:
+        data = calc_metrics(req)
+    except Exception as ex:
+        return Response(json.dumps({
+            'params': req,
+            'error': str(ex),
```
tylerhou: Is this safe? You might accidentally leak information (some exceptions should not be serialized and sent to all users).

metrics-api.py
```diff
@@ -185,12 +167,165 @@ def metrics_page():
...
+        start_time = time_interval['start_time']
+        end_time = time_interval['end_time']
+        req = create_calc_metrics_request(request, start_time, end_time)
+        try:
+            curr_data = calc_metrics(req)
+        except BufferError:
+            # no trains for the current time interval
+            continue
+        except Exception as ex:
+            raise(Exception(ex))
```
tylerhou: I don't think you need these two lines (because you're just catching and re-raising).

metrics-api.py
```diff
@@ -185,12 +167,165 @@ def metrics_page():
...
+            'headway_min': curr_data['headway_min'],
+            'wait_times': curr_data['wait_times'],
+            'trip_times': curr_data['trip_times'],
+        })
+
+@app.route('/metrics_by_interval', methods=['GET'])
+def metrics_by_interval():
+    route_id = request.args.get('route_id')
+    if route_id is None:
+        route_id = '12'
```
tylerhou: What are these constants?

metrics-api.py
```diff
@@ -185,12 +167,165 @@ def metrics_page():
...
+
+    start_date_str = request.args.get('start_date')
+    end_date_str = request.args.get('end_date')
+    date_str = request.args.get('date')
+
+    if date_str is not None:
+        start_date_str = end_date_str = date_str
+    else:
+        if start_date_str is None:
+            start_date_str = '2019-02-01'
```
tylerhou: Why this constant?

metrics-api.py
hathix: oh, these are just placeholders -- if you don't specify a route, we'll default to Bus Route 12 for demo's sake. Don't worry about it for now.

constants.py
```diff
@@ -0,0 +1,23 @@
+from datetime import datetime
```
hathix: Having a constants file is great! We should put more stuff in here later.

metrics-api.py
```diff
@@ -50,23 +51,20 @@ def route_config():
...
-def metrics_page():
-    metrics_start = time.time()
-
-    route_id = request.args.get('route_id')
+def calc_metrics(request):
+    route_id = request['route_id']
     if route_id is None:
         route_id = '12'
-    start_stop_id = request.args.get('start_stop_id')
+    start_stop_id = request['start_stop_id']
```
hathix: It's probably better to stick with the old `args.get` syntax, because the brackets only work if you know for sure the key exists ([see here](https://stackoverflow.com/questions/10434599/how-to-get-data-received-in-flask-request)). But in this case we can't be guaranteed that the user will pass the key. So the `args.get` syntax (which is more robust) should be preferred.

metrics-api.py
```diff
@@ -185,12 +167,165 @@ def metrics_page():
...
+        'date': request.args.get('date'),
+    }
+
+@app.route('/metrics', methods=['GET'])
+def metrics_page():
+    metrics_start = time.time()
+
+    req = create_calc_metrics_request(request)
+    try:
+        data = calc_metrics(req)
```
hathix: Great! I'm glad you extracted a lot of the business logic out into a helper function.

metrics-api.py
```diff
@@ -185,12 +167,165 @@ def metrics_page():
...
+            }, indent=2), status=400, mimetype='application/json')
+
+
+    route_config = nextbus.get_route_config('sf-muni', route_id)
+    start_stop_info = route_config.get_stop_info(start_stop_id)
+    end_stop_info = route_config.get_stop_info(end_stop_id) if end_stop_id else None
+
+    if direction_id is not None:
+        dir_info = route_config.get_direction_info(direction_id)
+    if dir_info is not None:
```
jtanquil: This line will throw an error if `direction_id` is `None`, since `dir_info` will never be assigned a value (happens when you try `http://localhost:5000/metrics_by_interval?start_date=2019-04-08`)

metrics-api.py
jtanquil: We should change the demo date. I believe the stored data we have now is from April and not February - not entirely sure, just that `2019-02-01` has no arrival data and `2019-04-08` does (@youngj can clarify)

metrics-api.py
```diff
@@ -185,12 +167,165 @@ def metrics_page():
...
+    }
+
+    data = {'intervals': []}
+
+    if start_time_str is not None and end_time_str is not None:
+        # round start_time down and end_time up to allow for even intervals
+        start_time = datetime.strptime(start_time_str, '%H:%M')
+        start_time.replace(microsecond=0, second=0, minute=0)
+        end_time = datetime.strptime(end_time_str, '%H:%M')
+        end_time.replace(microsecond=0, second=0, minute=0) - timedelta(hours=1)
```
jtanquil: This line doesn't actually change `end_time` since `end_time.replace` returns another `datetime` instance, similarly with line 267.

metrics-api.py
```diff
@@ -185,12 +167,165 @@ def metrics_page():
...
+        except Exception as ex:
+            return Response(json.dumps({
+                'params': params,
+                'error': str(ex),
+            }, indent=2), status=400, mimetype='application/json')
+
+
+    route_config = nextbus.get_route_config('sf-muni', route_id)
+    start_stop_info = route_config.get_stop_info(start_stop_id)
+    end_stop_info = route_config.get_stop_info(end_stop_id) if end_stop_id else None
```
jtanquil: Would it be possible to use the `RouteConfig` object in `calc_metrics` to get the relevant information instead of creating an identical `RouteConfig` instance?

constants.py
```diff
@@ -0,0 +1,23 @@
...
+    'start_time': datetime.strptime('10:00', '%H:%M'),
+    'end_time': datetime.strptime('16:00', '%H:%M')
+  },
+  {
+    'start_time': datetime.strptime('16:00', '%H:%M'),
+    'end_time': datetime.strptime('19:00', '%H:%M')
+  },
+  {
+    'start_time': datetime.strptime('19:00', '%H:%M'),
+    'end_time': datetime.strptime('03:00', '%H:%M')
```
exxonvaldez: This end time needs to represent 3am the following day.  As written I think the interval is basically empty since "end" is before "start".

metrics-api.py
```diff
@@ -185,12 +167,165 @@ def metrics_page():
...
 
+def add_intervals_to_data(arr, time_intervals):
+    for time_interval in time_intervals:
+        start_time = time_interval['start_time']
+        end_time = time_interval['end_time']
+        req = create_calc_metrics_request(request, start_time, end_time)
+        try:
+            curr_data = calc_metrics(req)
+        except BufferError:
+            # no trains for the current time interval
```
exxonvaldez: If there are no vehicles, the interval should still be in the returned array of intervals, just empty.  We can support displaying an error message also as a separate part of the returned data.

constants.py
youngj: Elsewhere we handle this issue by storing times of day as strings in the format "03:00+1" to represent 3am on the next day. `util.get_localized_datetime` will return the correct datetime when given a time string in this format. It seems that `DEFAULT_TIME_INTERVALS` could contain time strings in this format, instead of datetimes.

metrics-api.py
youngj: It may also be a little confusing to use the term `request` to mean something other than the flask request (which is also the first parameter). However, it's probably not necessary to have a separate function for this - could just pass the params into calc_metrics directly as kwargs, and repeat the 7 `request.args.get` lines in both places.

metrics-api.py
youngj: That's correct

metrics-api.py
youngj: Can use `util.get_localized_datetime` for this

metrics-api.py
```diff
@@ -185,12 +167,165 @@ def metrics_page():
...
+
+        hourly_time_intervals = []
+        while start_time.hour != end_time.hour:
+            curr_interval = {}
+            curr_interval['start_time'] = start_time
+            curr_interval['end_time'] = start_time + timedelta(seconds=3600)
+            hourly_time_intervals.append(curr_interval)
+            start_time += timedelta(seconds=3600)
+        try:
+            add_intervals_to_data(data['intervals'], hourly_time_intervals)
```
youngj: Could factor this try/except block and similar block starting at line 292 after if/else to avoid duplication

hathix:
@vicky11z , let's get this fixed up and merged soon -- #101 is waiting on this.

metrics-api.py
hathix: Can you just return an empty dataframe (or whatever data structure you use here)? I think this isn't a case when you should throw an error; it's pretty common for there to be no buses in a given time bound so it would make more sense to just say "nothing found" than give an error. Just my $0.02.

hathix: If this debate is too much to resolve now, you can make an issue and we can revisit it later.

metrics-api.py
hathix: yes, agree

metrics-api.py
hathix: Eh, we don't have any sensitive data, so I wouldn't worry about leaking anything. For debugging purposes this is fine. No need to over-engineer.

metrics-api.py
vicky11z: So I actually changed this method so that `request` is a dictionary that's passed in, not the actual request object. I think you do make a good point that this is a confusing name because it's not a real request. Too bad Python doesn't have explicit typing 😢

metrics-api.py
vicky11z: @youngj so in my next edit I'm actually converting these values back to strings (lines 272-277). I'm just creating these datetimes so I can perform time arithmetic. Opted not to use `util.get_localized_datetime`. If you know of a better way to do this than converting from string -> datetime -> back to string, that would be great. It's pretty ugly otherwise

metrics-api.py
vicky11z: Synced offline, going to return the time and a message like `No trains running at this time interval`.

metrics-api.py
vicky11z: Yeah that's a good idea - changed it to return null values

metrics-api.py
vicky11z: I think the intention is for this API to just be used internally.

metrics-api.py
```diff
@@ -152,44 +289,61 @@ def metrics_page():
...
+            'direction_id': params['direction_id'],
+            'start_date_str': params['start_date_str'],
+            'end_date_str': params['end_date_str'],
+            'start_time_str': time_interval['start_time'],
+            'end_time_str': time_interval['end_time'],
+        }
+        try:
+            curr_data = calc_metrics(calc_metrics_args)
+
+        except Exception as ex:
```
youngj: Is there an advantage to raising a generic Exception vs removing the try/except block and sending the original exception back to the caller?

metrics-api.py
youngj: Agreed - would be best to only call `nextbus.get_route_config` once. Could pass it as a parameter to calc_metrics. Maybe not necessary to change in this PR.

metrics-api.py
youngj: /metrics_by_interval may not actually need to return metadata about directions, stop titles, and the route title. I think I added these fields to /metrics because it was helpful for humans viewing the API endpoint directly (before we had a functioning React app), but they probably aren't actually needed or used by the React app.

metrics-api.py
vicky11z: Good point - going to remove.

vicky11z:
Note: I realized that this endpoint doesn't account for cross-date values. Like if you were to enter `start_date` 4/08 and `end_date` 4/09, and times `23:00` to `02:00`. It wouldn't calculate the correct errors for hourly intervals - for example, it would calculate `01:00` to `02:00` from 4/08 to 4/09. I can write that down as a fast follow for this PR (since we don't have multiple dates yet) or I can add that into this PR if people feel like it's necessary, but it feels like this needs to be merged to unblock Terrence's PR.

cc @exxonvaldez

youngj:
OK with me to implement that in a separate PR. Note that since compute_arrivals.py associates data between midnight to 3am with the previous day, that example would have the parameters start_date = 2019-04-08 , end_date = 2019-04-08 , start_time = 23:00 , and end_time = 02:00+1 .

metrics-api.py
```diff
@@ -126,18 +265,16 @@ def metrics_page():
...
     directions = [{'id': dir_info.id, 'title': dir_info.title} for dir_info in dir_infos]
 
     headway_min_arr = []
     waits = []
     if end_stop_id:
         completed_trips  = []
-
     for d in dates:
         try:
             history = arrival_history.get_by_date('sf-muni', route_id, d)
```
youngj: Since this is called in a loop, the arrival history for each date will be read from the file system several times, before being filtered by start/end time. Not sure how much of an impact it will have on performance, but probably worth profiling and optimizing in a later PR.

metrics-api.py
vicky11z: @exxonvaldez want to check if you're using the metadata or just the intervals?

exxonvaldez: Front end is not using the metadata here like dir_info, but the basic metadata is just good as a sanity check that we're getting the data that was requested and not some other data.  I guess you can make direction_id "required" and the request can fail if it wasn't given?  Otherwise you'd have to determine a default direction in the same way that the UI does.

metrics-api.py
```diff
@@ -83,33 +128,131 @@ def metrics_page():
...
+    data['end_stop_title'] = end_stop_info.title if end_stop_info else None
+    data['directions'] = directions
+
+    return Response(json.dumps(data, indent=2), mimetype='application/json')
+
+def calc_metrics(args, route_config):
+    route_id = args['route_id']
+    start_stop_id = args['start_stop_id']
+    if start_stop_id is None:
+        start_stop_id = '3476'
```
youngj: Currently when /metrics returns the params object, it doesn't show any default values set in calc_metrics, e.g http://localhost:5000/metrics returns:
```
{
  "params": {
    "start_stop_id": null,
    "end_stop_id": null,
    "route_id": "12",
    "direction_id": null,
    "start_date_str": null,
    "end_date_str": null,
    "date": null,
    "start_time_str": null,
    "end_time_str": null
  },
  "error": "Arrival history not found for route 12 on None"
}
```

Would be good to set default values for start_stop_id and start_date_str in metrics_page instead of calc_metrics (like how metrics_page sets the default route_id to 12) for consistency, or just remove all default values.

metrics-api.py
```diff
@@ -83,33 +128,131 @@ def metrics_page():
...
         }, indent=2), status=400, mimetype='application/json')
 
-    tz = pytz.timezone('US/Pacific')
-
-    route_config = nextbus.get_route_config('sf-muni', route_id)
     start_stop_info = route_config.get_stop_info(start_stop_id)
     end_stop_info = route_config.get_stop_info(end_stop_id) if end_stop_id else None
 
+    if direction_id is not None:
+        dir_info = route_config.get_direction_info(direction_id)
```
youngj: dir_info is undefined if direction_id is none, which causes a server error when loading /metrics_by_interval . Could remove code that returns direction metadata since the API clients probably won't use it.

metrics-api.py
```diff
@@ -83,33 +152,102 @@ def metrics_page():
...
-        'end_date': end_date_str,
-        'start_time': start_time_str,
-        'end_time': end_time_str,
+        'start_date_str': start_date_str,
+        'end_date_str': end_date_str,
+        'start_time_str': start_time_str,
+        'end_time_str': end_time_str,
     }
 
+    if direction_id is None:
```
youngj: I don't think direction_id is actually required (/metrics doesn't require it). It seems inconsistent to return an error if direction_id is None, while setting default values for route_id and start_stop_id. It seems to work fine if this block is removed, so I'd probably just remove it.

PR #97, opened 2019-05-09
remove a package.lock that broke the build

hathix:
Somehow a package lock got checked in that broke the build. Removing that


hathix:
nvm

PR #100, opened 2019-05-09
Beef up gitignore to include MacOS

hathix:
Just improving the gitignore to include more important lines so we don't accidentally include anything.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,75 @@
 
-# Created by https://www.gitignore.io/api/node,macos,flask,python,intellij
-# Edit at https://www.gitignore.io/?templates=node,macos,flask,python,intellij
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -334,22 +379,13 @@ typings/
...
-# End of https://www.gitignore.io/api/node,flask,python,intellij
-#
-# idea
-.idea
-# theia
-.theia
-# test notebook
-/models/Untitled.ipynb
 
-# Eclipse
```
exxonvaldez: Can you retain this custom section for Eclipse?  I added it.  Adding Mac to the generator is great.

hathix: Ah, good call. I just pushed a commit that added Eclipse to the generator.

PR #101, opened 2019-05-14
Performance across intervals of the day chart

exxonvaldez:
Note:  this front-end PR must wait on https://github.com/trynmaps/metrics-mvp/pull/96 , the new backend endpoint.

- Add react-vis javascript charting package via package.json and css import.  Appears to change package-lock.json.
- Update action creator, reducers, and components to call new backend endpoint and push "intervalData" down to new component.
- Control panel creates slightly different params for new endpoint (no start/end time for now, using default intervals).
- Move javascript helper function for getting 90th percentile value to new helpers file.
- Embed new InfoIntervalsOfDay component within Info component

Screen Shot 2019-05-14 at 11 30 48 AM

Added Vicky and Yi to keep them in the loop, Jim for action/reducer/props stuff.


hathix:
Great! Few thoughts:

* What's "average" vs "planning"? We should clarify that somewhere in the UI.
* Don't worry too much about package-lock.json; we can rebuild that if it's gone funky.
* Anything else we should put in the helpers file? If so, let's just file issues for them (no need to fix it in this PR).

frontend/src/actions/index.js
```diff
@@ -2,6 +2,13 @@ import axios from 'axios';
 
 import { metricsBaseURL } from '../config';
 
+export function fetchData (graphParams, intervalParams) {
+  return function (dispatch) {
+      dispatch(fetchGraphData(graphParams));
```
hathix: Is this basically a promise? Worth calling out how this function is supposed to be used (like giving a code snippet with an example of how to use it) because it's not immediately obvious.

frontend/src/components/Info.jsx
```diff
@@ -85,7 +73,7 @@ class Info extends Component {
     
     let travelVarianceTime = 0;
     if (tripTimes) {
-        travelVarianceTime = this.getPercentileValue(tripTimes, 90) - tripTimes.avg;
+        travelVarianceTime = getPercentileValue(tripTimes, 90) - tripTimes.avg;
```
hathix: can we not hardcode in 90 as a magic number? Ideally it would be either a parameter to this function or a named constant somewhere.

frontend/src/components/Info.jsx
```diff
@@ -227,7 +215,7 @@ class Info extends Component {
             <tr><th>Metric</th><th>Value</th><th>Grade</th><th>Score</th></tr>
             <tr>
             <td>Average wait</td><td>{Math.round(waitTimes.avg)} minutes<br/>
-            90% of waits under { Math.round(this.getPercentileValue(waitTimes, 90)) } minutes
+            90% of waits under { Math.round(getPercentileValue(waitTimes, 90)) } minutes
```
hathix: Can we explain somewhere why we're doing 90% (instead of 95%)? I recall you mentioned it's a best practice in the industry, but explain that here

frontend/src/components/InfoIntervalsOfDay.jsx
```diff
@@ -0,0 +1,272 @@
+import React, { Component } from 'react';
+import { XYPlot, HorizontalGridLines, XAxis, YAxis,
+    VerticalBarSeries, ChartLabel, Crosshair } from 'react-vis';
+import DiscreteColorLegend from 'react-vis/dist/legends/discrete-color-legend';
+import Form from 'react-bootstrap/Form';
+import Card from 'react-bootstrap/Card';
+import { getPercentileValue } from '../helpers/graphData'; 
+
+/**
+ * Bar chart of average and 90th percentile wait and time across the day.
```
hathix: This comment is super useful! It really helps me understand what I'm looking at here.

frontend/src/components/InfoIntervalsOfDay.jsx
```diff
@@ -0,0 +1,272 @@
...
+  }
+  
+  render() {
+    const { intervalData, intervalError } = this.props;
+
+    const intervals = intervalData ? intervalData.intervals : null; 
+    this.waitData = intervals ? intervals.map(this.mapInterval("wait_times")) : null;
+    this.tripData = intervals ? intervals.map(this.mapInterval("trip_times")) : null; 
+
+    const chartColors = ["#a4a6a9", "#aa82c5"]; // placeholder colors: gray and purple from nyc busstats
```
hathix: Can we put these in a constants file? I imagine we'll want to change the colors across our UI later as we develop a more cohesive brand identity

frontend/src/actions/index.js
exxonvaldez: Added a comment to clarify.  It's a React way of triggering two API calls at once.

frontend/src/components/Info.jsx
exxonvaldez: Created a new UIConstants.js file to hold this value for now.

frontend/src/components/InfoIntervalsOfDay.jsx
exxonvaldez: Moved to UIConstants.js.

frontend/src/components/Info.jsx
exxonvaldez: The 90% is fairly arbitrary, meaning "most waits" or "most trip times".  I added comments in UIConstants.h saything this about the "90" constant.

hathix:
Also #96 was just merged so we're good to merge thsi

PR #111, opened 2019-05-16
Update the favicon

hathix:


hathix:
Just tested it on Gitpod. It works but the icon doesnt look great when it gets scaled to 16x16. will add an issue to make a proper 16x16 icon with pixel art

PR #115, opened 2019-05-18
update docker-compose to use node_modules folder created when docker container originally built

youngj:
After switching from yarn to npm, `docker-compose up` has been taking a very long time to start the react container, due to the call to `npm install`, which sometimes seems to hang forever. (Also tried `npm ci` but it didn't seem to help.)

With this change, the react docker container uses the package.json, package.lock, and node_modules from when the docker image was originally built (which happens the first time you run `docker-compose up`), and only the `/frontend/{build,public,src}` subdirectories are shared between the host and the container. This makes it no longer necessary to run `npm install` during `docker-compose up`, so now the react container starts quickly (~5 seconds).

The downside is that any time node dependencies are changed in package.json, the docker container must be rebuilt (via `docker-compose build`). If a new node module is added by someone else, all other developers using old docker containers will get errors when trying to run code referencing the new module until they rebuild their docker containers. This will be annoying, but hopefully we won't be changing the dependencies in package.json very often. I think overall it will save time since developers run `docker-compose up` run much more often, and it will be much faster after this change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,9 @@
 # Created by https://www.gitignore.io/api/node,macos,flask,python,eclipse,intellij
 # Edit at https://www.gitignore.io/?templates=node,macos,flask,python,eclipse,intellij
 
+data/*.json
+data/*.csv
+
 ### Eclipse ###
 .metadata
 bin/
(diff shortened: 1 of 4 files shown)


hathix:
Do any changes need to be made to the readme? That is do developers need to change their usual workflow (in which case we should remind them)?

PR #119, opened 2019-05-26
Initial predictions model

vicky11z:
Wanted to get a Prediction(s) model down for comments - this pulls data from the nextbus predictions api. Eventually I want it to cache/store data similarly to how `RouteConfig` data is stored. This is a WIP but please leave comments & suggestions.

### DIFF
--- a/models/nextbus_predictions.py
+++ b/models/nextbus_predictions.py
@@ -2,120 +2,29 @@ import re
 import requests
 import json
 import os
-from xml.etree import ElementTree
 from datetime import datetime
 
 from models import predictions as p, nextbus
 
 STOPS_STR = '&stops='
 
-# when we try to get all routes
-'''
-Generate a list of strings in the format '&stops={route_id}|{stop_id}...'
-Since we can't substitute every stop on every route into an api call at once,
-segment the number of strings created by num_stops_per_route
-'''
-
-
-def get_routes_to_stops_str(agency_id: str, num_stops_per_str: int) -> str:
-    route_ids = [route.id for route in nextbus.get_route_list(agency_id)]
-    route_id_to_stop_ids = {}
-    for route_id in route_ids:
-        route_config = nextbus.get_route_config(agency_id, route_id)
-        route_id_to_stop_ids[route_id] = route_config.get_stop_ids()
-
-    route_strs = []
-    num_stops_so_far = 0
-    route_str = ''
-    for route_id in route_ids:
-        ret = gen_stop_str_for_route_limited(
-            route_id, route_id_to_stop_ids[route_id], num_stops_so_far)
-        route_str += ret['stop_str']
-        num_stops_so_far += ret['num_stops_so_far']
-        if num_stops_so_far >= num_stops_per_str:
-            route_strs.append(route_str)
-            route_str = ''
-            num_stops_so_far = 0
-    return route_strs
-
-
-def gen_stop_str_for_route_limited(route_id: str, stop_ids: list, num_stops_so_far: int) -> dict:
-    num_stops_so_far += len(stop_ids)
-    all_stops = STOPS_STR
-    stops = [f"{route_id}|{stop_id}" for stop_id in stop_ids]
-    all_stops += STOPS_STR.join(stops)
-    return {'stop_str': all_stops, 'num_stops_so_far': num_stops_so_far}
-
-def create_predictions_requests_limited(agency_id: str) -> list:
-    # TODO: error if response includes error
-
-    # we can maximally add ~400 stops to the string before API complains it's too long
-    multi_stops = get_routes_to_stops_str(agency_id, 400)
-    requests = []
-    for stops in multi_stops:
-        requests.append(
-            f"http://webservices.nextbus.com/service/publicJSONFeed?command=predictionsForMultiStops&a={agency_id}{stops}")
-    return requests
-
-def get_prediction_data_all(agency_id: str):
-    queried_time = datetime.now()
-    if re.match(r'^[\w\-]+$', agency_id) is None:
-        raise Exception(f"Invalid agency id: {agency_id}")
-
-    # TODO: cache responses
-    route_id_to_predictions = {}
-    for request_url in create_predictions_requests_limited(agency_id):
-        response = requests.get(request_url)
-        if response.status_code != 200:
-            response.raise_for_status()
-
-        route_id_to_predictions.update(
-            parse_prediction_response(queried_time, response))
-    return route_id_to_predictions
-
-
-# just trying to get string for one route's stops
-def gen_stop_str_for_route(route_id: str, stop_ids: list) -> str:
-    all_stops = STOPS_STR
-    stops = [f"{route_id}|{stop_id}" for stop_id in stop_ids]
-    all_stops += STOPS_STR.join(stops)
-    return all_stops
-
-def get_route_id_to_stops(agency_id: str):
-    route_ids = [route.id for route in nextbus.get_route_list(agency_id)]
-    route_id_to_stop_ids = {}
-    for route_id in route_ids:
-        route_config = nextbus.get_route_config(agency_id, route_id)
-        route_id_to_stop_ids[route_id] = route_config.get_stop_ids()
-    return route_id_to_stop_ids
-
-
-def create_predictions_requests(agency_id: str) -> dict:
-    route_id_to_stops = get_route_id_to_stops(agency_id)
-    
-    # gen prediction str for each route
-    route_to_request = {}
-    for route in route_id_to_stops:
-        stop_str = gen_stop_str_for_route(route, route_id_to_stops[route])
-        request = f"http://webservices.nextbus.com/service/publicJSONFeed?command=predictionsForMultiStops&a={agency_id}{stop_str}"
-        route_to_request[route] = request
-    return route_to_request
+def create_prediction_request_for_route(agency_id: str, route_id: str, stops: list) -> str:
+    stops = [f"{route_id}|{stop_id}" for stop_id in stops]
+    stop_str = STOPS_STR + STOPS_STR.join(stops)
+    return f"http://webservices.nextbus.com/service/publicJSONFeed?command=predictionsForMultiStops&a={agency_id}{stop_str}"
 
 # returns { route_id: [predictions] }
 def get_prediction_data_for_request(request_url: str) -> dict:
-    queried_time = datetime.now()
-
-    # TODO: cache responses
+    queried_time = datetime.now().timestamp()
     response = requests.get(request_url)
     if response.status_code != 200:
         response.raise_for_status()
     return parse_prediction_response(queried_time, response.json())
 
-def parse_prediction_response(queried_time: datetime, resp_json: dict) -> dict:
-    route_id_to_predictions = {}
+def parse_prediction_response(queried_time, resp_json: dict) -> list:
+    predictions_for_route = []
     predictions_by_route = resp_json['predictions']
     for prediction_info in predictions_by_route:
-        # print(prediction_info)
         route_id = prediction_info['routeTag']
         stop_id = prediction_info['stopTag']
         if 'direction' not in prediction_info:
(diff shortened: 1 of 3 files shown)


youngj:
Thanks for looking into this! Is there a plan for how predictions data would be used in the app? Since NextBus only provides an API for predictions per stop, I'm worried that Nextbus would block our IP addresses if we frequently pull predictions for all ~6000 stops. The nice thing about the NextBus vehicleLocations API is we only need to make 1 API request to get the data for all routes. 

- the code in nextbus_predictions.py could be added to nextbus.py rather than adding a new file
- it's much easier to use Nexbus's undocumented JSON API instead of their documented XML API -- just replace publicXMLFeed with publicJSONFeed

vicky11z:
@youngj that's a really good point. @exxonvaldez and I chatted about this a little bit, we were thinking it would be interesting to have this data to visualize the accuracy of predictions and allow us to rate metro lines with this as an additional data point. I think it gives us a unique metric to compare lines with - arrival data gives us an idea of the current movement of trains, the schedule allows us to see how much expected vs actual times differ, but predictions allow us to adjust expectations based on day-to-day delays (e.g. the schedule will be thrown off drastically if a bus breaks down, but predictions will show us how the schedule adjusts in this case). Terence was talking about Mary (?) charts to help visualize this data.  

@exxonvaldez could you add some of your thoughts as well?

vicky11z:
Also found this endpoint - http://webservices.nextbus.com/service/publicXMLFeed?command=predictionsForMultiStops&a=<agency_tag>&stops=<stop 1>&stops=<stop 2>&stops=<stop 3> 
that might help us minimize some of the calls since we can call all stops & trains in this way instead of each one individually.

exxonvaldez:
From an end user perspective, the question is "how much can I trust the predictions?"  From my own experience, I know that predictions for stops near the start of a line are much less accurate, and I'm not sure if they can be trusted at all.

Prediction accuracy includes:

- How often does the vehicle arrive as predicted or early or late?
- How often does a predicted vehicle disappear, or a new vehicle suddenly appear?
- How often do predictions not decrease linearly with time?

Assuming we can scrape the predictions, my main concern is that the predictions are just another view on the location data.  That is, they just take the vehicle locations and use the schedule to determine how long it should take between stops, doing some interpolation as needed.  They might naively assume vehicles turn around immediately at the ends of lines (when in practice they might go out of service or take a lengthy break), and they might assume vehicles will just appear at the start of lines at the right intervals.

If it turns out the predictions are actually interesting in their own right and more than just vehicle locations with schedules applied to them, then it should be possible to generate a bunch of stats as well as Marey chart-like plots of how predictions and vehicles linearly converge on stops over time.  Or where something deviates from that.

EddyIonescu:
Thanks for working on this, as this has been a request for a quite a while (see https://github.com/trynmaps/orion/issues/32, @frhino was especially interested in this). The initial idea was to add it in Orion (it's in nodejs, but you're welcome to add a new python script that stores everything in S3 as Orion currently does), which is what's currently fetching all the vehicle locations. However, we'd probably want to deploy it separately given the risk of having the IP blocked (as to not prevent Orion from updating vehicle locations).

Also, the predictions have an `isAffectedByLayover` field (unfortunately there aren't any transit apps that show this, but it'd be really useful), which should address the issue that predictions tend to only really be accurate for vehicles already on their trip (this is mostly due to the operator shortage, where supervisors shift terminal departure times on-the-fly in order to reduce service gaps, as well as operators starting early or late because they feel like it).

vicky11z:
@youngj or @jtanquil could you take a look at this when you have the time? it's ready for review finally 🎉

models/nextbus_predictions.py
```diff
@@ -0,0 +1,130 @@
...
+
+'''
+Generate a list of strings in the format '&stops={route_id}|{stop_id}...'
+Since we can't substitute every stop on every route into an api call at once,
+segment the number of strings created by num_stops_per_route
+'''
+
+
+def get_routes_to_stops_str(agency_id: str, num_stops_per_str: int) -> str:
+    route_ids = [route.id for route in nextbus.get_route_list(agency_id)]
```
youngj: Could make route_ids a parameter to get_prediction_data so the caller could request a subset of routes instead of all routes (like trynapi.get_state)

models/nextbus_predictions.py
```diff
@@ -0,0 +1,130 @@
...
+    for route_id in route_ids:
+        ret = gen_stop_str_for_route(
+            route_id, route_id_to_stop_ids[route_id], num_stops_so_far)
+        route_str += ret['stop_str']
+        num_stops_so_far += ret['num_stops_so_far']
+        if num_stops_so_far >= num_stops_per_str:
+            route_strs.append(route_str)
+            route_str = ''
+            num_stops_so_far = 0
+    return route_strs
```
youngj: is the last route_str not added to route_strs?

models/nextbus_predictions.py
```diff
@@ -0,0 +1,130 @@
...
+
+    route_strs = []
+    num_stops_so_far = 0
+    route_str = ''
+    for route_id in route_ids:
+        ret = gen_stop_str_for_route(
+            route_id, route_id_to_stop_ids[route_id], num_stops_so_far)
+        route_str += ret['stop_str']
+        num_stops_so_far += ret['num_stops_so_far']
+        if num_stops_so_far >= num_stops_per_str:
```
youngj: since it appends the stops to the string before checking how many there are, there could be more than 400. could this cause Nextbus API to return an error?

models/nextbus_predictions.py
```diff
@@ -0,0 +1,130 @@
...
+    route_id_to_stop_ids = {}
+    for route_id in route_ids:
+        route_config = nextbus.get_route_config(agency_id, route_id)
+        route_id_to_stop_ids[route_id] = route_config.get_stop_ids()
+
+    route_strs = []
+    num_stops_so_far = 0
+    route_str = ''
+    for route_id in route_ids:
+        ret = gen_stop_str_for_route(
```
youngj: may be simpler if code in gen_stop_str_for_route is inlined here instead of having a separate function

models/nextbus_predictions.py
```diff
@@ -0,0 +1,130 @@
...
+                route_id, stop_id, queried_time, predictions, route_id_to_predictions)
+            route_id_to_predictions.update(curr_map)
+
+    return route_id_to_predictions
+
+
+def gen_route_id_to_predictions(route_id, stop_id: str, queried_time, predictions: list, route_id_to_predictions: dict):
+    if isinstance(predictions, list):
+        for prediction in predictions:
+            new_prediction = gen_prediction(
```
youngj: the content of this for loop could be defined as an inner function to reuse code for else condition

models/nextbus_predictions.py
```diff
@@ -0,0 +1,130 @@
...
+        # print(prediction_info)
+        route_id = prediction_info['routeTag']
+        stop_id = prediction_info['stopTag']
+        if 'direction' not in prediction_info:
+            continue
+        directions = prediction_info['direction']
+        if isinstance(directions, list):
+            # meaning there are multiple directions at this stop
+            for direction in directions:
+                predictions = direction['prediction']
```
youngj: the content of this for loop could be defined as an inner function to reuse code for else condition

models/predictions.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class Prediction:
+    def __init__(self, route_id: str, stop_id: str, vehicle: str, minutes: str, epoch_time: str, queried_time: datetime, direction_tag: str):
+        self.route_id = route_id
+        self.stop_id = stop_id
+        self.vehicle = vehicle
+        self.predicted_minutes = minutes
+        self.epoch_time = epoch_time
+        self.queried_time = queried_time
```
youngj: Is queried_time necessary since the API returns epoch_time? 

It's a bit inconsistent to have two variables ending in `_time` that use different formats to represent time (datetime, and string containing Unix timestamp in milliseconds) which are both different from the normal way metrics-mvp represents time (integer containing Unix timestamp in seconds). Probably would be simpler for predictions to represent time as a Unix timestamp in seconds.

models/nextbus_predictions.py
```diff
@@ -0,0 +1,130 @@
...
+                curr_map = gen_route_id_to_predictions(
+                    route_id, stop_id, queried_time, predictions, route_id_to_predictions)
+                route_id_to_predictions.update(curr_map)
```
youngj: is this line necessary? It looks like gen_route_id_to_predictions already updates route_id_to_predictions by reference.

models/nextbus_predictions.py
```diff
@@ -0,0 +1,130 @@
...
+
+    # TODO: cache responses
+    route_id_to_predictions = {}
+    for request_url in create_predictions_requests(agency_id):
+        response = requests.get(request_url)
+        if response.status_code != 200:
+            response.raise_for_status()
+
+        route_id_to_predictions.update(
+            parse_prediction_response(queried_time, response))
```
youngj: if get_routes_to_stops_str could need to split a route across multiple requests, route_id_to_predictions.update would overwrite the existing predictions for a route ID. to allow that, could pass route_id_to_predictions as a parameter to parse_prediction_response (instead of calling route_id_to_predictions.update).

.gitignore
```diff
@@ -464,4 +464,7 @@ typings/
 
 #adding data folder
 data/*.json
-data/*.csv
\ No newline at end of file
+data/*.csv
+
+# docker compose config
+docker-compose.yml
```
youngj: docker-compose.yml shouldn't be added to .gitignore

models/nextbus_predictions.py
```diff
@@ -0,0 +1,130 @@
...
+    return route_id_to_predictions
+
+
+def parse_prediction_response(queried_time, resp):
+    route_id_to_predictions = {}
+    resp_json = resp.json()
+    predictions_by_route = resp_json['predictions']
```
youngj: I get an `KeyError: 'predictions'` at this line. The API response was:

`{'copyright': 'All data copyright San Francisco Muni 2019.', 'Error': {'content': 'For agency=sf-muni stop s=34341 is on none of the directions for r=28 so cannot determine which stop to provide data for.', 'shouldRetry': 'false'}}`

models/predictions.py
vicky11z: Got rid of epoch_time, kept queried_time because I figured that would be more useful for caching. Converted epoch_time to Unix timestamp in seconds.

vicky11z:
Reopening this, at least for testing purposes.

hathix:
Messy stuff. Thanks for your patience & determination!

models/nextbus_predictions.py
```diff
@@ -0,0 +1,68 @@
...
+    else:
+        new_prediction = gen_prediction(
+            predictions, route_id, stop_id, queried_time)
+        ret_predictions.append(new_prediction)
+    return ret_predictions
+
+
+def gen_prediction(prediction, route_id, stop_id, queried_time):
+    vehicle = prediction['vehicle']
+    minutes = prediction['minutes']
```
youngj: Nextbus API also returns a `seconds` field - may be better to store that instead of minutes since just in case the extra precision is useful later on.

Could also store affectedByLayover and block in case they are useful later on.

models/nextbus_predictions.py
```diff
@@ -0,0 +1,68 @@
...
+import os
+from datetime import datetime
+
+from models import predictions as p, nextbus
+
+STOPS_STR = '&stops='
+
+def create_prediction_request_for_route(agency_id: str, route_id: str, stops: list) -> str:
+    stops = [f"{route_id}|{stop_id}" for stop_id in stops]
+    stop_str = STOPS_STR + STOPS_STR.join(stops)
```
youngj: Perhaps would be easier to understand as:
```
stop_str = "&".join([f"stops={route_id}|{stop_id}" for stop_id in stops])
```
and then `a={agency_id}&{stop_str}` on the next line

models/nextbus_predictions.py
```diff
@@ -0,0 +1,68 @@
...
+        response.raise_for_status()
+    resp_json = response.json()
+    if 'predictions' not in resp_json:
+        # request errored out, will raise an error here later
+        return
+    return parse_prediction_response(queried_time, resp_json)
+
+def parse_prediction_response(queried_time, resp_json: dict) -> list:
+    predictions_for_route = []
+    predictions_by_route = resp_json['predictions']
```
youngj: Having variables named `predictions_for_route` and `predictions_by_route` in the same function could be confusing, probably easiest to remove predictions_by_route by replacing it with resp_json['predictions'] on the next line.

models/predictions.py
```diff
@@ -0,0 +1,97 @@
...
+    if re.match(r'^[\w\-]+$', agency) is None:
+        raise Exception(f"Invalid agency: {agency}")
+
+    if re.match(r'^[\w\-]+$', datetime_str) is None:
+        raise Exception(f"Invalid date: {datetime_str}")
+
+    return os.path.join(util.get_data_dir(), f"predictions_{agency}_{datetime_str}.json")
+
+
+def get_predictions_for_route(agency: str, route_id: str) -> list:
```
youngj: Recommend moving this function to nextbus.py, to indicate that this function calls Nextbus API. Also this would allow predictions.py not to have any Nextbus specific code, making it easier to reuse it in the future for other transit data providers besides Nextbus.

models/nextbus_predictions.py
```diff
@@ -0,0 +1,68 @@
+import re
```
youngj: I'd recommend moving the content of nextbus_predictions.py into nextbus.py. It doesn't seem necessary to have a separate file for this, and using nextbus.py would make it more natural to use, e.g.

```
import nextbus
predictions = nextbus.get_predictions_for_route('sf-muni','27')
```

hathix:
We should merge this soon. Anyone want to address Jesse's last few comments and merge this in?

akgupta89:
This is currently abandoned. @youngj or @jtanquil do you want to finish this off?

PR #120, opened 2019-05-29
Optimize arrival time calculations

youngj:
This PR implements several changes to the algorithm for computing arrival times from GPS observations to improve correctness and performance.

**Variable eclipse distance**

Instead of using a fixed 200m radius to determine when a bus reaches a stop, it now uses the smaller of 200m or the distance to the nearest adjacent stop along the route. It also requires the bus to be closer to that stop than the previous or next stop. This fixes most weird issues near the end of certain routes where the bus never leaves a 200m radius of the second-to-last (or third-to-last) stop.

There are still some unfixed edge cases like the 30 to Jefferson Loop which reaches the corner of Chestnut and Divisadero twice while going in the same "direction".

**Second pass to add missing stops**

Sometimes the first pass of the algorithm will find an arrival at stop `i` and stop `i+2` but not stop `i+1` for some reason. If this occurs there is now a second pass that fills in small gaps of 1 or 2 stops by checking for arrivals again with a 300m radius in the relevant time period and not requiring the vehicle to be closer to that stop than the previous or next stop.

**One-way only routes**

There are some routes that only run inbound in the morning and outbound in the afternoon, but the algorithm was computing arrivals when the bus was returning along the route in the opposite direction but presumably not picking up passengers. For now these routes are hardcoded to ignore certain directions based on the time of day. Perhaps that could be determined dynamically once we have schedule data.

**Performance improvements**

This code computes arrival times about 7 times faster than before. Assuming the state from tryn-api is already cached locally, it takes about 2.5 minutes (on my dev machine) to compute 1 day of arrival times for all SF Muni routes. 

The improved performance will make it more feasible to compute historical arrival times from the route state that is already stored, and eventually to provide near-real-time arrival stats for the current day.

Computing arrivals for one route typically takes a couple of seconds, e.g.:

```
# time python compute_arrivals.py --date=2019-05-24 --route=KT
time = [2019-05-24 03:00:00-07:00, 2019-05-25 03:00:00-07:00)
state already cached
retrieved state in 0.0 sec
loading state for route KT from cache: /app/data/state_sf-muni/2019-05-24/state_sf-muni_KT_1558692000_1558778400.json
KT: 0.0 generating data frame of GPS observations
KT: 0.2 resampling 128906 GPS observations
KT: 0.9 computing distances from 280373 resampled GPS observations to stops
KT: 1.8 computing possible arrivals
KT: 2.3 cleaning arrivals
KT: 2.9 found 6672 arrivals
KT: 3.0 saving arrival history
KT: 3.02 done

real    0m3.767s
user    0m3.430s
sys     0m0.350s
```

Much of this performance increase was achieved by avoiding expensive operations on pandas series/dataframes in inner loops and using numpy operations instead. I used cProfile a lot to figure out where the code was being slow, e.g.:

```
# python -m cProfile -s cumtime compute_arrivals.py --date=2019-05-24 --route=KT
```

For example, apparently simple Pandas operations like `bus['DT'] = bus.TIME - bus.PREV_TIME` turn out to involve a lot of overhead in creating pandas series, doing operations on series, and adding new series to a data frame. It is much faster to extract the numpy arrays from the series using `.values`, and then do operations directly on numpy arrays like this:

```
time_values = bus['TIME'].values
prev_time_values = np.r_[np.nan, time_values[:-1]]
dt_values = time_values - prev_time_values
```

Probably the biggest increase in performance was from optimizing `get_possible_arrivals_for_stop` to avoid calling `.groupby` by using the fact that each "eclipse" was just a contiguous range within the eclipses data frame, so the relevant data for each eclipse could be retrieved from a slice of the numpy arrays in the eclipses data frame.

**Route state caching**

compute_arrivals.py now always stores downloaded route state locally, with a separate file per route per day. When compute_arrivals.py is called without a `--route` argument, it will first download the entire day's worth of state for all routes from tryn-api, then save the state for each route in a separate cache file. Then, it will load only that route's cache file from disk when computing arrivals for the route. 

This ends up making the arrival computations for all routes faster because there is more free memory and the Python garbage collector doesn't need to work as hard. 

Storing each route's state in a separate cache file also allows rerunning compute_arrivals.py with a different `--route` argument without needing to re-download the state from tryn-api. 

**Cache path subdirectories**

Arrival and route state cache data files are now stored in date-specific subdirectories of data/ to make it easier to navigate and delete stale data.

### DIFF
--- a/models/eclipses.py
+++ b/models/eclipses.py
@@ -1,8 +1,7 @@
 import json
 import time
 import requests
-
-from datetime import datetime, timedelta, timezone
+from datetime import datetime, date, timedelta, timezone
 from geopy.distance import distance
 import pytz
 import pandas as pd
@@ -11,7 +10,7 @@ import math
 
 from typing import List, Union
 
-from . import nextbus
+from . import nextbus, util
 
 def produce_buses(route_state: dict) -> pd.DataFrame:
     buses = pd.io.json.json_normalize(route_state,
@@ -51,9 +50,13 @@ def resample_bus(bus: pd.DataFrame) -> pd.DataFrame:
     # is reasonably small (allowing a smaller radius around stop and more precise arrival times),
     # but not too small to create an unnecessarily large number of rows (slower to calculate)
     new_rows = []
+    target_dist = 25
+
+    bus['NUM_SAMPLES'] = np.floor(bus['MOVED_DIST'] / target_dist) # may be 0
+
     for row in bus.itertuples():
-        if row.DT < 180 and row.MOVED_DIST >= 50 and row.MOVED_DIST < 2500:
-            num_samples = math.floor(row.MOVED_DIST / 25)
+        if row.NUM_SAMPLES > 1 and row.NUM_SAMPLES < 100 and row.DT < 180:
+            num_samples = int(row.NUM_SAMPLES)
             for i in range(1, num_samples):
                 frac = i / num_samples
                 new_rows.append((
@@ -65,15 +68,34 @@ def resample_bus(bus: pd.DataFrame) -> pd.DataFrame:
                     # 1 # uncomment for debugging
                 ))
         new_rows.append((
-            row.VID, row.DID, row.LAT, row.LON, row.TIME,
+            row.VID,
+            row.DID,
+            row.LAT,
+            row.LON,
+            row.TIME
             # 0 # uncomment for debugging
         ))
+
+    # adding an invalid row at the end of each vehicle's observations allows simplifying
+    # get_possible_arrivals_for_stop() (and making it slightly faster).
+    # when it filters rows near a stop, this row will always be filtered out,
+    # so the row index will always have a gap in it even if two vehicles adjacent in the buses
+    # both happen to be near the same stop
+    new_rows.append((
+        row.VID,
+        '',
+        0,
+        0, # 0,0 is in the middle of the ocean so there are no stops near here
+        0
+        # 0 # uncomment for debugging
+    ))
 
     resampled_bus = pd.DataFrame(new_rows, columns=[
         'VID','DID','LAT','LON','TIME',
         # 'INTERP' # whether a sample was interpolated isn't needed by algorithm, but useful for debugging
     ])
     resampled_bus['TIME'] = resampled_bus['TIME'].astype(np.int64)
+
     return resampled_bus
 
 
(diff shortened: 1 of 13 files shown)


models/eclipses.py
```diff
@@ -93,85 +129,268 @@ def haver_distance(latstop,lonstop,latbus,lonbus):
...
     return distance
 
-def find_arrivals(buses: pd.DataFrame, route_config) -> pd.DataFrame:
+PM = [('12:00', None)]
+AM = [(None, '12:00')]
+
+invalid_direction_times_map = {
+    'sf-muni': {
+        'NX': {
+            'NX___I_F00': PM,
```
EddyIonescu: Just wondering, what are your thoughts about using Nextbus to get the list of stops for a given direction? It can be fetched from the directions field here: http://restbus.info/api/agencies/sf-muni/routes/F
One issue I see is that drivers sometimes forget to switch their sign and so the directionID sent out isn't accurate, so I'm fine with leaving it as is.

youngj: We're already using Nextbus to fetch the list of stops for a given direction. 

This code handles cases when a stop is valid for a given direction, but not at a particular time of day.

models/eclipses.py
```diff
@@ -93,85 +129,268 @@ def haver_distance(latstop,lonstop,latbus,lonbus):
...
-
-    route_id = route_config.id
+    #tz = pytz.timezone('US/Pacific')
+    #buses['DATE_TIME'] = buses.TIME.apply(lambda t: datetime.fromtimestamp(t, tz))
 
     possible_arrivals_arr = []
+
+    # add distance from each observation to each valid stop along this route
+    # in a column named DIST_{stop_id} ... assumes there are not too many stops
+    # that we run out of memory
```
EddyIonescu: How many observations would be expected here? (4 vehicles / minute, for an hour?). If up to 100 sops works then it's fine, just wondering if it should be mentioned.

models/eclipses.py
youngj: This happens sometimes, but the issue occurs even if the operator doesn't select the wrong direction, because this algorithm ignores the direction selected by the operator.

models/eclipses.py
youngj: For example, after interpolation the KT had about 280,000 observations on May 24.

Issue #90, opened 2019-05-08
react-d3-components has issues


exxonvaldez:
Same chart in React-Vis vs react-d3-components, trying to match NYC bus stats as closely as possible just as a test:

![Screen Shot 2019-05-08 at 4 12 15 PM](https://user-images.githubusercontent.com/44861283/57414379-b797c000-71ac-11e9-8b49-b690c74ad4b0.png)
In creating a new "phases of the day" chart, I came across these issues with react-d3-components:

* Tooltips when hovering over charts do not work.
  * Possibly due to conflict with react-bootstrap CSS.
  * Could not find a good fix, seems hard to override third party packages.
* Documentation is not available, but there are good set of examples.
* Customization options are limited.
  * Example: cannot remove axis lines.
  * The original (non-React) D3 API allows clients to remove and modify svg elements it creates, but here we cannot do that.
* No built in support for chart legends.
* Last pull request merged June 2018, project is looking for maintainers.

I looked for articles about popular React charting packages:

* https://medium.com/dailyjs/data-visualization-libraries-for-react-developers-in-2019-a2b9c01262f8
  * (Victory, React-Vis by Uber)
* http://voidcanvas.com/top-10-react-graph-chart-libraries-with-demo/
  * (ReCharts, Victory, VX, React-Vis, Nivo, ChartJS, Rumble, highCharts, Easy Chart, fusionCharts)
* https://www.overloop.io/blog/top-5-react-chart-libraries
  * (Nivo, React-vis, Victory, ReCharts, ChartJS)

Of these, React-Vis looked good visually. It offered more customization options (and correspondingly more complex component hierarchy). It uses D3-inspired scales.

I tried reimplementing a chart in React-Vis.  Worked pretty well.  Issues:

* Documentation is sometimes out of date.
* Link to view source of examples is sometimes missing.
* ~~Some bugs.  For example, height of legend lines is not actually settable due to some hardcoded css.~~ My mistake, included wrong version of react-vis css file.

exxonvaldez:
![Screen Shot 2019-05-08 at 4 12 15 PM](https://user-images.githubusercontent.com/44861283/57414364-acdd2b00-71ac-11e9-93e1-0afbff763c26.png)

hathix:
NYC BusStats uses Plot.ly; there's [a plugin to make Plot.ly work with React](https://plot.ly/javascript/react/) so we could use that later if we want.

hathix:
[...]

PR #121, opened 2019-05-30
Replacing react-d3-components charts with react-vis

exxonvaldez:
As mentioned in https://github.com/trynmaps/metrics-mvp/issues/90 , we wanted to replace the react-d3-components charts with react-vis charts.  These new charts are more customizable, allow hover tooltips, and for now, have a fixed width so that the page doesn't get super wide for unusual histograms.  If needed, later we can make these flexible width.

Normal histograms:

Screen Shot 2019-05-30 at 3 28 56 AM

Unusual histograms:

Screen Shot 2019-05-30 at 3 29 49 AM

### DIFF
--- a/frontend/src/components/Info.jsx
+++ b/frontend/src/components/Info.jsx
@@ -1,16 +1,21 @@
 import React, { Component } from 'react';
 import { css } from 'emotion';
-import { BarChart } from 'react-d3-components';
 import Card from 'react-bootstrap/Card';
 import InfoIntervalsOfDay from './InfoIntervalsOfDay';
-import { getPercentileValue } from '../helpers/graphData'; 
+import { getPercentileValue, getBinMin, getBinMax } from '../helpers/graphData'; 
 import { PLANNING_PERCENTILE } from '../UIConstants';
 import * as d3 from 'd3';
+import { XYPlot, HorizontalGridLines, XAxis, YAxis, VerticalRectSeries,
+  ChartLabel, Crosshair } from 'react-vis';
+
 
 class Info extends Component {
   constructor(props) {
     super(props);
-    this.state = 0;
+    this.state = {
+        crosshairValues: { }, // tooltip starts out empty
+    };
+
   }
   
   computeGrades(headwayMin, waitTimes, tripTimes, speed) {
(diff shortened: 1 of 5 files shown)


frontend/src/components/Info.jsx
```diff
@@ -412,23 +481,43 @@ minutes, max
...
+                text="minutes"
+                className="alt-x-label"
+                includeMargin={false}
+                xPercent={0.90}
+                yPercent={0.94}
+                />       
+
+                { this.state.crosshairValues.trip && (
+                  <Crosshair values={this.state.crosshairValues.trip}
+                    style={{line:{background: 'none'}}} >
```
EddyIonescu: I noticed we have this style in multiple spots, thoughts on putting it in an scss file and applying it via classname?

exxonvaldez:
> Looks good to me! Just wondering, did you also check out plotly.js (also built on top of d3 - https://plot.ly/javascript/)? Asking since I had a great time using it for another project, although react-vis seems pretty straightforward and maybe more flexible.

Plotly didn’t come up when I did a survey of React plotting libraries (see https://github.com/trynmaps/metrics-mvp/issues/90).  Neel mentioned on that issue that nyc bus stats uses Plotly, and it looks pretty comprehensive to me.  We can definitely look at it the next time we add more charts or when we have a high fidelity UI design.

frontend/src/components/Info.jsx
exxonvaldez: Can do, though the project currently only has css files, no scss.  Is css good enough?  (I haven’t worked with scss much so a bit ignorant here)

exxonvaldez: Oh, actually that style attribute is a javascript object with a tree of style properties.  I can refactor it into UIConstants.js as a javascript constant.

EddyIonescu: Yeah I think it is. It looks like they'd support styling via className rather than having a bunch of JS style constants - https://github.com/uber/react-vis/blob/master/docs/style.md

exxonvaldez: Thanks for the doc pointer, that’s helpful (admittedlty react-vis’ docs are not a strong point).

.rv-crosshair {
  position: absolute;
  font-size: 11px;
  pointer-events: none;
}

.rv-crosshair__line {
  background: #47d3d9;
  width: 1px;
}

I went through their scss source files on github and found the class we want to override.  I think the css className option is only for overriding the component’s top level css (e.g. Crosshair’s .rv-crosshair class), and not the “internal” css classes.

PR #122, opened 2019-06-04
Weird arrivals error

vicky11z:
For some reason, this combo of params:
Screen Shot 2019-06-03 at 11 54 00 PM
errors the `metrics_by_interval` param out as it is the first bus of the day. Put together this hacky solution for now because I don't think I understand what the code is doing here enough, but the endpoint will end up returning `null` if `06:00-07:00` is within a timeframe that is being calculated by the interval for this particular line in this direction. Would like to hear advice/relevant context.

### DIFF
--- a/errors.py
+++ b/errors.py
@@ -1,3 +1,7 @@
 class StopNotOnRouteError(Exception):
    """Raised when stop does not exist on route"""
    pass
+
+class FirstBusOfDayError(Exception):
+   """Raised for errors when calculating timing before the first bus of the day"""
+   pass
--- a/metrics-api.py
+++ b/metrics-api.py
@@ -11,7 +11,7 @@ import math
 from models import metrics, util, arrival_history, wait_times, trip_times, nextbus
 import constants
 import sys
-from errors import StopNotOnRouteError
+from errors import StopNotOnRouteError, FirstBusOfDayError
 
 """
 This is the app's main file!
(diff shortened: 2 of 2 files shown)


jtanquil:
So what's happening here is that in `calc_metrics`, when the condition `headway_min.empty` is checked, even though `headway_min` only contains the single `NaN` entry, the condition `headway_min.empty` is still false. The error is occuring because `metrics.get_headways_stats` passes an empty DataFrame (since it passes the `headway_min` df with all the `NaN` entries removed) which is eventually passed to `numpy.percentile`, which is where the error is thrown.

Changing the condition to something like

`if len(headway_min.dropna()) > 0:`

should work. Also, since it's possible for a query to have no headways but still have wait times or trips, the conditional should be three separate ones. Something like

`    if len(headway_min.dropna()) > 0:
        headway_min_hist = metrics.get_headways_stats(headway_min)
    if len(waits.dropna()) > 0:
        wait_times_hist = metrics.get_wait_times_stats(waits, tz)
    if len(completed_trips.dropna()) > 0:
        trip_times_hist = metrics.get_trip_times_stats(completed_trips, start_stop_id, end_stop_id) if end_stop_id and both_stops_same_dir else None`

returns data that matches what you see from the command line scripts.

vicky11z:
Gotcha, thanks Jesse!

PR #123, opened 2019-06-05
refactor metrics calculations; improve performance of arrival history and wait times

youngj:
This PR initially was about improving performance for computing wait times over all stops and routes (for the isochrone map). In the process of doing that, I ended up refactoring the metrics API calculations to make it easier to maintain and add new metrics without reducing performance. 

**Refactor metrics API calculations**

This refactor allows calculating stats for wait times, trip times, headways, etc., in separate methods, while reusing the arrival history and data frames to avoid reducing performance.

The metrics API creates one RouteMetrics object for a particular route, and calls methods on that object to compute wait times, trip times, and headways for various date/time ranges. These methods separate the logic for wait times, trip times, and headways to make the code easier to maintain. These methods also allow the caller to specify which stats they want (e.g. avg, count, median, percentiles, histogram, etc) so the server doesn't do unnecessary work if the client doesn't need a particular statistic.

The RouteMetrics object caches arrival history and data frame objects so that each metric doesn't need to reload the raw arrival data. Once we add date ranges to the UI, this will become more important because it might be necessary to load >100 days of arrival data for wait times, trip times, headways, etc.

RouteMetrics also will make it straightforward to combine /metrics_by_interval and /metrics into the same API method so that the data doesn't need to be loaded twice (although not implemented in this PR). I think it would make sense to have one metrics method that allows the client to specify what metrics it wants via a GraphQL query.

**Improve arrival history performance by removing datetimes**

Profiling showed that adding datetime information to the arrival history data frame had a major impact on performance (not really noticeable in the metrics API, but noticeable when computing wait times/trip times over all stops). Since the various metrics for wait times, trip times, headways etc. only use timestamps, and datetimes are only necessary in the command line scripts when printing the raw data, I updated the arrival history data frame to work entirely on timestamps.

**Improve wait times performance and calculate exact wait time statistics**

Using numpy instead of pandas to calculate the sampled wait times made a big difference in performance when computing wait times over all stops.

I also implemented exact calculations for average wait times, quantiles and histograms (i.e., calculations that don't rely on sampling the wait times every 60 seconds) as described in the comments for wait_times.py. 

The exact methods are somewhat faster than the optimized sampled methods, although the performance difference is not very significant because both approaches can compute wait times for all stops on all routes for all displayed time intervals one day in about 20 seconds or less. The exact methods return slightly more precise wait time statistics (average wait times calculated with a sampling time of 60 seconds have a mean absolute error of about 8 seconds across all stops).

I'm still working on the script to compute and cache wait times over all stops for different time ranges, so that isn't included in this PR. I also didn't address performance of computing trip times in this PR, but planning to work on that soon.


models/metrics.py
```diff
@@ -1,110 +1,284 @@
...
+import numpy as np
+
+ROUND_DIGITS = 3
+DEFAULT_STAT_KEYS = ['count', 'avg', 'min', 'median', 'max']
+
+# Represents a range of days with a time range within each day.
+# RouteMetrics can calculate various statistics over a range.
+class Range:
+    def __init__(self, dates: list, start_time_str: str, end_time_str: str, tz: pytz.timezone):
+        self.dates = dates                      # list of datetime.date objects
```
jtanquil: Would it be useful to allow Range objects to specify different start/end time ranges for each day? (i.e. 12PM-6PM on one day and 8AM-8PM another day)

metrics-api.py
```diff
@@ -79,47 +77,81 @@ def metrics_page():
...
+
+        data['trip_times'] = route_metrics.get_trip_time_stats(
+            direction_id, start_stop_id, end_stop_id,
+            rng, keys
+        )
+
+        data['headway_min'] = route_metrics.get_headway_min_stats(
+            direction_id, start_stop_id,
+            rng, keys
+        )
```
jtanquil: The ability to pass params for the keys and metrics to compute seems like a good idea for a future feature.

models/metrics.py
youngj: It doesn't seem like the UI is going to need to have different time ranges on different days combined in the same statistics. Anyway this just keeps the same functionality the metrics API currently has. The main purpose of the Range object is to reduce the number of date/time related parameters that need to be passed to the various methods.

models/wait_times.py
```diff
@@ -1,48 +1,326 @@
...
+        # if the logic above is correct,
+        # the first returned point should be (min wait time, 0.0), and
+        # the last returned point should be (max wait time, 1.0)
+        if points[-1][1] != 1 or points[0][1] != 0:
+            # should never get here unless the code is broken
+            print(points)
+            print(interval_headways)
+            print(sorted_wait_time_values)
+            print(end_wait_time)
+            raise AssertionError('Invalid cumulative distribution')
```
jtanquil: There are some parameters that are causing this error to be thrown - for example, `http://localhost:5000/metrics?route_id=38&start_stop_id=4261&end_stop_id=4292`.

youngj: Good catch. This happens when the time_values array is not already sorted, which only happened for routes like 38 that have stops with multiple directions. Fixed by sorting the times in metrics.py. Also switched to using departure times instead of arrival times for wait times and headways stats since departure times are more relevant.

jtanquil:
Nice, the cdf error doesn't appear to be happening anymore. I'm still getting a few errors: `http://localhost:5000/metrics?route_id=S&start_stop_id=5728&end_stop_id=5730` throws a `TypeError`:
image

youngj:
Good catch - timestamp was None instead of np.nan. fixed

models/trip_times.py
```diff
@@ -23,6 +23,6 @@ def find_dest_arrival_time(row):
 
     s1_df['dest_arrival_time'] = s1_df.apply(find_dest_arrival_time, axis=1)
     s1_df['trip_min'] = (s1_df.dest_arrival_time - s1_df.DEPARTURE_TIME)/60
-    s1_df['dest_arrival_time_str'] = s1_df['dest_arrival_time'].apply(lambda timestamp: datetime.fromtimestamp(timestamp, tz).time() if not np.isnan(timestamp) else None)
+    s1_df['dest_arrival_time_str'] = s1_df['dest_arrival_time'].apply(lambda timestamp: datetime.fromtimestamp(timestamp, tz).time() if timestamp is not None else None)
```
jtanquil: This change causes a different error to be thrown: 
image

youngj: Not sure how that happens, but it should be fixed now. I'm planning to work on optimizing trip time computations next, so this code will probably be rewritten soon anyway.

PR #126, opened 2019-06-09
Improve performance of computing trip times

youngj:
This makes it practical to compute trip time statistics over all pairs of stops on all routes for various time ranges. Computing average, median, 10th and 90th percentiles for all pairs of stops in all time ranges in the UI for one day now takes less than 3 minutes on my machine.

To do this, compute_arrivals.py is updated to store a "trip ID" for each arrival in the arrival history. The trip ID is an arbitrary integer that is the same for a group of arrivals for the same vehicle and direction that have an ascending stop index.

Storing the trip ID for each arrival makes it fast to use numpy to find the arrival times at stop s2 corresponding to departure times at stop s1. For each trip ID seen at s1, trip_times.py finds the index where the same trip ID appears at s2, and looks up the corresponding arrival time.

Also fixes a bug where the first GPS location of the day for each bus was always filtered out by resample_bus.


models/trip_times.py
```diff
@@ -1,27 +1,85 @@
...
+    # element at the end (which will be ignored).
+    #
+    # The s1 arrays do not need to be sorted.
+    #
+    # When the caller needs to compute trip times for many pairs of stops at once,
+    # this function enables better performance by allowing the caller to pre-sort and pre-extend
+    # the s2 arrays for each stop.
+
+    # find potential indexes of s1 trip IDs in s2 array (ignoring extra value at the end of s2)
+    possible_match_indexes = np.searchsorted(s2_trip_values_extended[:-1], s1_trip_values, side='left')
```
youngj: Looks like https://docs.scipy.org/doc/numpy/reference/generated/numpy.intersect1d.html with return_indices=True would be more straightforward than np.searchsorted for matching s1 departures to s2 arrivals (assuming it's not slower)

youngj:
np.intersect1d turned out to be quite a bit slower than np.searchsorted, but I found a library sortednp that supports intersecting two sorted arrays and returning the matching indexes (https://gitlab.sauerburger.com/frank/sortednp). snp.intersect is somewhat faster than np.searchsorted and also makes the code simpler. 

While comparing the different algorithms for computing trip times to ensure they returned the same results, I encountered an issue where some vehicles were arriving at the same stop twice in a row. Added a function to filter out duplicate stops with the same STOP_INDEX when computing arrivals.

models/eclipses.py
```diff
@@ -486,26 +461,125 @@ def calc_nadir(eclipse_start_index, eclipse_end_index) -> tuple:
...
+        nonlocal start_trip
+
+        stop_index_values = dir_arrivals['STOP_INDEX'].values
+        if len(stop_index_values) == 0:
+            return dir_arrivals
+
+        stop_index_diff = np.diff(stop_index_values, prepend=-999999)
+
+        new_trip = stop_index_diff <= 0
+        trip_id = np.cumsum(new_trip) + start_trip
```
EddyIonescu: Just to make it easier to read through, could you add in a comment what the trip_id is supposed to be? (ie. what you wrote in the PR comment: `The trip ID is an arbitrary integer that is the same for a group of arrivals for the same vehicle and direction that have an ascending stop index.`).

PR #128, opened 2019-06-12
First pass at refactoring backend

jtanquil:
This PR cleans up `metrics-api.py` by moving the validation logic to `backend_util.py` and the data fetching logic to `backend_fetch.py`. It also combines the `/metrics_by_interval` endpoint into `/metrics` by introducing the `use_intervals` and `interval_length` parameters to `/metrics`. It also adds a `keys` parameter that allows the client to filter the stats (avg, max, min, etc) it needs.


jtanquil:
Thanks for the feedback! Definitely makes sense to wait on a refactor until we set up GraphQL.

jtanquil:
I plan on making a PR implementing graphQL before refactoring `metrics-api.py` so I think this can be closed.

PR #130, opened 2019-06-13
Converted small components from react components to react hooks

akgupta89:
Eventually we will migrate to react hooks / functional components with everything. We are choosing to build out new components with react hooks, so switching the smallest components to react hooks is the first step in the right direction.

### DIFF
--- a/.idea/codeStyles/codeStyleConfig.xml
+++ b/.idea/codeStyles/codeStyleConfig.xml
@@ -1,0 +1,5 @@
+<component name="ProjectCodeStyleConfiguration">
+  <state>
+    <option name="PREFERRED_PROJECT_CODE_STYLE" value="Default" />
+  </state>
+</component>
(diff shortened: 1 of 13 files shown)


akgupta89:
> Looks good to me (linting feedback is on the other review). Technically these aren't using hooks, just the now-preferred functional component style. Maybe the README.md file should be updated with our React decisions going forward to make onboarding easier.

You're right, I need to be more careful with my vernacular. Also agreed with updating the README.md to explain our decisions. I'm not sure if we should make changes to this PR or open something separate to handle this.

exxonvaldez:
> Also agreed with updating the README.md to explain our decisions. I'm not sure if we should make changes to this PR or open something separate to handle this.

Seems fine to use this PR since it's small (once you dedupe the linting) and helps illustrate the new patterns, plus there's Material UI work in progress as of last week.

PR #131, opened 2019-06-13
Auto formatting on pre-commit

akgupta89:
Autofixes **SOME** of the annoying warnings associated with the airbnb linter style format. Some of the files modified are auto fixed with this PR.

As far as using this, everything works as normal (nothing new to learn)

_i am ~~a god among men~~ iron man_


frontend/.eslintrc.json
```diff
@@ -8,6 +9,8 @@
   "rules": {
     "class-methods-use-this": "off",
     "react/jsx-filename-extension": [1, { "extensions": [".js", ".jsx"] }],
-    "jsx-a11y/label-has-for": "off"
+    "jsx-a11y/label-has-for": "off",
+    "prettier/prettier": "error"
+//    "max-len": ["error", 80]
```
exxonvaldez: Can you explain why this is commented out?  Should it be removed or comments added to explain why it is commented out?

frontend/package.json
```diff
@@ -45,12 +40,26 @@
     "test": "react-scripts test",
     "eject": "react-scripts eject",
     "heroku-postbuild": "npm build",
-    "lint": "node ./node_modules/eslint/bin/eslint.js --fix --ext .jsx",
-    "lint:all": "node ./node_modules/eslint/bin/eslint.js --fix --ext .jsx src/**/*.jsx src/*.jsx"
+    "lint": "eslint src --fix --ext jsx",
```
exxonvaldez: Looks like `--ext .jsx` is the convention?  (see https://eslint.org/docs/user-guide/command-line-interface)

exxonvaldez: Also, should we be linting our .js files?

frontend/package-lock.json
```diff
@@ -998,6 +998,15 @@
       "resolved": "https://registry.npmjs.org/@restart/hooks/-/hooks-0.2.13.tgz",
       "integrity": "sha512-riuYub4Xx/Pk+YUPRHZPwJd2+rAJSsRd4Pqkqvhzu+AbkDdNi4y+kYEQicnAofbfTA5apRPm3cK9sWc6Gj5DNg=="
     },
+    "@samverschueren/stream-to-observable": {
```
exxonvaldez: I'm unclear as to whether package-lock.json should be committed.  It seems like it often results in conflicts (which are probably moot since docker-compose build will need to be run).  Seems like not committing creates less work for everybody else?

frontend/.eslintrc.json
akgupta89: I believe the airbnb code style is 80 characters long, so perhaps this should be uncommented out and we should enforce the style. I'm not a fan but we should definitely adhere to a base guideline, I'll make a change and uncomment it.

exxonvaldez: Ok.  This may only apply to string constants, based on a cursory search.

frontend/package.json
akgupta89: Good catch, thanks!

I left it off because I didn't validate what it would be touching, ideally we should lint format js files. I'll turn it on, hopefully it didn't mess anything up.

frontend/package-lock.json
akgupta89: Ideally we should be using package-lock.json and docker-compose should be running `npm ci` instead of `npm install`  to enforce that we all have the same dependencies.

akgupta89:
Made changes based on suggestions

PR #132, opened 2019-06-16
Roadmap UI: Allow the UI to load all route configs up front instead of one at a time on demand.

exxonvaldez:
The UI design in our roadmap works best with full route configs loaded up front.  This is so that the "spider" map can show all routes that stop near a point, and so that the "home" screen can list all routes and directions (directions are not part of the current route list).  The route configs will be used together with the precomputed trip/wait isochrone data to show grades and rankings for all routes.  Instead of a `/routes` and `/routeConfig` endpoints, there is just a single `\routes` endpoint that returns all route configs.

The volume of data fetched from Nextbus is about 780kb, and with added whitespace becomes 900kb sent from the back end to the front end.

This change is meant to be easily reversible in case issues are discovered with it.  There are other ways for the front end to load all the route configs progressively (such as making one `/routeConfig` call per route) but avoiding the complexity of having only some configs loaded for now.  If we decide to keep this change permanently, then there is some cleanup on the UI side to remove fetchRouteConfig and on the back end to refactor common code in get_route_config and get_route_list.

Note that the Nextbus API only allows 100 routes to be fetched this way.  Muni currently has 84.

The backend now caches a single route config file, instead of one file per route, for up to a day.

### DIFF
--- a/metrics-api.py
+++ b/metrics-api.py
@@ -30,7 +30,19 @@ def ping():
 @app.route('/routes', methods=['GET'])
 def routes():
     route_list = nextbus.get_route_list('sf-muni')
-    data = [{'id': route.id, 'title': route.title} for route in route_list]
+    
+    data = [{
+        'id': route.id,
+        'title': route.title,
+        'directions': [{
+            'id': dir.id,
+            'title': dir.title,
+            'name': dir.name,
+            'stops': dir.get_stop_ids()
+        } for dir in route.get_direction_infos()],
+        'stops': {stop.id: {'title': stop.title, 'lat': stop.lat, 'lon': stop.lon} for stop in route.get_stop_infos()}
+    } for route in route_list]
+    
     return Response(json.dumps(data, indent=2), mimetype='application/json')
 
 @app.route('/route', methods=['GET'])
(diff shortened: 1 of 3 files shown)


metrics-api.py
```diff
@@ -30,7 +30,19 @@ def ping():
...
+        'directions': [{
+            'id': dir.id,
+            'title': dir.title,
+            'name': dir.name,
+            'stops': dir.get_stop_ids()
+        } for dir in route.get_direction_infos()],
+        'stops': {stop.id: {'title': stop.title, 'lat': stop.lat, 'lon': stop.lon} for stop in route.get_stop_infos()}
+    } for route in route_list]
+    
     return Response(json.dumps(data, indent=2), mimetype='application/json')
```
youngj: Could remove `indent=2` to reduce response size from 914 kB to 598 kB.

For production, we should probably have nginx in front of flask to gzip and cache responses to reduce bandwidth and reduce load on the flask backend. Don't need to implement that now though.

exxonvaldez: Thanks, good suggestions.  Will remove indent and file github issue for nginx.

vh02: Re: adding nginx to reduce load on backend, we could also consider using [Envoy](https://www.loggly.com/blog/benchmarking-5-popular-load-balancers-nginx-haproxy-envoy-traefik-and-alb/), which was built at Lyft.

exxonvaldez: > Re: adding nginx to reduce load on backend, we could also consider using [Envoy](https://www.loggly.com/blog/benchmarking-5-popular-load-balancers-nginx-haproxy-envoy-traefik-and-alb/), which was built at Lyft.

Thanks, Vickie, added your suggestion to #133.

PR #134, opened 2019-06-17
Paths and stop select

ckrajewski:
adding support for clicking stops
Still need to account for:
  -clicking a little awkward (doesn't always register click area)
  - different directions other than inbound/outbound


exxonvaldez:
Note: I'm merging these changes into PR #140.  With the graphParams values moved into Redux, managing the selected route, direction, and stops gets a lot simpler.

exxonvaldez:
Closing in favor of #140, all changes here are there.

PR #135, opened 2019-06-18
MapSpider and MapShield component introduction

exxonvaldez:
This is the introduction of the `MapSpider` and `MapShield` components, part of the roadmap UI.  The `MapSpider` component takes a full list of route configs and uses it to show routes that run near a point on a map.  The routes also show "shields" (the `MapShield`component -- circled route numbers using svg) at the terminals. The plotted routes can be clicked but currently do not dispatch any events, as we will probably redo where state is stored in the roadmap UI.
Screen Shot 2019-06-18 at 1 03 34 AM
Chris' `MapStops` and `MapSpider` will need to be normalized to look/behave more alike, but I think it is more maintainable to have multiple, separate map components, rather than overloading one map component with unrelated features. Though it's possible to merge them together if done thoughtfully.

I've moved the haversine functions into a new `RouteCalculations` helper file as `MapSpider` also uses them.  This helper file also has utilities for filtering out problematic routes (such as the owls and the S) and spurious directions.  Some of the other methods will be used in future PRs that deal with scoring and grading of entire routes.


frontend/src/helpers/routeCalculations.js
```diff
@@ -0,0 +1,199 @@
...
+    
+    "J": {
+      directionsToIgnore: ["J____I_D10"] // this is to 23rd and 3rd
+    },
+    "L": {
+      directionsToIgnore: ["L____I_U53"]
+    },
+    "M": {
+      "M____O_D00": {
+        ignoreFirstStop: true // Embarcadero & Folsom is not a real stop
```
youngj: could ignoreFirstStop/ignoreLastStop be implemented as stopsToIgnore (a list of stop IDs)? it seems like there might be some cases where we would need to ignore the first N stops or the last N stops.

frontend/src/components/MapShield.jsx
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Renders an SVG "shield" representing a Muni route.
+ */
+export default function MapShield(props) {
+
+  const routeText = props.routeText;
+  const color = props.color;
+  const waitScaled = props.waitScaled; // 0, 1, or 2
+  return (
+      <svg width="40px" height="40px" viewBox="0 0 40 40" role="img">
```
youngj: interesting, didn't realize it was possible to create a SVG in react like this.

it looks odd to me that the route shields are positioned to the bottom right of the terminal. what do you think about centering them on the terminal?

frontend/src/components/MapSpider.jsx
```diff
@@ -0,0 +1,405 @@
...
+   * Take only stops within CLICK_RADIUS_MI miles and sort by distance.
+   */
+  findStops() {
+    const { routes } = this.props;
+    const latLng = this.state.latLng;
+    const latLon = { lat: latLng.lat, lon: latLng.lng };
+    let stopsByRouteAndDir = [];
+    
+    const filteredRoutes = filterRoutes(routes);
+    for (let i = 0; i < filteredRoutes.length; i++) { // optimize this on back end
```
youngj: we could use the same code that the isochrone map uses to find nearby stops once that is merged in

frontend/src/components/MapSpider.jsx
```diff
@@ -0,0 +1,405 @@
...
+
+    stopsByRouteAndDir = stopsByRouteAndDir.filter(stop => stop.miles < CLICK_RADIUS_MI);
+    stopsByRouteAndDir.sort((a, b) => a.miles - b.miles);    
+    
+    return stopsByRouteAndDir;
+  }
+
+  
+  /**
+   * Returns { distance: (miles); stop: stopObject }.
```
youngj: comment differs from actual return value

frontend/src/components/MapSpider.jsx
```diff
@@ -0,0 +1,405 @@
...
+   * 
+   * TODO: make shield clickable            
+   */
+  generateShield = (startMarker, waitScaled) => {
+
+    const lastStop = startMarker.downstreamStops[startMarker.downstreamStops.length - 1];
+    const shieldPosition = [ lastStop.lat, lastStop.lon ];
+    const routeColor = this.routeColor(startMarker.routeIndex % 10); 
+
+    const icon = L.divIcon({
```
youngj: I'm not familiar with the way things are normally done with react-leaflet but it seems unusual to create real DOM objects when the component is rendered, instead of virtual DOM objects. could this cause performance issues?

frontend/src/components/MapShield.jsx
youngj: it seems like circles/rounded rectangles with text inside could be implemented with normal divs using CSS border-radius -- maybe simpler than SVG?

exxonvaldez:
> When running this I got the error `Module not found: Can't resolve 'react-leaflet-control' in '/app/frontend/src/components'` . Fixed by adding `"react-leaflet-control": "^2.1.1",` to package.json and running docker-compose build.

Thanks for catching that, pushed a fix.  I had created a new branch out of some files from my "experimental" branch but wasn't running docker-compose build when switching branches.

frontend/src/helpers/routeCalculations.js
exxonvaldez: We could, I was making this logic up as I found data issues.  I did implement an option where you could put in one stop id, and it would treat that as the first stop on the route (for `ignoreFirstStop`) or as the last stop (for `ignoreLastStop`).  This will be used for determining end to end trip time, but doesn't currently affect computation of route length, which is based on best guess as to the right GTFS shape.  And the shape is not linked to the stops, it's just coordinates and distance along route, so more logic would be needed to "trim" the shape if stops are ignored.  I'll make a note of a that in my experimental branch.

frontend/src/components/MapShield.jsx
exxonvaldez: Found the svg technique here:  https://medium.com/@nikjohn/creating-a-dynamic-jsx-marker-with-react-leaflet-f75fff2ddb9 .  I think in react-leaflet, the real Leaflet library is rendering the DOM behind React's back:  https://react-leaflet.js.org/docs/en/intro.html .  Leaflet itself uses a lot of svg for circle markers and polylines, seems to perform well.  I can try divs.  I'm also interested in using a dynamic partial circle around text to represent things like scores out of 100.

The design language for the map is borrowed from the official Muni map ( https://www.sfmta.com/sites/default/files/pdf_map/2019/06/sfmta-metro-06.12.2019-web.pdf ).  So that's where the three thickness of lines and three sizes of shields come from.  And there, the terminals are shown usually next to the last stop.  Maybe there should be a small solid circle at the terminal, just like on the Muni map.

frontend/src/components/MapSpider.jsx
exxonvaldez: Thanks, pushed a better comment.

frontend/src/components/MapSpider.jsx
exxonvaldez: Good question.  Based on https://react-leaflet.js.org/docs/en/intro.html and what I've seen while coding, there's not much of a virtual DOM for Leaflet maps.  I think it means state changes can trigger a full rerender of the map, so performance is bad if hovering over Leaflet objects changes the map's state.  Other than that, the real-only DOM seems to perform pretty well.

frontend/src/components/MapShield.jsx
exxonvaldez: Added a circle at the terminal.

PR #140, opened 2019-06-26
New UI screens "Dashboard" and "Route"

exxonvaldez:
I decided to extend what Chintan had started.  He created a new Material UI `Dashboard` screen and added it to our navigation.  I put the `MapSpider` component on the left and created a new Material UI `RouteTable` on the right, which lists either all the routes or just the ones being mapped.  Not to be confused with React routing.

I also created a `Route` screen which for now is essentially the previous UI (`ControlPanel` on the left, and `Info` on the right).  The `Route` screen will eventually have a new `InfoSummary` component to provide route-level stats, as shown in the roadmap.  So you'll see either `InfoSummary` or `Info` on the right, depending on whether or not you've selected stops.

This PR isn't quite ready for review yet, but I wanted to give people an idea of where I was in case there's duplication of effort.
-  I'm moving the `ControlPanel`'s graphParams to redux as one big quick and dirty params object for now (Jim might have a better implementation going).  This is so that from the Dashboard screen we can go directly to the "trip view" when the user clicks a stop on the map. 
- I also need to improve `MapSpider` and `RouteTable` state handling. `MapSpider` needs to access Redux state to know where the last click was, and clicking a route in the RouteTable needs to set the selected route in graphParams and go to the Route View.
- I'll need to merge in Chris' PR to get map-based stop selection working, but also rework it to take  state info about selected stops from Redux as we phase out parts of Control Panel.

Screen Shot 2019-06-25 at 4 48 13 PM
Screen Shot 2019-06-25 at 4 45 50 PM

### DIFF
--- a/frontend/public/index.html
+++ b/frontend/public/index.html
@@ -7,9 +7,9 @@
   <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.2.1/css/bootstrap.min.css" integrity="sha384-GJzZqFGwb1QTTN6wy59ffF1BuGJpLSa9DkKMp0DgiMDm4iYMj70gZWKYbI706tWS"
     crossorigin="anonymous" />
 
-  <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/leaflet/1.4.0/leaflet.css" />
-  <!-- Font for Material UI-->
-  <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:300,400,500,700&display=swap" />
+  <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/leaflet/1.5.1/leaflet.css" />
+  <!-- Font for Material UI (Oswald for react-bootstrap) -->
+  <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:300,400,500,700|Oswald:300,400&display=swap" />
   <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons" />
   <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no" />
   <meta name="theme-color" content="#000000" />
@@ -28,9 +28,6 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
   <title>OpenTransit</title>
-  <style>
-    @import url('https://fonts.googleapis.com/css?family=Oswald:300,400|Roboto');
-  </style>
 </head>
 
 <body>
--- a/frontend/public/index.html
+++ b/frontend/public/index.html
@@ -1,30 +1,24 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
-    <meta charset="utf-8" />
-    <link rel="shortcut icon" href="%PUBLIC_URL%/favicon.ico" />
-    <link
-      rel="stylesheet"
-      href="https://maxcdn.bootstrapcdn.com/bootstrap/4.2.1/css/bootstrap.min.css"
-      integrity="sha384-GJzZqFGwb1QTTN6wy59ffF1BuGJpLSa9DkKMp0DgiMDm4iYMj70gZWKYbI706tWS"
-      crossorigin="anonymous"
-    />
 
-     <link
-      rel="stylesheet"
-      href="//cdnjs.cloudflare.com/ajax/libs/leaflet/1.4.0/leaflet.css"
-    />
-    <meta
-      name="viewport"
-      content="width=device-width, initial-scale=1, shrink-to-fit=no"
-    />
-    <meta name="theme-color" content="#000000" />
-    <!--
+<head>
+  <meta charset="utf-8" />
+  <link rel="shortcut icon" href="%PUBLIC_URL%/favicon.ico" />
+  <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.2.1/css/bootstrap.min.css" integrity="sha384-GJzZqFGwb1QTTN6wy59ffF1BuGJpLSa9DkKMp0DgiMDm4iYMj70gZWKYbI706tWS"
+    crossorigin="anonymous" />
+
+  <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/leaflet/1.4.0/leaflet.css" />
+  <!-- Font for Material UI-->
+  <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:300,400,500,700&display=swap" />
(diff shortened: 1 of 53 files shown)


hathix:
Looks awesome!

exxonvaldez:
Okay, I've addressed the previously mentioned "to do" items.  Moving `ControlPanel`'s state to Redux has simplified all the code.  You can try it out at https://ot-terence.herokuapp.com/ .  I think this is ready to be reviewed and merged.  Things to be worked on next:

- New date/time picker. (Arun)
- Transitions from Dashboard (System) screen to Route screen and back need work. (No owner)
   * Should the Route View screen be floating "on top" of the Dashboard and dismissed via an "x" button?  Or do we slide/pan from one to the other?
- Stop selection via the map is working, but needs refinement (especially for lines with more than 2 directions).  (Chris?)
   * Perhaps only show the current direction after selecting the first stop or if a direction was chosen.  
   * Could allow clicking on segments between stops to select the stop at the end of the segment, as in MapSpider.
   * Should the map start out automatically showing the full extent of the selected route?  Or just the same bounds as MapSpider?
- With Jesse's isochrone code merged in, can add more features. (Terence)
   - `MapStops` to show speed along the route.
   - Route summary to be shown when two stops not yet selected.
   - Route table to show computed waits, speeds, scores for each route.
   - Perhaps add a test page to help look for data glitches (e.g., plot trip times across a route for all 84 routes).

frontend/src/components/MapSpider.jsx
```diff
@@ -342,7 +376,7 @@ class MapSpider extends Component {
   render() {
     const { position, zoom } = this.props;
 
-    const mapClass = { width: '500px', height: '500px' };
+    const mapClass = { width: '100%', height: this.state.height };
     
     return (
       <Map center={position || SF_COORDINATES} zoom={zoom || ZOOM} style={mapClass}
```
youngj: I'm not able to see the attribution on the map anymore, perhaps blocked by the horizontal scrollbar?

Also should use attribution for Stamen tiles:
```
Map tiles by <a href="http://stamen.com">Stamen Design</a>, under <a href="http://creativecommons.org/licenses/by/3.0">CC BY 3.0</a>. Data by <a href="http://openstreetmap.org">OpenStreetMap</a>, under <a href="http://www.openstreetmap.org/copyright">ODbL</a>.
```

frontend/src/components/MapSpider.jsx
```diff
@@ -322,16 +346,26 @@ class MapSpider extends Component {
...
+         * this.setDirectionId(startMarker.direction.id);
+         */
             
-            /* If this code was integrated with ControlPanel, we would do:
-             *
-             * this.setRouteId(startMarker.routeID);
-             * this.setDirectionId(startMarker.direction.id);
-             */
+        const {onGraphParams} = this.props;
+        onGraphParams({
```
youngj: Since clicking one of the lines on MapSpider shows stats about trips to the stop near that click (from the stop nearest the initial click), I'd expect that clicking the MapShield at the end of the route shows stats about trips to the end of the route (from the stop nearest the initial click). The current behavior ignores the initial click and requires the user needs to select two stops again, which seems inconsistent.

frontend/src/components/MapStops.jsx
```diff
@@ -1,75 +1,149 @@
...
+    let selectedRoute = null;
+    let routeStops = null;
+    let populatedRoutes = null;
+
+    if (routes && graphParams) {
+      selectedRoute = routes.find(route => route.id === graphParams.route_id);
+
+      if (selectedRoute) {
+        routeStops = {};
+        for (let direction of selectedRoute.directions) {
```
youngj: Circles for inbound and outbound stops are drawn over each other which makes it hard to select the stops for one of the two directions. If a direction is selected in the control panel, it seems like MapStops should only draw circles for that direction (as noted in your earlier comment).

frontend/public/index.html
```diff
@@ -33,15 +27,16 @@
...
-@import url('https://fonts.googleapis.com/css?family=Oswald:300,400|Roboto');
-</style>
-  </head>
-  <body>
-    <noscript>You need to enable JavaScript to run this app.</noscript>
-    <div id="root"></div>
-    <!--
+  <title>OpenTransit</title>
+  <style>
+    @import url('https://fonts.googleapis.com/css?family=Oswald:300,400|Roboto');
```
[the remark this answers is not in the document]
exxonvaldez: Yeah, I was able to combine them, will push in an update.

frontend/src/components/RouteTable.jsx
```diff
@@ -0,0 +1,351 @@
...
+                      hover
+                      onClick={event => handleClick(event, row)}
+                      role="checkbox"
+                      aria-checked={isItemSelected}
+                      tabIndex={-1}
+                      key={row.id}
+                      selected={isItemSelected}
+                    >
+                      <TableCell component="th" id={labelId} scope="row" padding="none">
+                        <a href="#">{row.title}</a> {/* TODO: use a real /route/... link here or button, React complains about this dummy href */}
```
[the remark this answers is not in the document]
exxonvaldez: There's a lot of ways to do the interaction.  Currently MapStops is like Routesy in that first you select a route, and then choose a stop from all possible directions.  The Routesy map defaults to a close zoom so you often see just one inbound stop and one outbound stop next to each other.  It's optimized for getting predictions near your current location, so it's a little different from the "route shopping" use case.  Routesy also has a full list of routes and you tap one to "expand" it and see all directions underneath it.

Earlier, I started building route + direction selection into the table where each row was a dropdown menu of directions for a route, but then realized MapStops didn't respond to the direction being selected.  It always shows all stops in all directions. I think changing it to displaying stops in only the selected direction would make it easier to use.  We can still support a mode where no direction is selected.  Sometimes the user may not know which direction they want, or may not care about direction if they want to see overall route stats.

We can also do route + direction as distinct rows (with or without a header row for just the route).  The route + direction rows could always be visible or behind a toggle.  Lots of choices, something we can experiment with after this is merged in.

frontend/src/components/MapSpider.jsx
exxonvaldez: Good catches.  There was a layout issue where the map extended below the bottom of the screen, and I forgot to update the attribution. I have a fix to push.

frontend/src/components/MapSpider.jsx
exxonvaldez: Have a fix to push.

frontend/src/components/MapStops.jsx
exxonvaldez: I've put this in.  One downside:  there is no easy way to unselect the direction once chosen, if you accidentally clicked the wrong direction.  You can go back to the Dashboard and select the route again.  Or you can change the direction using the ControlPanel's dropdown.

PR #142, opened 2019-06-27
Use 511 api for predictions

vicky11z:
The 511 API is way easier to use!
Get all prediction data from 511 api (aka siri api) and save it in the Prediction or PredictionsByStop classes. Want to develop this iteratively so it's not fully set up but from just printing stuff out it looks pretty good.

Named the api file `siri_api.py` because imports don't play nice when the files start with numbers...? Open to suggestions. The api docs call it the "siri api".


vicky11z:
If you want to play around with the output, you could run this python script:
```
from models import siri_api 
pbs = siri_api.get_predictions_by_stop_id(
    '2b81cac7-b863-4f47-bc87-aab157fc052c', 'SF', '13009')
print(pbs)
print(pbs.stop_id)
print(pbs.prediction_data)
for p in pbs.prediction_data:
    print(p.vehicle_id)
    print(p.get_predicted_minutes_to_arrival())

```

models/predictions.py
```diff
@@ -0,0 +1,101 @@
...
+
+    if re.match(r'^[\w\-]+$', datetime_str) is None:
+        raise Exception(f"Invalid date: {datetime_str}")
+
+    return os.path.join(util.get_data_dir(), f"predictions_{agency}/{datetime_str}/predictions_{agency}_{datetime_str}_{stop_id}.json")
+
+
+def get_predictions_by_stop(agency: str, stop_id: str, datetime: datetime) -> PredictionsByStop:
+    # TODO: round to the nearest 5 minutes
+    cache_path = get_cache_path(agency, stop_id, datetime)
```
exxonvaldez: Is the plan to collect predictions about every 5 minutes and snap the timestamp to an even multiple of 5?

models/siri_api.py
```diff
@@ -0,0 +1,58 @@
...
+        new_prediction = p.Prediction.from_data(prediction_data)
+        stop_id = new_prediction.stop_id
+        if stop_id not in stops_to_predictions:
+            stops_to_predictions[stop_id] = p.PredictionsByStop(
+                stop_id, queried_time, [], agency)
+        stops_to_predictions[stop_id].add_prediction(new_prediction)
+    return stops_to_predictions
+
+
+def get_predictions_by_stop_id(api_key: str, agency: str, stop_id: str) -> p.PredictionsByStop:
```
exxonvaldez: Not sure if we plan to use the "by stop" api only for for small scale testing?  Maybe add comments if that is the case.

models/predictions.py
vicky11z: Yup

models/predictions.py
```diff
@@ -0,0 +1,101 @@
...
+            queried_time=data['queried_time'],
+            prediction_data=predictions,
+            agency=data['agency'],
+        )
+
+    def get_data(self):
+        return {
+            'queried_time': self.queried_time,
+            'stop_id': self.stop_id,
+            'prediction_data': self.prediction_data,
```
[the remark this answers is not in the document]
vicky11z: Sorry not sure what you mean here?

models/siri_api.py
vicky11z: I wanted to do it this way because I figured that it would be most useful to grab predictions by each stop for passing to the frontend, but I can add other methods if you think they would be useful.

vicky11z: However considering what Jesse said below of it not being worthwhile caching all of these, going to get rid of it and we can just rely on `siri_api.get_predictions_by_stop_id`

models/predictions.py
```diff
@@ -0,0 +1,101 @@
...
+            'agency': self.agency,
+        }
+
+
+def get_cache_path(agency: str, stop_id: str, datetime: datetime) -> str:
```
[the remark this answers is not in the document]
vicky11z: Sounds good

models/predictions.py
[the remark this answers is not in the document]
youngj: It seems like get_data should return data that can be serialized to JSON, but self.prediction_data is actually Prediction objects - if you call json.dumps on this I think it would raise `TypeError: Object of type Prediction is not JSON serializable`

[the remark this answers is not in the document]
vicky11z: Oh gotcha, that makes sense

vicky11z:
Closing because there isn't a reliable way to associate 511 api returns to nextbus - working off old PR instead https://github.com/trynmaps/metrics-mvp/pull/119

PR #143, opened 2019-06-27
Compute and cache trip time and wait time statistics for all routes

youngj:
To make it possible for the web app to visualize route performance for multiple routes at once, it is necessary to pre-compute the statistics for trip times and wait times that the frontend would need.

Wait time statistics are stored in files containing a JSON object, with an object mapping Nextbus stop IDs for a particular direction to statistic values, e.g.:
```
{"version":"v1","start_time":1559840400,"end_time":1559862000,"stat":"median","routes":{"E":{"E____O_F00":{"5184":13.6,"3092":11.8,"3095":11.9,"4502":12.0,"4529":12.2,"4516":12.3,"4518":12.2,"4504":12.2,"4534":12.1,"7283":12.1,"7794":12.3,"4506":12.3,"5234":12.5,"5239":12.3,"5241":17.0,"median":12.2},"E____I_F00":{"5240":13.2,"5237":13.1,"7145":14.2,"7795":13.1,"4513":13.8,"4532":13.8,"4503":13.9,"4517":13.9,"4515":13.8,"7281":14.0,"4501":14.2,"4530":14.0,"5174":13.8,"5175":13.7,"35184":13.8,"median":13.8}},"F":{"F____O_F00":{"5184":5.3...
```

Trip time statistics are also stored in files containing a JSON object, with an object mapping "departure" Nextbus stop IDs for a particular direction to an object mapping "arrival" Nextbus stop IDs to statistic values, e.g.:
```
{"version":"i1","start_time":1559840400,"end_time":1559862000,"stat":"median","routes":{"E":{"E____O_F00":{"5184":{"3092":0.9,"3095":2.6,"4502":5.3,"4529":7.0,"4516":8.0,"4518":9.5,"4504":11.0,"4534":13.0,"7283":14.6,"7794":19.9,"4506":22.1,"5234":24.1,"5239":29.5,"5241":33.0},"3092":{"3095":1.4,"4502":4.5,"4529":5.7,"4516":6.9,"4518":8.3,"4504":9.6,"4534":11.5,"7283":13.2,"7794":18.7,"4506":20.8,"5234":22.6,"5239":28.6,"5241":31.7},"3095":{"4502":2.4,
```

Each filename contains a version code, agency ID, date, stat ID, and time range, e.g: `trip-times_v1_sf-muni_2019-06-06_median_1000_1600.json.gz`

I think the median is probably the best single statistic to use for wait times and trip times, since it is not significantly affected by outliers in a small fraction of observations, whereas the average (mean) could  be significantly changed by a single outlier (https://en.wikipedia.org/wiki/Robust_statistics). If we use the average, I think there would be a high possibility of showing statistics that are obviously wrong that would reduce the user's trust in OpenTransit's statistics in general.

Similarly, min and max may also be significantly affected by outliers. Calculating the 10th percentile and 90th percentile allows showing the variability of trip times and wait times with better robustness to outliers. 

Each cache file can either store an individual statistic (e.g. median) or a list of statistics (e.g. 10th percentile, median, 90th percentile). When only an individual statistic is needed, storing it in its own file will improve performance when displaying statistics. For example, the file containing median trip times for all pairs of stops for one day is about 400 KB gzipped, but the file containing 10th percentile, median, and 90th percentile trip times for all pairs of stops for one day is about 800 KB gzipped. A smaller size will allow the user's browser to fetch the data from S3 faster. However, in other cases, it might be useful to be able to fetch multiple statistics at once. For example, if the front end needed to compute the interdecile range (90th percentile minus 10th percentile) for all pairs of stops, they would be able to fetch one file containing both the 90th percentile and the 10th percentile.

For wait time statistics, the median of the statistic values for all stops in a direction is also computed and stored under the special stop ID "median". This would make it easier for the frontend to show the typical wait time or wait time distribution for all stops along a route. It assumes that there is never a Nextbus stop ID named "median".

The statistics are computed over each time interval that is available in the frontend. In order to approximate statistics like median over a date range, the frontend could fetch the statistics for each date within the range, and compute the median or average of the statistics for each day. If we need to support large date ranges where it is infeasible for the frontend to fetch the statistics for each day, it may be necessary in the future to compute rollup stats for multiple days.

Currently I do not have access to write to the opentransit-precomputed-stats S3 bucket. S3 returns a `botocore.exceptions.ClientError: An error occurred (AccessDenied) when calling the PutObject operation: Access Denied` when I try to write to it.

In order for the frontend to be able to fetch these statistics directly from S3 without routing them through the Flask backend, the S3 bucket will need the following CORS configuration (configured on the Permissions tab in the S3 console): 

```
<?xml version="1.0" encoding="UTF-8"?>
<CORSConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
<CORSRule>
    <AllowedOrigin>*</AllowedOrigin>
    <AllowedMethod>GET</AllowedMethod>
</CORSRule>
</CORSConfiguration>
```

With the various performance optimizations implemented in this pull request and in previous pull requests, currently it takes about 30 seconds on my computer to compute the statistics for one day of wait times, and about 1 minute to compute the statistics for one day of trip times.


youngj:
With help from @EddyIonescu , the trip time and wait time stats for 2019-06-06 are now stored in the opentransit-precomputed-stats S3 bucket, e.g.:

https://opentransit-precomputed-stats.s3.amazonaws.com/wait-times/w1/sf-muni/2019/06/06/wait-times_w1_sf-muni_2019-06-06_median.json.gz
https://opentransit-precomputed-stats.s3.amazonaws.com/wait-times/w1/sf-muni/2019/06/06/wait-times_w1_sf-muni_2019-06-06_p10-median-p90_0700_1000.json.gz
https://opentransit-precomputed-stats.s3.amazonaws.com/trip-times/i1/sf-muni/2019/06/06/trip-times_i1_sf-muni_2019-06-06_median.json.gz
https://opentransit-precomputed-stats.s3.amazonaws.com/trip-times/i1/sf-muni/2019/06/06/trip-times_i1_sf-muni_2019-06-06_p10-median-p90_0700_1000.json.gz

The isochrone-map branch shows an example of how to use the cached wait times and trip times from JavaScript:

https://github.com/trynmaps/metrics-mvp/blob/isochrone-map/frontend/map/isochrone-worker.js#L137

compute_wait_times.py
```diff
@@ -0,0 +1,210 @@
...
+# are stored for each stop pair.
+#  p10 = 10th percentile
+#  median = 50th percentile
+#  p90 = 90th percentile
+#  p<Nm = probability of a wait time less than N minutes
+
+stat_groups = {
+    'p10-median-p90': ['p10','median','p90'],
+    'plt5m-30m': ['p<5m','p<10m','p<15m','p<20m','p<25m','p<30m'],
+    'median': 'median',
```
exxonvaldez: Can you make a stat group that has something like median, p90, and p<20m?  These are the values currently feeding into the scoring and grading code.

models/trip_times.py
```diff
@@ -60,3 +65,100 @@ def get_matching_trips_and_arrival_times(
 def sort_parallel(arr, arr2):
     sort_order = np.argsort(arr)
     return arr[sort_order], arr2[sort_order]
+
+DefaultVersion = 'i1'
```
exxonvaldez: I remember asking you about the signficance of "i", not sure if you want to put a comment about it for others who might be curious, up to you.

youngj: Just changed it from i1/w1 to v1 to be less confusing since the letter isn't really significant.

compute_wait_times.py
youngj: done, it is called median-p90-plt20m

PR #144, opened 2019-07-02
Basic method for computing OWL arrivals

marqcar:
This is the first pass at properly calculating arrival times for OWL routes.


marqcar:
Likely closing in favor of #145.

PR #145, opened 2019-07-02
Alternative arrival calculation for OWL routes

marqcar:
For standard routes we calculate arrival times for a given day starting at 3 AM local time so midnight-3am buses are associated with previous day. For OWL routes this causes unintuitive metrics as the route is split into two days. Define OWL routes and parameterize the start time so that we get metrics that are appropriate for each route type.  This is a refactor of #144.

### DIFF
--- a/compute_arrivals.py
+++ b/compute_arrivals.py
@@ -8,6 +8,49 @@ import time
 import boto3
 import gzip
 
+OWL_LINES = ['90', '91', 'K_OWL', 'L_OWL', 'M_OWL', 'N_OWL', 'T_OWL']
+
+def get_arrival_history(as_of_date: datetime, time_zone: pytz.timezone, increment: timedelta,
+                        agency: str, route_ids: list, start_hour: int, continue_index):
+    start_dt = time_zone.localize(datetime(as_of_date.year, as_of_date.month, as_of_date.day, hour=start_hour))
+    end_dt = start_dt + increment
+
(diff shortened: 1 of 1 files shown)


marqcar:
@youngj I was having trouble testing the route metrics on this branch but compute_arrivals.py is working correctly.

compute_arrivals.py
```diff
@@ -41,40 +86,23 @@
     continue_index = route_ids.index(args.continue_route) if args.continue_route is not None else None
```
youngj: continue_index won't work anymore after the route_ids are split into different lists. This functionality isn't really necessary so I'd suggest removing continue_index and the continue-route option.

compute_arrivals.py
```diff
@@ -41,40 +86,23 @@
...
-
-            arrivals_df = eclipses.find_arrivals(route_state, route_config, d, tz)
-
-            history = arrival_history.from_data_frame(agency, route_id, arrivals_df, start_time, end_time)
-
-            print(f'{route_id}: {round(time.time()-t1,1)} saving arrival history')
+        # for standard routes start each "day" at 3 AM local time so midnight-3am buses are associated with previous day
+        get_arrival_history(as_of_date=d,
+                            time_zone=tz,
+                            increment=incr,
```
youngj: May be simpler to make incr a local variable in get_arrival_history instead of a parameter

PR #146, opened 2019-07-04
Route wildcard paths to React in production environment; use nginx to cache/gzip responses; use GKE for production environment

youngj:
The frontend currently expects that any URL not starting with /frontend should be routed via React, but the production backend does not actually serve these URLs. For example if you go to http://opentransit.herokuapp.com/ and click "About", the URL becomes http://opentransit.herokuapp.com/about , but if you load http://opentransit.herokuapp.com/about directly, Flask returns a 404.

To fix this, this PR prefixes all Flask API paths with /api, so that any path not starting with /api or /frontend can be handled by the React main HTML page.

However, serving static assets from Flask probably will not result in good performance in production. For better performance, this PR allows using nginx in production to serve the static frontend assets and proxy requests starting with /api to the Flask backend. Nginx also gzips responses to reduce bandwidth and page load times, and caches API responses to reduce load on the Flask backend if multiple users are requesting the same URLs.

However, this proxying approach doesn't work well with Heroku which only allows one "web" process per application to serve HTTP requests. I couldn't find a way to get nginx and flask working together in Heroku, but it still seems valuable to support deploying to Heroku to make it easy for developers to deploy their branches for other people to see. To allow deploying to Heroku, this PR still supports the "all in one" approach where the compiled React assets are served by the Flask backend (without using nginx). In this case the Flask routes now support wildcard paths so that React-routed URLs like /about are rendered correctly.

To support nginx and flask proxying, I think the best approach is to use Kubernetes since we already have a GKE cluster running and a lot of remaining free credits on GCP. Using Kubernetes will also make it easy to implement cron jobs for computing arrival times and other stats (not implemented in this PR), and will make it easier to scale the frontend and backend processes independently as needed. This commit includes YAML files for creating the necessary deployments and services in a Kubernetes cluster via `kubectl apply -f <filename>`. In addition to deploying to GKE, this can also be tested locally using minikube (https://kubernetes.io/docs/setup/learning-environment/minikube/).

There is also a trigger in Google Cloud Build that will run the commands in cloudbuild.yaml whenever a commit is merged to the master branch, which builds the nginx and flask docker containers and updates the deployments in GKE. The GKE service is currently available at http://35.239.52.255/ . I'd recommend configuring a DNS A record like sf-muni.opentransit.city to point to 35.239.52.255 to make it easier to access.

This commit also optimizes the Dockerfile to use multi-stage builds and avoid reinstalling packages unless necessary, which should make `docker-compose build` faster most of the time and should also create smaller images (since the production images no longer need the React dev tools).

There is also a shell script `./docker-shell.sh` (Linux/Mac) or `docker-shell` (Windows) to make it easier to run a shell in the Flask dev container (a shortcut for `docker exec -it metrics-flask-dev bash`)


README.md
```diff
@@ -198,4 +195,13 @@ Commit your local changes to a feature branch (i.e. not master), then submit a p
 If you ever need to use a new pip library, make sure you run `pip freeze > requirements.txt`
 so other contributors have the latest versions of required packages.
 
-If you're on a Mac and you experience issues with Docker eating your CPU cycles and battery, set `CHOKIDAR_USEPOLLING=false` inside `docker-compose.yml`. Note: this will disable hot reloading.
+If you're on a Mac and you experience issues with Docker eating your CPU cycles and battery, create a docker-compose.override.yml file like this:
+
```
EddyIonescu: Wish I'd known this before, thanks for adding this!

README.md
```diff
@@ -198,4 +195,13 @@ Commit your local changes to a feature branch (i.e. not master), then submit a p
...
+```
+services:
+  react-dev:
+    environment:
+      CHOKIDAR_USEPOLLING: "false"
+```
+
+Note: this will disable hot reloading so you will need to restart `docker-compose up` after making changes to the frontend.
```
exxonvaldez: I had Arun add this part about CHOKIDAR_POLLING for me.  Without hot reloading, my expereince is that the browser automatically refreshes the page after a source file is saved (and all page state gets reset).  My expectation with hot reloading is that the new code is loaded without a page refresh (but I haven't used it).

youngj:
I think all that we need to do to migrate to Google Cloud is to configure a domain name 35.239.52.255 and stop sending people to opentransit.herokuapp.com . Also once we have a domain name we can configure HTTPS using a Google load balancer.

PR #148, opened 2019-07-04
always load leaflet.css with https:// URL since http URL redirects to https anyway

youngj:
There was a HTTP 307 redirect each time the app is loaded from a http URL, slowing down the page load slightly.


youngj:
FYI, URLs starting with // will be loaded from the same protocol used by the page itself, either http:// or https://

PR #149, opened 2019-07-04
reduce CPU/battery use on OS X; share .aws folder in flask container; update readme

youngj:
A few changes to improve the developer experience: 

- Disable CHOKIDAR_USEPOLLING in docker-compose.yml.  This reduces CPU/battery usage a lot when using Docker on OS X. Since CHOKIDAR_USEPOLLING is necessary in Docker on Windows to enable automatic recompilation of frontend code, explained how to enable it via docker-compose.override.yml
- Document how to share an .aws folder in the flask-dev docker container using docker-compose.override.yml. This makes it easier to run scripts that write to S3 which need an AWS access key.
- Various updates and corrections to README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,37 @@ Scrape timetables from GTFS data stored locally in `inpath` and extract them to 
 python gtfs_scraper.py --inpath=path/to/google_transit
 ```
 
+## AWS Credentials
+
+If you need to write files to S3 from your development environment (e.g. running compute_arrivals.py with the --s3 flag),
+you will need to get AWS credentials from one of the project admins.
+
+The scripts load AWS credentials from the standard locations, such as ~/.aws/credentials, using the "default" profile by default.
+
+The ~/.aws/credentials file should look something like this:
+
+```
+[default]
+aws_access_key_id = ....
+aws_secret_access_key = ....
+```
+
+However, if you are using Docker, the command line scripts will load the ~/.aws/credentials file from within the flask-dev Docker container.
+
+To make it easy to access AWS credentials in your Docker container, you can create a .aws directory
+somewhere on your host machine (e.g. in your host home directory) and share it with your Docker container
+by creating a docker-compose.override.yml file like this:
+
+```
+version: "3.7"
+services:
+  flask-dev:
+    volumes:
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -210,13 +204,16 @@ Commit your local changes to a feature branch (i.e. not master), then submit a p
...
+version: "3.7"
 services:
   react-dev:
     environment:
-      CHOKIDAR_USEPOLLING: "false"
+      CHOKIDAR_INTERVAL: "5000"
 ```
 
-Note: this will disable hot reloading so you will need to restart `docker-compose up` after making changes to the frontend.
+Note: if you disable CHOKIDAR_USEPOLLING, you will need to restart `docker-compose up` after making changes to the frontend.
```
exxonvaldez: https://medium.com/devcupboard/dockerizing-development-environment-in-mac-performance-debugging-dac16aa4524
This article says setting Chokidar_usepolling=false just switches to a different polling implementation (at least for Macs), which works better on my Mac (only 7% CPU utilization).  With Chokidar true and an interval of 2500, the utilization was 18%.  There was no different in reload behavior with true vs false -- in both cases, the page reloaded when a file was edited.

The above article also mentions turning off watching of the node_modules, but doesn't explain how.

youngj: Thanks for the information, I didn't realize that CHOKIDAR_USEPOLLING wasn't even necessary to enable automatic recompilation in Docker on Mac OS X. It is necessary to get automatic recompilation in Docker on Windows. Since most people are probably using OS X, I removed it from docker-compose.yml and updated the readme to explain how to enable it on Windows via docker-compose.override.yml.

I also added some console.log statements in chokidar to see what it is doing. Chokidar is being called with a "ignored" option that is supposed to ignore the node_modules directory, but chokidar is calling fs.watchFile on a lot of files in the node_modules directory anyway. This seems like a bug in chokidar. The latest chokidar release 3.0.0 says it has "Massive CPU & RAM consumption improvements", so perhaps this issue is fixed there, but the latest version of webpack-dev-server requires chokidar 2.x.  I think webpack-dev-server would need to update to support chokidar 3.x in order for us to benefit from the massive CPU & RAM consumption improvements.

PR #151, opened 2019-07-07
frontend/public directory is not used in production

youngj:
The frontend/public directory includes source files that are not used in production. Only the frontend/build directory is needed to serve the production frontend.


youngj:
Good idea, added note in the readme about Create React App with link to documentation.

PR #152, opened 2019-07-09
use Google managed SSL certificate in GKE

youngj:
I already applied these YAML files were applied to the GKE cluster to create a Google-managed certificate for muni.opentransit.city and an HTTPS load balancer that routes https://muni.opentransit.city/ and http://muni.opentransit.city/ to the web service. The web service is now exposed on port 80 on each host in the cluster instead of using a TCP load balancer like before (which didn't support SSL termination). 

With this configuration Google will automatically renew the SSL certificate as long as muni.opentransit.city points to the load balancer.

I generally followed the instructions at https://cloud.google.com/kubernetes-engine/docs/how-to/managed-certs but without creating a reserved IP address (not sure if this means that GKE will automatically delete the IP address if the Ingress is deleted)

### DIFF
--- a/kubernetes/managed-certificate.yaml
+++ b/kubernetes/managed-certificate.yaml
@@ -1,0 +1,7 @@
+apiVersion: networking.gke.io/v1beta1
+kind: ManagedCertificate
+metadata:
+  name: muni-certificate
+spec:
+  domains:
+    - muni.opentransit.city
--- a/kubernetes/web-ingress.yaml
+++ b/kubernetes/web-ingress.yaml
@@ -1,0 +1,10 @@
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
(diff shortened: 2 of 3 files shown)


exxonvaldez:
> looks like Chrome doesn't like that the site's requesting content from tile.stamen (for the map) via an http endpoint, so I'm guessing that'll be needed to get the shiny lock.

Good catch.  Per http://maps.stamen.com:
> If you'd like to display these map tiles on a website that requires HTTPS, use our tile SSL endpoint by replacing `http://tile.stamen.com` with `https://stamen-tiles.a.ssl.fastly.net`

They also have a protocol-agnostic url, but seems like best practice is https everywhere?

youngj:
Probably easiest to load the map tiles from https everywhere, I can make another PR for that.

PR #153, opened 2019-07-10
Show speed along route

exxonvaldez:
Porting over some more stuff from my experimental branch.  To show speeds along a route, we need to fetch the precomputed trip times from S3.  I used Jesse's isochrone branch as a guide for implementing storage of the trip times and wait times in Redux.  helpers/precomputed.js has some of the same utility methods.

Currently, I'm using the `handleGraphParams` action creator as the trigger for loading the cache, so that when a route is chosen, we make sure the cache is loaded.  Loading the Dashboard screen will probably also have to trigger loading since the RouteTable will use the cache also.

Plotting speeds will probably need some refinement.  Speeds are plotted whether just one or all directions are shown.  If multiple directions are visible, then you can't see all the speeds.

Should be deployed to http://ot-terence.herokuapp.com

### DIFF
--- a/frontend/src/actions/index.js
+++ b/frontend/src/actions/index.js
@@ -1,5 +1,6 @@
 import axios from 'axios';
 import { metricsBaseURL } from '../config';
+import { getTimePath, getStatPath } from '../helpers/precomputed.js';
 
 export function fetchGraphData(params) {
   return function(dispatch) {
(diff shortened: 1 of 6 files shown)


frontend/src/reducers/routesReducer.js
```diff
@@ -13,6 +13,8 @@ const initialState = {
     date: '2019-06-06',
   },
   spiderLatLng: null, 
+  tripTimesCache: {},
```
youngj: The organization of data within the "state" seems difficult to understand/remember. For example the list of routes is state.routes.routes and the data returned by /api/metrics is state.fetchGraph.graphData . I'm not really sure what the best practices are for organizing state. Would it work to have one top level "reducer" instead of separate reducers for fetchGraph/routes/page, so the list of routes could be state.routes , the graph data state.graphData , and trip times cache state.tripTimesCache ? Probably not an issue to resolve in this PR.

frontend/src/reducers/routesReducer.js
```diff
@@ -25,6 +27,12 @@ export default (state = initialState, action) => {
       return { ...state, fetched: true, graphParams: Object.assign({}, state.graphParams, action.payload) };
     case 'RECEIVED_ROUTES_ERROR':
       return state;
+    case 'RECEIVED_PRECOMPUTED_TRIP_TIMES':
+      return { ...state, fetched: true, tripTimesCache: { ...state.tripTimesCache,
```
youngj: It looks like the fetched property is only written and not read at the moment. It seems like state.routes.fetched is being set to true for several different events so it's not clear what its meaning is, probably could remove it

frontend/src/components/MapStops.jsx
```diff
@@ -35,11 +38,134 @@ class MapStops extends Component {
...
+    // refactor time/date handling along with actions/index.js
+    
+    let timeStr = graphParams.start_time ? graphParams.start_time + '-' + graphParams.end_time : '';
+    let dateStr = graphParams.date;    
+    
+    if (!this.props.tripTimesCache[dateStr + timeStr + 'median']) {
+      return -1;
+    }
+    
+    const tripTimesForRoute = this.props.tripTimesCache[dateStr + timeStr + 'median'].routes[routeID];
```
youngj: Could create local variable for `this.props.tripTimesCache[dateStr + timeStr + 'median']` since used twice

frontend/src/actions/index.js
```diff
@@ -76,6 +77,52 @@ export function fetchRoutes() {
...
 
+export function fetchPrecomputedWaitAndTripData(params) {
+  return function(dispatch, getState) {
+
+    let timeStr = params.start_time ? params.start_time + '-' + params.end_time : '';
+    let dateStr = params.date;
+    
+    const tripTimesCache = getState().routes.tripTimesCache;  
+  
+    let tripTimes = tripTimesCache[params.dateStr + timeStr + 'median'];
```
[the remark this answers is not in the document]
exxonvaldez: Yes, thanks, good catch.

frontend/src/actions/index.js
```diff
@@ -76,6 +77,52 @@ export function fetchRoutes() {
...
+      axios.get(s3Url)
+      .then(response => {
+        dispatch({ type: 'RECEIVED_PRECOMPUTED_TRIP_TIMES', payload: [response.data, dateStr + timeStr + 'median'] })
+      })
+      .catch(err => { /* do something? */ })
+    }
+  
+    const waitTimesCache = getState().routes.waitTimesCache;  
+
+    let waitTimes = waitTimesCache[params.dateStr + timeStr + 'median'];
```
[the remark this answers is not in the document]
exxonvaldez: Also fixed.

frontend/src/reducers/routesReducer.js
exxonvaldez: That's a good question, I'm not sure what the answer is.  There is one child reducer just for "pages", which I'm assuming is a redux-first-router idiom (so that's state.pages.*).  We could probably combine  `routesReducer` (where I'm adding the new stuff) and `fetchGraphReducer` (which only has the metrics results) into one other child reducer (state.[someReducerName].*).

frontend/src/components/MapStops.jsx
exxonvaldez: Good idea, fixed.

frontend/src/reducers/routesReducer.js
exxonvaldez: Yeah, I've been copying "fetched" around for a bit, because I didn't know what it was for and was new to reducers.  There are also state.routes.fetching and state.graphData.fetching flags that are initialized to false and never set.  Perhaps at some point we will put in "loading" indicators, but we'd need a boolean per network call.  I thought about scrubbing it out, but it's also a vague reminder to do indicators.

exxonvaldez:
> I like the UI, it makes it easy to see slow segments of a route, and also see how they change over time by selecting different time ranges.

I didn't think to test that, good thing it worked!  We could show time of day charts if you hover over a segment of a route, and a bar chart showing speed for the last N weekdays/weekend days.   Not sure if that's a backend API or pure front end.

frontend/src/reducers/routesReducer.js
exxonvaldez: I just decided to scrub these, as it seems more like clutter than a reminder.

PR #154, opened 2019-07-10
Refactoring speed computation and using actual median waits in RouteTable

exxonvaldez:
I moved the speed calculation code that deals with the trip times cache back into `precomputed.js` so that it can be reused.  The diff is hard to read, but basically `precomputed.js` is back to being like Jesse's isochrone functions.  I put in some JSDocs for each function.  The differences are that fetching of the s3 files is done elsewhere, and date and time are passed in using graphParams.

The routes table on the Dashboard screen is now using the wait times cache, to show the median wait for each route.  Actually, it's average of the median wait in each direction.

### DIFF
--- a/frontend/src/components/RouteTable.jsx
+++ b/frontend/src/components/RouteTable.jsx
@@ -229,7 +229,7 @@ function RouteTable(props) {
       if (!waitForDir) {
           return NaN;
       }
-      return total + getWaitTimeForDirection(waitTimesCache, graphParams, route.id, direction.id).median;  
+      return total + waitForDir.median;  
     }, 0);
     return sumOfMedians/directions.length;
   }
--- a/frontend/src/components/RouteTable.jsx
+++ b/frontend/src/components/RouteTable.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import clsx from 'clsx';
 import PropTypes from 'prop-types';
 import { lighten, makeStyles } from '@material-ui/core/styles';
@@ -16,11 +16,12 @@ import Tooltip from '@material-ui/core/Tooltip';
 import FilterListIcon from '@material-ui/icons/FilterList';
 
 import { filterRoutes } from '../helpers/routeCalculations';
+import { getWaitTimeForDirection } from '../helpers/precomputed';
 import { connect } from 'react-redux';
 import { push } from 'redux-first-router'
 import Link from 'redux-first-router-link'
 
-import { handleGraphParams } from '../actions';
+import { handleGraphParams, fetchPrecomputedWaitAndTripData } from '../actions';
 
 function desc(a, b, orderBy) {
   if (b[orderBy] < a[orderBy]) {
@@ -176,6 +177,10 @@ function RouteTable(props) {
   const [selected, setSelected] = React.useState([]);
   const dense = true;
 
+  useEffect(() => {
+    props.fetchPrecomputedWaitAndTripData(props.graphParams); // like componentDidMount, runs only once
+  }, []);
+  
   function handleRequestSort(event, property) {
     const isDesc = orderBy === property && order === 'desc';
     setOrder(isDesc ? 'asc' : 'desc');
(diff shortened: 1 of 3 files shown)


frontend/src/helpers/precomputed.js
```diff
@@ -2,84 +2,103 @@
...
+  {
+    return tripTimeValues;
+  }
+  if (stat === 'p10')
+  {
+    return getTripTimeStat(tripTimeValues, 0);
+  }
+  if (stat === 'p90')
+  {
+    return getTripTimeStat(tripTimeValues, 2);
```
hathix: The 0 and 2 feel like "magic numbers" here. Maybe make it an enum so that you clarify that index 0 == 10th percentile and index 2 == 90th percentile?

frontend/src/helpers/precomputed.js
```diff
@@ -2,84 +2,103 @@
...
+  const statValues = {};
+  for (let endStopId in tripTimeValues)
+  {
+    statValues[endStopId] = tripTimeValues[endStopId][index];
+  }
+  return statValues;
 }
 
+/**
+ * Maps the given stat to a stat group (part of the file path).
```
hathix: Give an example? This didn't make sense to me on first read

frontend/src/components/RouteTable.jsx
```diff
@@ -210,6 +215,25 @@ function RouteTable(props) {
...
+  /**
+   * Averages together the median wait in all directions for a route.
+   * 
+   * @param {any} waitTimesCache
+   * @param {any} graphParams
+   * @param {any} route
+   */
+  function getAverageOfMedianWait(waitTimesCache, graphParams, route) {
+    const directions = route.directions;
+    const sumOfMedians = directions.reduce((total, direction) => {
```
hathix: Hey, that's some nice functional programming!

frontend/src/components/RouteTable.jsx
```diff
@@ -290,65 +306,16 @@ function RouteTable(props) {
...
- * These are static placeholder precomputed average waits per route values.  This will be replaced
- * eventually by values generated from the precomputed wait times for all stops on a route.
- * 
- * These are used to draw routes in three widths by frequency: widest is the top third most frequent,
- * medium width for middle third of frequency, and thinnest for the third least frequent.
- */
-function getAllWaits() {
-  const allWaits = 
-
-[{"routeID":"38BX","wait":169.84285714285716},{"routeID":"38AX","wait":159.70769230769233},
```
hathix: So are there any more hardcoded waits or no?

frontend/src/components/RouteTable.jsx
```diff
@@ -210,6 +215,25 @@ function RouteTable(props) {
...
+   * @param {any} route
+   */
+  function getAverageOfMedianWait(waitTimesCache, graphParams, route) {
+    const directions = route.directions;
+    const sumOfMedians = directions.reduce((total, direction) => {
+      const waitForDir = getWaitTimeForDirection(waitTimesCache, graphParams, route.id, direction.id);
+      if (!waitForDir) {
+          return NaN;
+      }
+      return total + getWaitTimeForDirection(waitTimesCache, graphParams, route.id, direction.id).median;
```
youngj: total + waitForDir.median

exxonvaldez: > total + waitForDir.median

Whoops, will fix.

frontend/src/components/RouteTable.jsx
exxonvaldez: > So are there any more hardcoded waits or no?

Actually, yes.  The spider map has hardcoded waits to choose thick/normal/thin lines for routes based on the ranked wait (most frequent, middle frequency, least frequent).  This could be made dynamic although probably almost unnoticeable.

frontend/src/helpers/precomputed.js
exxonvaldez: You're right, these numbers are tied to the fact that there is a file with the stats p10, median, and p90 bundled together and used by the isochrone branch, as well as another file that has just the median alone.  The best documentation for the precomputed stats and their file structure and json structure is here: https://github.com/trynmaps/metrics-mvp/pull/143 .  Things may shake out a bit as the UI shifts to a different stat group.  I'll see what I can do to beef up the comments: mainly I'll add that link.

frontend/src/helpers/precomputed.js
exxonvaldez: Will update the comments.  We have a "median" stat group and a "p10-median-p90" stat group used by the isochrone branch.  There's also "median-p90-plt20m" which the UI will use eventually.  So the idea is when fetching a particular stat, this function tells you which stat group to use.

exxonvaldez: Also the function favors smaller stat groups as they are smaller files.

PR #159, opened 2019-07-11
use https to load map tiles to avoid mixed content warning

youngj:
Updated base URL to https://stamen-tiles.a.ssl.fastly.net . Also, my editor automatically removed extra whitespace at the end of line.

### DIFF
--- a/frontend/src/components/MapStops.jsx
+++ b/frontend/src/components/MapStops.jsx
@@ -39,7 +39,7 @@ class MapStops extends Component {
         );
       });
       route.unshift(this.populateSpeed(routeStops, direction_id));
-      
+
       // put polyline first so clickable markers are drawn on top of it
       route.unshift(<Polyline key={'polyline-' + direction_id} color="white" weight={10} positions={routeStops[direction_id]} opacity={1} />);
     }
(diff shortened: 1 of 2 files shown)


frontend/src/components/MapStops.jsx
```diff
@@ -256,16 +256,16 @@ class MapStops extends Component {
       <Map center={position || SF_COORDINATES} zoom={zoom || ZOOM} style={mapClass}>
         <TileLayer
           attribution='Map tiles by <a href="http://stamen.com">Stamen Design</a>, under <a href="http://creativecommons.org/licenses/by/3.0">CC BY 3.0</a>. Data by <a href="http://openstreetmap.org">OpenStreetMap</a>, under <a href="http://www.openstreetmap.org/copyright">ODbL</a>.'
-          url="http://tile.stamen.com/toner-lite/{z}/{x}/{y}.png"
+          url="https://stamen-tiles.a.ssl.fastly.net/toner-lite/{z}/{x}/{y}.png"
```
hathix: Is this the only non-whitespace change? If so, LGTM

youngj: There's a similar change in MapSpider.jsx, but that's all

PR #161, opened 2019-07-11
Remove --continue-route parameter

youngj:
Remove the --continue-route parameter on compute_arrivals.py since it probably isn't necessary anymore.

### DIFF
--- a/compute_arrivals.py
+++ b/compute_arrivals.py
@@ -84,7 +84,7 @@ if __name__ == '__main__':
         compute_arrivals_for_date(
             d, start_hour=3, tz=tz,
             agency=agency,
-            route_ids=list(set(route_ids) - set(owl_routes)),
+            route_ids=[r for r in route_ids if r not in owl_routes],
             s3=args.s3
         )
         # owl routes run from 1AM to 5AM so they are associated with the same day
--- a/compute_arrivals.py
+++ b/compute_arrivals.py
@@ -2,7 +2,7 @@ from models import arrival_history, util, trynapi, nextbus, eclipses
 import json
 import math
 import argparse
-from datetime import datetime, timedelta
+from datetime import datetime, date, timedelta
 import pytz
 import time
 import boto3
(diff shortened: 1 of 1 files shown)


compute_arrivals.py
```diff
@@ -2,18 +2,20 @@
...
-from datetime import datetime, timedelta
+from datetime import datetime, date, timedelta
 import pytz
 import time
 import boto3
 import gzip
 
 OWL_LINES = ['90', '91', 'K_OWL', 'L_OWL', 'M_OWL', 'N_OWL', 'T_OWL']
 
-def get_arrival_history(as_of_date: datetime, time_zone: pytz.timezone, increment: timedelta,
```
hathix: Why are we getting rid of the increment parameter?

compute_arrivals.py
```diff
@@ -58,7 +57,6 @@ def get_arrival_history(as_of_date: datetime, time_zone: pytz.timezone, incremen
     parser.add_argument('--start-date', help='Start date (yyyy-mm-dd)')
     parser.add_argument('--end-date', help='End date (yyyy-mm-dd), inclusive')
     parser.add_argument('--s3', dest='s3', action='store_true', help='store in s3')
-    parser.add_argument('--continue-route', help='Continue at a particular route after an error')
```
hathix: I agree this param is not needed

compute_arrivals.py
```diff
@@ -81,28 +79,18 @@ def get_arrival_history(as_of_date: datetime, time_zone: pytz.timezone, incremen
...
-    continue_index = route_ids.index(args.continue_route) if args.continue_route is not None else None
-
     for d in dates:
         # for standard routes start each "day" at 3 AM local time so midnight-3am buses are associated with previous day
-        get_arrival_history(as_of_date=d,
-                            time_zone=tz,
-                            increment=incr,
-                            agency=agency,
-                            route_ids=list(set(route_ids) - set(owl_routes)),
-                            start_hour=3,
```
hathix: Are we getting rid of the `start_hour` fix for buses running before 3am? If so why?

youngj: It's still there. I just changed the order of parameters to put the date-related parameters together

compute_arrivals.py
youngj: It is not needed as a parameter because it is always the same (one day).

PR #172, opened 2019-07-11
Changed buttons to material ui Buttons

juanmanual:


exxonvaldez:
The Material UI buttons look a lot better than native.  My suggestions:

- Include a "gps" icon in the "routes near me" button: https://material.io/tools/icons/?icon=gps_fixed&style=baseline
- Align the buttons in a horizontal row with some space between them.

Floating action buttons seems worth a try for the higher contrast against the map, though it's described as being reserved for the one primary action on the screen.  I think these contained buttons have good contrast already though, with the drop shadow.

PR #176, opened 2019-07-17
Introduction of RouteSummary component (like NYC Busstats route overview)

exxonvaldez:
Screen Shot 2019-07-16 at 11 44 34 PM
Screen Shot 2019-07-16 at 11 46 47 PM

Try this out at https://ot-terence.herokuapp.com/

This PR introduces the `RouteSummary` component (first screenshot), which is like the NYC Bus Stats route overview.  Some of you have seen this before in prototype form.  It charts trip time along a route in a direction.  There are also displays of median speed, median wait (plus rankings and scores for these two values), and an overall grade and score for a route.  `RouteSummary` is shown when the user has selected a route and direction but not start and end stops.

The `RouteTable` now has speeds and scores (second screenshot).

For now, I am using precomputed, hardcoded route lengths instead of trying to determine the route length from GTFS data files on the fly.  There's a lot of complexity around fetching, parsing, and traversing the routes/trips/shapes files.  Also the GTFS data does not change very often, and the route lengths even less often.

With these hardcoded route lengths, it is possible to use Jesse's precomputed trip times to do on the fly speed and score calculations for all of the routes, which `RouteSummary` needs for rankings.

Note that the `RouteSummary` uses Bootstrap widgets and needs updating to Material UI, but the React-vis chart is still current.  This chart can be used to plot scheduled trip times for a route alongside actual trip times, when the scheduled trip time data becomes available to the front end.

Both `RouteSummary` and `RouteTable` receive the current `graphParams` and wait/trip caches in their props, and then call shared helper functions in `routeCalculations.js` to get wait times, speeds, and scores.

I am using what I call "heuristics" (simple data-driven rules) already in `routeCalculations` to ignore problematic beginning and ending stops of routes when it comes to travel time computations.  Sometimes a line never actually travels to a last stop in its config, or only certain directions actually go to the last stop, or only at certain times of day.  

Possible next steps from here:

- Create a "data quality" test page (not part of the main UI) that plots trip times and wait times for all routes in all directions, plus that frequency vs speed quadrant chart.  This should be the last of the work in  progress code that has been sitting in my experimental branch.

- Refine the display of grades and scores.  Overall scores for a route could be enhanced to include  travel variability and long wait probability metrics.  For a rating a trip between two stops, perhaps we can simplify it to give the user broad indications of which metrics of the trip are good/okay/bad.

- Refine other parts of the UI, as noted in Neel's github issues.


exxonvaldez:
Neel and Jesse, thanks for the good feedback.

Screen Shot 2019-07-20 at 1 01 03 AM

>     * When I open the routes page, I don't see the grade -- I need to select "outbound" or "inbound." It's not obvious that you have to do this. Either default to outbound or inbound, or force the user to choose.

Jesse made a similar comment.  I decided to show the `RouteSummary` even when direction is not yet chosen.  Busstats is not like this, they force a direction.  We'll show route-level stats: score, wait, and speed -- basically the same stats from the `RouteTable`, but with more details, like rankings and subscores.

>     * Can we add some color coding? Currently both panels are light blue. It seems like a prime opportunity to make them green/yellow/red based on the score.

Light blue is the Bootstrap "info" color, they have warning, error, etc.  I switched `RouteSummary` to use Material UI widgets, which has a similar palette but that's really more of a primary/secondary/other button thing.  I chose some Material UI colors: green/light green/yellow/red, with darker shades for better quartiles.  For route score, I got rid of the letter grade, it seemed like using a color was better.  I think this still needs symbols/icons/textures to be accessible to color blind users, like a 4-star version of https://material-ui.com/components/rating/ or a set of 4 icons or symbols to represent the quartiles.  Will create an issue for accessibility.
 
>     * Mention that the /100 is points. Not clear where that score comes from.

I'm reluctant to use the word "points", maybe I can put "Score: 52/100" in front of the subscores.  We're going to need in-UI explanations for our scoring system, we could put toggles on each box that reveals the subscores and how they were calculated.   Now that there are quartile colors, the subscores are less important.  Maybe push them to the bottom right corner of each box?

>     * Can we compare the median speed to the average speed of cars, walking, bikes, and the muni system overall? [www.busstat.nyc](http://www.busstat.nyc) does something like this.

I looked all over and couldn't find that, but it sounds familiar.   Seems like an "infobubble" (or material UI Popper https://material-ui.com/components/popper/ ) that shows the speed on a scale would be good.  I'm not sure if there is a current stat for the Muni overall speed, just averages for trolley bus/diesel bus/light rail from about 5 years ago.

>     * Not required for this PR, but it might be cool to stretch the graph's X-axis by the distance between stops. This would be especially useful for express routes that go from downtown straight to the outskirts, like:

Jesse asked for this also, but we'll need true distances to each stop to do this.  Will open an issue.

>     * Call it "route grade" instead of "route summary"

I added a "Route score" box.  There should be something that divides the direction-independent stats from the chart that is direction-dependent. 

>     * Is it possible to add x and y axis labels to the graph? The "minutes" label would look better off to the left of the axis

Done.

>     * Give a label to the x axis; currently it only shows the stop numbers. Either label the stops on the x-axis (preferred) or just say "stop number" on the x-axis.

There isn't room to label the stops (sometimes there are like 50+), so "Stop Number" is added.

Jesse wrote:

> After the user selects a start stop, it seems like there isn't a way to unselect it and get back to the route summary, though that's probably out of scope for this PR.

> Currently the control panel and the route overview both use a lot of space, which requires most desktop users to scroll to see the map and the graph(s), so at some point we should probably update the layout to use the screen space more efficiently.

Agree on both points.  One thought I had was to replace the bread crumbs with "Chips" https://material-ui.com/components/chips/ .  The idea being that you could delete a chip to clear selections.

Another option is not to use Chips, but instead move the route/direction/stops dropdowns up into the blue header bar to replace the breadcrumbs.  NYC bus stats kind of does this, their pickers are in a row at the top, but their stop pickers are only via the map, which isn't good from an accessibility standpoint.  We could also add a close (X) button at the top of the Info component that "dismisses" the info, resetting you back to RouteSummary with stops deselected.

frontend/src/components/RouteSummary.jsx
```diff
@@ -0,0 +1,251 @@
...
+    const direction_id = graphParams.direction_id;
+    
+    const distObj = allDistances.find(distObj => distObj.routeID === route_id);
+    dist = distObj ? distObj.distance : null;
+
+    tripTime = getEndToEndTripTime(props, route_id, direction_id);
+    
+    if (dist <= 0 || isNaN(tripTime)) { speed = "?"; } // something wrong with the data here
+    else {
+      speed = Number.parseFloat(dist) / tripTime * 60.0 / 1609.344;  // initial units are meters per minute, final are mph
```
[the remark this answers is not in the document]
exxonvaldez: Decided to go with a new helper function metersToMiles, seems to improve readability where it gets used.

frontend/src/helpers/routeCalculations.js
```diff
@@ -167,12 +170,306 @@ export function ignoreFlag(routeID, directionID, flagName) {
...
+  const {tripTimesForFirstStop, directionInfo} = getTripTimesUsingHeuristics(props, routeID, directionID);
+
+  if (!tripTimesForFirstStop) { return []; } // no precomputed times
+
+  const route = props.routes.find(route => route.id === routeID);
+
+  const dataSeries = directionInfo.stops.map((stop, index) => { return {
+    x: index,
+    y: tripTimesForFirstStop[stop] ? tripTimesForFirstStop[stop] : 0,
+        title: route.stops[stop].title
```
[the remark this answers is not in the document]
exxonvaldez: Thanks, fixed.

frontend/src/helpers/routeCalculations.js
```diff
@@ -167,12 +170,306 @@ export function ignoreFlag(routeID, directionID, flagName) {
...
+
+  let lastStop = null;
+
+  /*
+   * For determining end to end trip time, but doesn't currently affect computation of route length,
+   * which is based on best guess as to the right GTFS shape. And the shape is not linked to the stops,
+   * it's just coordinates and distance along route, so more logic would be needed to "trim" the shape
+   * if stops are ignored.
+   */
+  if (Number.isInteger(ignoreLast)) {
```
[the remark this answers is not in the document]
exxonvaldez: Good point.  Switched the heuristics to also treat stop IDs as strings.

frontend/src/helpers/routeCalculations.js
```diff
@@ -167,12 +170,306 @@ export function ignoreFlag(routeID, directionID, flagName) {
...
+  }
+
+  let tripTimesForFirstStop = tripTimesForDir[firstStop];
+
+  // if this stop doesn't have trip times (like the oddball J direction going to the yard, which we currently ignore)
+  // then find the stop with the most trip time entries
+
+  if (!tripTimesForFirstStop) {
+    //console.log("No trip times found for " + routeID + " from stop " + firstStop + ".  Using stop with most entries.");
+    tripTimesForFirstStop = Object.values(tripTimesForDir).reduce((accumulator, currentValue) =>
```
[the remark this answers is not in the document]
exxonvaldez: Good question.  I probably tried that first, but maybe ran into an edge case where the beginning section of a route had sporadic data.  Picture the 9, which has multiple terminals depending on time of day.  The directionInfo.stops first lists stops only used for starting terminal 1, then stops only used for starting terminal 2, then the stops for the "main" terminal.  But maybe you only get a partial run for a vehicle that was on Terminal 1 and then it went to the yard.

frontend/src/helpers/routeCalculations.js
```diff
@@ -270,10 +274,12 @@ export function getTripDataSeries(props, routeID, directionID) {
 
   const route = props.routes.find(route => route.id === routeID);
 
-  const dataSeries = directionInfo.stops.map((stop, index) => { return {
-    x: index,
+  // omit the first stop since trip time is always zero
+
+  const dataSeries = directionInfo.stops.slice(1).map((stop, index) => { return {
+    x: index + 1,
     y: tripTimesForFirstStop[stop] ? tripTimesForFirstStop[stop] : 0,
```
youngj: Using a y value of 0 makes the graph look weird if there is no data for a stop (e.g. last stop of 5 inbound). I think that dropping those points looks better, e.g:

```
  const dataSeries = [];

  directionInfo.stops.map((stop, index) => {
     if (tripTimesForFirstStop[stop]) {
       dataSeries.push({
        x: index + 1,
        y: tripTimesForFirstStop[stop],
        title: route.stops[stop].title,
      });
    }
  });

  return dataSeries;
```

frontend/src/components/RouteSummary.jsx
```diff
@@ -97,63 +105,89 @@ function RouteSummary(props) {
...
+      <div style={{ padding: 12 }}>
+      <Grid container spacing={3}>
+
+        <Grid item xs>            
+          <Paper className={classes.grade}>
+          <Typography variant="overline">Route score</Typography><br/>
+          <Grid container justify="space-between">
+          <Grid item>
+          <Typography variant="h3" display="inline">{grades.totalScore}</Typography>
+          <Typography variant="h5" display="inline">/{grades.highestPossibleScore}</Typography>
```
youngj: I think the route score would be more easily understandable if they were out of 100 instead of 200.

frontend/src/components/RouteSummary.jsx
```diff
@@ -97,63 +105,89 @@ function RouteSummary(props) {
...
+
+        <Grid item xs>            
+          <Paper className={classes.wait}>
+
+            <Typography variant="overline">Median wait</Typography>
+            <Typography variant="h3">{ route.wait < 0 ? "?" : route.wait.toFixed(0) } minutes</Typography>
             
-            { waitRanking ? ` (#${waitRanking} of ${routes.length} for shortest wait)` : null }
+            { waitRanking ? `#${waitRanking} of ${routes.length} for shortest wait` : null }
             { waitRanking ? <small>&nbsp;({ grades.medianWaitScore }/100)</small>
```
youngj: Showing wait scores and speed scores might confuse people who don't know what the x/100 means. Could we hide the wait scores and speed scores, and show them in a tooltip if someone hovers over the route score (as part of an explanation of how the route score is calculated)?

frontend/src/components/RouteSummary.jsx
```diff
@@ -97,63 +105,89 @@ function RouteSummary(props) {
...
+          </Grid>
+
+        <Grid item xs>            
+          <Paper className={classes.wait}>
+
+            <Typography variant="overline">Median wait</Typography>
+            <Typography variant="h3">{ route.wait < 0 ? "?" : route.wait.toFixed(0) } minutes</Typography>
             
-            { waitRanking ? ` (#${waitRanking} of ${routes.length} for shortest wait)` : null }
+            { waitRanking ? `#${waitRanking} of ${routes.length} for shortest wait` : null }
```
youngj: The total number of routes for the wait time ranking (76) is different from the number of routes for the route score or trip time ranking (75). I think it would be clearer to use the same number of routes for all 3 boxes.

frontend/src/components/RouteSummary.jsx
```diff
@@ -194,7 +195,7 @@ function RouteSummary(props) {
             
             <br/>
 
-            Length: { (dist / 1609.344).toFixed(1) } miles
+            Length: { metersToMiles(dist).toFixed(1) } miles
```
youngj: The "length" line makes the speed box taller than the other 2 boxes, which looks odd when the screen is wide enough to display all 3 boxes on one line. Could probably move the length somewhere else so that all 3 boxes have the same height.

frontend/src/components/RouteSummary.jsx
```diff
@@ -0,0 +1,294 @@
...
+          </Grid>
+
+          </Paper>
+          </Grid>
+
+        <Grid item xs>            
+          <Paper className={classes.wait}>
+
+            <Typography variant="overline">Median wait</Typography>
+            <Typography variant="h3">{ route.wait < 0 ? "?" : route.wait.toFixed(0) } minutes</Typography>
```
youngj: Since the font size is quite large, maybe abbreviate "minutes" to "min" so it's less likely to wrap to another line?

exxonvaldez:
Screen Shot 2019-07-23 at 4 25 11 PM

Here's round 3.

frontend/src/helpers/routeCalculations.js
exxonvaldez: Done.

frontend/src/components/RouteSummary.jsx
exxonvaldez: Okay, it's out of 100, just for route scores for now.  I think Preston was trying to clean up the Info UI.

frontend/src/components/RouteSummary.jsx
exxonvaldez: Subscores have been replaced with a read-only 0-5 star display with half stars.  There is a stealth tooltip over the wait and speed stats.  Showing the full calculation of route score is bigger feature (the display would resemble the table shown in scoring a trip but cleaner).

frontend/src/components/RouteSummary.jsx
exxonvaldez: Did some refactoring so the wait scoring/ranking matches speeds/grades.

frontend/src/components/RouteSummary.jsx
exxonvaldez: By adding the 0-5 stars display, heights no longer line up.  It's not really the objective of this particular grid to stay on one row, especially if I add the 20 min wait probability and travel variability subscores.  Also don't have a good place to park length yet, will keep it in mind as this develops.

frontend/src/components/RouteSummary.jsx
exxonvaldez: Tried that, but wound up shrinking the font for "minutes" instead.

exxonvaldez:
I'm using a new package, material-ui/labs, so you have to run docker-compose build first.

frontend/src/components/RouteSummary.jsx
```diff
@@ -151,31 +103,38 @@ function RouteSummary(props) {
...
           <Paper className={classes.wait}>
 
             <Typography variant="overline">Median wait</Typography>
-            <Typography variant="h3">{ route.wait < 0 ? "?" : route.wait.toFixed(0) } minutes</Typography>
-            
-            { waitRanking ? `#${waitRanking} of ${routes.length} for shortest wait` : null }
-            { waitRanking ? <small>&nbsp;({ grades.medianWaitScore }/100)</small>
-                : " ranking not available"}
+            <br/>
+            <Tooltip title={ waitRanking ? 'Subscore: ' + grades.medianWaitScore + '/100' : null}>
```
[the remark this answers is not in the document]
exxonvaldez: Good point.  The Rating component has its own tooltip mechanism which only is for editable Ratings (for explaining what a hovered-over quantity of stars means).  Putting a Tooltip around the Rating did not work, presumably it doesn't support being wrapped like that.  So I moved the Tooltip around the card (Paper).

exxonvaldez:
> The current scoring system seems harsh in giving 0 stars for all routes with median wait times >= 10 minutes, even if that is what is scheduled. 

Happy to discuss this and get others' input.  I do think 20+ min scheduled headways impact usability of a line significantly, especially with combined with unevenly distributed vehicles or missing runs, and the long headways really puts those routes in a different class of service.

> When computing the score, could we also compare the median wait time with the scheduled wait time if the backend could provide the scheduled stats?

Hmm, my first thought would be some kind of subscore for aggregating on-time performance to replace the probability of a 20 min wait (which was sort of a placeholder representing gaps in service).  We should investigate what is going on when the median wait differs significantly from the scheduled wait -- if it is skipped runs, we could compile a skipped run percentage for each route.

Issue #155, opened 2019-07-11
Get rid of the outdated home/about/landing tabs


hathix:
We should just have dashboard/route now

![Screenshot 2019-07-10 at 6 15 53 PM](https://user-images.githubusercontent.com/728084/61014982-c5013000-a33e-11e9-874e-acff1a7ad061.png)

hathix:
We could just get rid of the weird gray tab bar entirely tbh

PR #179, opened 2019-07-18
Remove header nav

pb-nowa:
Removed old links from the top of /landing /home and /
#155 
Screen Shot 2019-07-17 at 6 51 26 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # OpenTransit's Metrics MVP
+
 [Check out our demo!](http://muni.opentransit.city/)
 
 ## Getting started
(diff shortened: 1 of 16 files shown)


hathix:
It looks like there are a lot of changes in here that aren't yours. Do you know where they came from? Ideally you'd only commit the things that are your own. So base it off the current master.

hathix:
OK, so i checked and most of the changes are just linter stuff. No functional changes. Let me see a demo then I'll approve.

hathix:
i think i hit merge twice and caused a bug. this has merged, it's fine

PR #188, opened 2019-07-18
Update gtfs scraper to update date_ranges.csv

jtanquil:
Previously, the GTFS scraper wouldn't update the list of possible date ranges for timetables when adding additional timetables, which would cause errors when fetching timetables for certain dates. Now the scraper will append additional date ranges to `date_ranges.csv` when it runs.

### DIFF
--- a/models/gtfs.py
+++ b/models/gtfs.py
@@ -168,17 +168,17 @@ class GtfsScraper:
         print(f"{datetime.now().time().isoformat()}: Uploaded {s3_path} to s3 bucket.")
 
     def save_date_ranges(self, s3 = False, ver = "v1"):
-        df = self.get_date_ranges()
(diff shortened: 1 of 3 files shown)


models/gtfs.py
```diff
@@ -168,17 +168,17 @@ def upload_to_s3(self, s3_path: str, df: pd.DataFrame):
         print(f"{datetime.now().time().isoformat()}: Uploaded {s3_path} to s3 bucket.")
 
     def save_date_ranges(self, s3 = False, ver = "v1"):
```
hathix: This confused me just a bit. Can you clarify that this stores on S3 AND makes a local cached copy?

hathix: Like, looking through the function I was confused what it did. It made sense once you explained that you upload to S3 and cache locally. Just add comments to the function saying that.

PR #203, opened 2019-07-25
Adding Routing for Route Page

ckrajewski:


frontend/src/components/RouteTable.jsx
```diff
@@ -288,7 +288,6 @@ function RouteTable(props) {
                   return (
                     <TableRow
                       hover
-                      onClick={event => handleClick(event, row)}
```
hathix: why'd we get rid of this?

frontend/src/components/RouteTable.jsx
```diff
@@ -303,14 +302,15 @@ function RouteTable(props) {
                       >
                         <Link
                           to={{
-                            type: 'RECEIVED_GRAPH_PARAMS',
+                            type: 'ROUTESCREEN',
```
hathix: this is because the RECEIVED_GRAPH_PARAMS tag is a misnomer, right? let's just make sure that that tag is no longer used in the app.

frontend/src/components/RouteTable.jsx
```diff
@@ -303,14 +302,15 @@ function RouteTable(props) {
...
                             payload: {
                               route_id: row.id,
                               direction_id: null,
                               start_stop_id: null,
                               end_stop_id: null,
-                            },
-                            query: { route_id: row.id },
+                              path:'/test'
```
hathix: why this?

ckrajewski:
Holding off on merge until I figure out:

- how to structure the URL in a rest format (instead of just variables)
- firing route page logic when supplied with variables

frontend/src/components/RouteTable.jsx
ckrajewski: deleted sorry

frontend/src/components/RouteTable.jsx
ckrajewski: It was just a typo
ROUTESCREEN is correct; it refers to our page in the routesMap js file
RECEIVED_GRAPH_PARAMS is an action type that is still valid (can be dispatched to)

frontend/src/components/RouteTable.jsx
ckrajewski: We don't need this.
This onClick was before we had routes working
The onClick event would bubble up, triggering logic to function like redux first router right @exxonvaldez ?

PR #207, opened 2019-07-25
Extract route geometry from GTFS files

youngj:
This PR makes it possible for the frontend to draw the actual route traveled between two stops (instead of drawing straight lines between each stop), and to get the distance along a route between two stops, which can be used to for more accurate speed calculations.

The command line script save_routes.py downloads and parses the GTFS specification (hardcoded for Muni for now), matches up GTFS shapes with Nextbus directions and stops, and saves the configuration for all routes to S3 (https://opentransit-precomputed-stats.s3.amazonaws.com/routes_v1_sf-muni.json.gz). The S3 object contains data merged from Nextbus and GTFS, extending the JSON data structure previously returned by /api/routes. The frontend can then request this S3 URL instead of /api/routes .

For each direction, the JSON object contains a coords array defining the shape of the route, where the values are objects containing lat/lon properties:
```
"coords":[
  {"lat":37.80707,"lon":-122.41727}
  {"lat":37.80727,"lon":-122.41562},
  {"lat":37.80748,"lon":-122.41398},
  {"lat":37.80768,"lon":-122.41234},
  ...
]
```

For each direction, the JSON object also contains a stop_geometry object where the keys are stop IDs and the values are objects with a distance property (cumulative distance in meters to that stop along the GTFS shape), and an after_index property (index into the coords array of the last coordinate *before* that stop). 
```
"stop_geometry":{
    "5184":{"distance":8,"after_index":0},
    "3092":{"distance":279,"after_index":1},
    "3095":{"distance":573,"after_index":3},
    "4502":{"distance":1045,"after_index":8},
    ...
}
```

Isochrone.jsx shows how to draw the route between any two stops as a Polyline: draw first stop, draw coords from index firstStop.after_index + 1 to lastStop.after_index, then draw the last stop.

In order to match a Nextbus direction with a GTFS shape_id, save_routes.py finds the GTFS shape_id for that route where distance(first coordinate of shape, first stop location) + distance(last coordinate of shape, last stop location) is a minimum. 

In order to determine where a Nextbus stop appears along a GTFS shape, save_routes.py finds the closest coordinate of the GTFS shape to the Nextbus stop, then finds whether the stop is closer to the line between that GTFS shape coordinate and the previous GTFS coordinate, or the line between that GTFS shape coordinate and the next GTFS coordinate. (This may not always be correct for shapes that loop back on themselves.)

To calculate distances, save_routes.py projects the lat/lon coordinates of the GTFS shape and Nextbus stops (in degrees) to x/y coordinates (in meters). Since the number of meters per degree longitude doesn't vary much over a city like SF, this projection approximates the number of meters per degree longitude as a constant based on a chosen latitude in the city.

This script stores the route configuration in the S3 bucket opentransit-precomputed-stats . We could create a new bucket for this. However could also consider whether it would be better to only use one S3 bucket for all data since it would make it easier for other people to set up their own S3 backend with a different bucket.

Currently the script just overwrites the one S3 path, but this process could be extended in the future to store different paths for different dates, to allow fetching historical data for route configurations.

### DIFF
--- a/frontend/src/actions/index.js
+++ b/frontend/src/actions/index.js
@@ -65,7 +65,7 @@ export function resetIntervalData() {
 export function fetchRoutes() {
   return function(dispatch) {
     axios
-      .get('https://opentransit-precomputed-stats.s3.amazonaws.com/routes_v1_sf-muni.json.gz?v3')
+      .get('https://opentransit-precomputed-stats.s3.amazonaws.com/routes_v2_sf-muni.json.gz')
       .then(response => {
         dispatch({ type: 'RECEIVED_ROUTES', payload: response.data.routes });
       })
--- a/frontend/src/actions/index.js
+++ b/frontend/src/actions/index.js
@@ -65,11 +65,9 @@ export function resetIntervalData() {
 export function fetchRoutes() {
   return function(dispatch) {
     axios
-      .get('/api/routes', {
-        baseURL: metricsBaseURL,
-      })
+      .get('https://opentransit-precomputed-stats.s3.amazonaws.com/routes_v1_sf-muni.json.gz?v3')
       .then(response => {
-        dispatch({ type: 'RECEIVED_ROUTES', payload: response.data });
+        dispatch({ type: 'RECEIVED_ROUTES', payload: response.data.routes });
       })
       .catch(err => {
         dispatch({ type: 'RECEIVED_ROUTES_ERROR', payload: err });
--- a/frontend/src/actions/index.js
+++ b/frontend/src/actions/index.js
@@ -2,6 +2,8 @@ import axios from 'axios';
 import { metricsBaseURL } from '../config';
 import { getTimePath, getStatPath } from '../helpers/precomputed.js';
 
+export const routesUrl = 'https://opentransit-precomputed-stats.s3.amazonaws.com/routes_v2_sf-muni.json.gz';
+
 export function fetchGraphData(params) {
   return function(dispatch) {
     axios
@@ -65,7 +67,7 @@ export function resetIntervalData() {
 export function fetchRoutes() {
   return function(dispatch) {
     axios
-      .get('https://opentransit-precomputed-stats.s3.amazonaws.com/routes_v2_sf-muni.json.gz')
+      .get(routesUrl)
       .then(response => {
         dispatch({ type: 'RECEIVED_ROUTES', payload: response.data.routes });
       })
--- a/frontend/src/screens/Isochrone.jsx
+++ b/frontend/src/screens/Isochrone.jsx
@@ -1,6 +1,6 @@
 import React, { Fragment } from 'react';
 import { connect } from 'react-redux';
-import { Map, TileLayer } from 'react-leaflet';
+import { Map, TileLayer, Polyline } from 'react-leaflet';
 import L from 'leaflet';
 import Control from 'react-leaflet-control';
 import * as turf from '@turf/turf';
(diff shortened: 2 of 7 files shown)


hathix:
As we just mentioned IRL, would be helpful to put the text from this PR in the tops of the affected files. The text provides awesome context, so let's make sure it's easily findable.

frontend/src/actions/index.js
```diff
@@ -65,11 +65,9 @@ export function resetIntervalData() {
 export function fetchRoutes() {
   return function(dispatch) {
     axios
-      .get('/api/routes', {
-        baseURL: metricsBaseURL,
-      })
+      .get('https://opentransit-precomputed-stats.s3.amazonaws.com/routes_v1_sf-muni.json.gz?v3')
```
hathix: Do you want to put the versions of the API (v1, v3) in the constants file? Easier to update that way

frontend/src/screens/Isochrone.jsx
```diff
@@ -258,26 +258,46 @@ class Isochrone extends React.Component {
...
+                                        for (let i = fromStopIndex; i <= toStopIndex; i++)
+                                        {
+                                            let stopInfo = routeInfo.stops[dirInfo.stops[i]];
+                                            tripPoints.push(stopInfo);
+                                        }
                                     }
+                                }
 
-                                    let polyLine = L.polyline(stopPoints).addTo(map);
+                                if (tripPoints.length)
```
hathix: Would help to have a comment describing what you're doing here and what a `polyline` is, for those new to the code

frontend/src/screens/Isochrone.jsx
```diff
@@ -258,26 +258,46 @@ class Isochrone extends React.Component {
...
+                                    let polyLine = L.polyline(tripPoints).addTo(map);
                                     polyLine.bindTooltip(routeInfo.id, {direction:'center', opacity:0.9, permanent:true});
 
                                     this.tripLayers.push(polyLine);
-
-                                    let fromStopInfo = routeInfo.stops[tripItem.fromStop];
-                                    let toStopInfo = routeInfo.stops[tripItem.toStop];
-
                                     this.tripLayers.push(L.circle(fromStopInfo, 40, {color:'#090', fillOpacity:0.8, stroke:false}).addTo(map));
                                     this.tripLayers.push(L.circle(toStopInfo, 40, {color:'#900', fillOpacity: 0.8, stroke:false}).addTo(map));
```
hathix: Note that we'll probably want to factor out the `color`s into a common CSS/JS file. this is something we should do across the codebase, so not necessary to address here, but if we have time let's start doing it.

save_routes.py
```diff
@@ -0,0 +1,211 @@
...
+
+routes_data = []
+
+deg_lat_dist = util.haver_distance(center_lat, center_lon, center_lat-0.1, center_lon)*10
+deg_lon_dist = util.haver_distance(center_lat, center_lon, center_lat, center_lon-0.1)*10
+
+def project_xy(lon, lat, z=None):
+    return (round((lon - center_lon) * deg_lon_dist, 1), round((lat - center_lat) * deg_lat_dist, 1))
+
+def distance(x1, x2, y1, y2):
```
hathix: is there a reason why we use haversine distance in some places and euclidean in others?

save_routes.py
```diff
@@ -0,0 +1,211 @@
...
+        terminal_dist_order = np.argsort(terminal_dists)
+
+        best_shape_index = terminal_dist_order[0]
+
+        shape_id = route_shape_ids[best_shape_index]
+        used_shape_ids[shape_id] = True
+
+        print(f'  {direction_id} = {shape_id} (n={route_shape_id_counts[best_shape_index]})')
+
+        geometry = geometries[best_shape_index]
```
hathix: clarify in a comment what the best shape index is?

save_routes.py
```diff
@@ -0,0 +1,211 @@
...
+        geom_length = len(xy_geometry.coords)
+        xy_array = np.array(xy_geometry)
+        x_array = xy_array.T[0]
+        y_array = xy_array.T[1]
+
+        prev_x_array = np.r_[x_array[0], x_array[:-1]]
+        prev_y_array = np.r_[y_array[0], y_array[:-1]]
+        geom_dist = np.cumsum(distance(x_array, prev_x_array, y_array, prev_y_array))
+
+        for stop_id in dir_stop_ids:
```
hathix: this took a bit of work to parse, perhaps comment this part of the algorithm in particular

save_routes.py
```diff
@@ -0,0 +1,211 @@
...
+        for i, shape_id in enumerate(route_shape_ids):
+            if shape_id not in used_shape_ids:
+                print(f'  ? = {shape_id} (n={route_shape_id_counts[i]})')
+
+data_str = json.dumps({
+    'version': 'v1',
+    'routes': routes_data
+}, separators=(',', ':'))
+
+cache_path = f'{util.get_data_dir()}/routes_v1_{agency_id}.json'
```
hathix: put the v1 in a constant (with the v1 just above) so it's more robust?

youngj: Good point, fixed

save_routes.py
youngj: Sure, split into function and added some comments

save_routes.py
youngj: The euclidean distance function wasn't actually necessary, so I replaced it with haversine distance. The euclidean coordinates were only needed for shapely to be able to calculate the distance in meters between a point (the stop) and a line segment (between two coordinates of the GTFS shape).

frontend/src/screens/Isochrone.jsx
youngj: Maybe later

frontend/src/screens/Isochrone.jsx
youngj: Added a comment. Several of the Leaflet components use Polylines to draw a series of line segments, seems fairly self-explanatory

frontend/src/actions/index.js
youngj: Updated to v2 and removed query string. Not sure which constants file you're referring to.

save_routes.py
```diff
@@ -0,0 +1,317 @@
...
+            for shape_id in route_shape_ids:
+                geometry = shapes[shapes['shape_id'] == shape_id]['geometry'].values[0]
+                geometries.append(geometry)
+
+                shape_start = geometry.coords[0]
+                shape_end = geometry.coords[-1]
+
+                start_dist = util.haver_distance(first_stop_info.lat, first_stop_info.lon, shape_start[1], shape_start[0])
+                end_dist = util.haver_distance(last_stop_info.lat, last_stop_info.lon, shape_end[1], shape_end[0])
+
```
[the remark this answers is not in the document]
youngj: partridge, the GTFS parsing library we're using, seems to ignore the shape_dist_traveled column. Instead of returning 1 row for each point in the shape, it returns 1 row for the entire shape with the geometry as a shapely.geometry.LineString object (https://github.com/remix/partridge/blob/d913619c578a200c49d1628d49404a904cfaf557/partridge/geo.py). Consequently it seems easier to recalculate the cumulative distance ourselves from the points in the geometry.

frontend/src/actions/index.js
```diff
@@ -65,11 +65,9 @@ export function resetIntervalData() {
 export function fetchRoutes() {
   return function(dispatch) {
     axios
-      .get('/api/routes', {
-        baseURL: metricsBaseURL,
-      })
+      .get('https://opentransit-precomputed-stats.s3.amazonaws.com/routes_v2_sf-muni.json.gz')
```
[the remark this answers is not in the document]
youngj: Good point, added a constant in actions.js. Since isochrone-worker.js can't access the React code directly, I pass the URL as a query parameter. This URL might not be a constant in the future if there are different URLs to get the route config on different dates.

save_routes.py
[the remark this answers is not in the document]
EddyIonescu: I see, I guess the XY suffices. As long as the discrepancy isn't too large (I'm thinking of all the steep hills), but should be really minor anyways and I have no idea as to how the `shape_dist_traveled` field is actually calculated.

PR #208, opened 2019-07-25
Make a reusable component for the hamburger menu

### LINKED ISSUE #206: Make a reusable component for the hamburger menu
#205 does a good job of adding the isochrone link to the menu. But the hamburger menu code is copy-pasted throughout the app! We should make that its own React component and reuse it across each page. Let's call it `<Sidebar>`.

alison-codes:
Fixes #206

### DIFF
--- a/frontend/src/components/SidebarButton.jsx
+++ b/frontend/src/components/SidebarButton.jsx
@@ -1,0 +1,59 @@
+import React from 'react';
+import Drawer from '@material-ui/core/Drawer';
+import IconButton from '@material-ui/core/IconButton';
+import List from '@material-ui/core/List';
+import ListItem from '@material-ui/core/ListItem';
+import ChevronLeftIcon from '@material-ui/icons/ChevronLeft';
+import MenuIcon from '@material-ui/icons/Menu';
+
+
+class SidebarButton extends React.Component {
+    constructor(props) {
+        super(props);
+
+        this.state = {
+            drawerOpen: false,
+        };
+
+        this.handleToggleDrawer = this.handleToggleDrawer.bind(this);
+    }
+
+    handleToggleDrawer() {
+        this.setState({ drawerOpen: !this.state.drawerOpen });
+    }
+
+    render() {
+        return <div>
+              <IconButton
+              color="inherit"
+              aria-label="Open drawer"
+              onClick={this.handleToggleDrawer}
+              edge="start"
+            >
+              <MenuIcon />
+            </IconButton>
+            <Drawer
+                variant="persistent"
+                anchor="left"
+                open={this.state.drawerOpen}
+            >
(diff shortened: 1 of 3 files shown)


frontend/src/components/SidebarButton.jsx
```diff
@@ -0,0 +1,59 @@
...
+        return <div>
+              <IconButton
+              color="inherit"
+              aria-label="Open drawer"
+              onClick={this.handleToggleDrawer}
+              edge="start"
+            >
+              <MenuIcon />
+            </IconButton>
+            <Drawer
```
hathix: Hm, can you put the Drawer in its own component? It feels weird if the *entire* drawer lives within the button component. It would be better if the button just instantiated a new `<SideBar></SideBar>` component instead of putting all the code for it in here.

exxonvaldez: > Hm, can you put the Drawer in its own component? It feels weird if the _entire_ drawer lives within the button component. It would be better if the button just instantiated a new `<SideBar></SideBar>` component instead of putting all the code for it in here.

I talked with Neel, and since the drawer shares state with the button, it's probably simpler to keep them together for now.

alison-codes: Leaving it simpler, yet functional, for now was my thought too. Though, I completely agree having the drawer live within a component named SidebarButton is confusing. I can rework next time.

Thithi32: Actually the idea was to keep the Drawer state inside the component so we wouldn't need to handle it from an outside button. But I agree that it feels weird to have it inside the header bar. It would have just took us longer :p

frontend/src/components/SidebarButton.jsx
```diff
@@ -0,0 +1,59 @@
...
+                    <IconButton
+                        color="inherit"
+                        aria-label="Open drawer"
+                        onClick={this.handleToggleDrawer}
+                        edge="start"
+                    >
+                        <ChevronLeftIcon />
+                    </IconButton>
+                    <List>
+                        <ListItem><a href="/">Dashboard</a></ListItem>
```
exxonvaldez: I just noticed that \<a href\> doesn't preserve Redux state, which we need for the Dashboard.  The idea is that date and time picker (which is only a time picker at the moment while the date picker is being rebuilt) will be on the Dashboard and RouteScreen, and maintain its settings persistently as the user navigates around.  The type of link we want to use is a Navlink, you can find examples of this is in the currently unused Navigation.jsx component, which was a navbar that was just removed from the UI.

hathix: Yes, you'll want to use components from https://github.com/trynmaps/metrics-mvp/blob/master/frontend/src/components/Navigation.jsx

PR #211, opened 2019-08-01
Adding Routing for Route Page

ckrajewski:


frontend/src/components/RouteTable.jsx
```diff
@@ -290,14 +289,14 @@ function RouteTable(props) {
                       >
                         <Link
                           to={{
-                            type: 'RECEIVED_GRAPH_PARAMS',
+                            type: 'ROUTESCREEN',
```
ckrajewski: ROUTESCREEN is for the RoutesMap page
It tells redux first router to navigate to our routes page (shows routes and route summary info)

frontend/src/routesMap.js
```diff
@@ -1,9 +1,24 @@
...
-  ROUTESCREEN: '/route',
-  RECEIVED_GRAPH_PARAMS: '/route',
+  ROUTESCREEN: {
+    path: '/route/:route_id/(direction)*/:direction_id?/(start_stop)*/:start_stop_id?/(end_stop)*/:end_stop_id?',
+    thunk: async (dispatch, getState) => {
+      const {location} = getState();
+      const {route_id, direction_id,start_stop_id, end_stop_id } = location.payload;
+
+
+      dispatch(handleGraphParams({
```
ckrajewski: this fires the same action that ControlPanel does to handle the rendering of route info

frontend/src/components/RouteTable.jsx
```diff
@@ -275,7 +275,6 @@ function RouteTable(props) {
                   return (
                     <TableRow
                       hover
-                      onClick={event => handleClick(event, row)}
```
exxonvaldez: Let's remove the handleClick method -- it came from the sample table code, and is supposed to be about selecting rows, which isn't something we actually supporting.

frontend/src/routesMap.js
```diff
@@ -1,9 +1,24 @@
+import { redirect, NOT_FOUND } from 'redux-first-router'
```
exxonvaldez: Are these imports being used?

frontend/src/routesMap.js
```diff
@@ -1,9 +1,24 @@
...
 export default {
   HOME: '/home',
   ABOUT: '/about',
   LANDING: '/landing',
   ISOCHRONE: '/isochrone',
   DASHBOARD: '/',
-  ROUTESCREEN: '/route',
-  RECEIVED_GRAPH_PARAMS: '/route',
+  ROUTESCREEN: {
+    path: '/route/:route_id/(direction)*/:direction_id?/(start_stop)*/:start_stop_id?/(end_stop)*/:end_stop_id?',
```
hathix: this URL is pretty long and confusing. could you put in a comment showing some of the possible URLs that are supported? or ideally, you could have some other function that computes the URL given, say, a dictionary of potential keys and their names.

frontend/src/components/RouteTable.jsx
hathix: this PR doesn't call the method anymore... but does the method exist in our code somewhere?

frontend/src/routesMap.js
ckrajewski: I can add comments
The ()* marking is a wildcard
It means that if given a parameter, it will parse it out
Otherwise, it ignores it

frontend/src/components/RouteTable.jsx
ckrajewski: We don't need this.
This onClick was before we had routes working
The onClick event would bubble up, triggering logic to function like redux first router right @exxonvaldez ?
I'll remove the handleClick function

frontend/src/routesMap.js
ckrajewski: No I'll remove them

frontend/src/components/RouteTable.jsx
```diff
@@ -184,35 +184,7 @@ function RouteTable(props) {
     setOrder(isDesc ? 'asc' : 'desc');
     setOrderBy(property);
   }
-
-  function handleClick(event, route) {
```
ckrajewski: removed onClick function

frontend/src/routesMap.js
ckrajewski: @hathix 
added comments

PR #220, opened 2019-08-01
Data diagnostic page

exxonvaldez:
Screen Shot 2019-07-31 at 10 54 50 PM

[...]

Screen Shot 2019-07-31 at 10 55 31 PM

I call this a "data diagnostic" screen.  It helps developers find routes with anomalous route speeds or route waits.  It has a speed vs wait chart at the top, followed by travel time across stops for all routes in each direction.  Once we have a date and time picker available on all screens, this screen will show data for different dates/times -- currently, it's using the initial date/time in graphParams.  

From the screen shots, you can see that the 57 has a different number of stops with precomputed trip times in each direction, and a large dip in the inbound direction.  We may need a different way of computing overall speed for a route other than end to end time (with first/last stop ignored for certain routes).


hathix:
This is really valuable! Few questions:

* Does the scatter plot show both inbound and outbound routes? Are they separate or merged? If they're merged, it might help to separate them out to keep it consistent with the line graphs, which separate inbound and outbound. 
* Can we add a small tag to indicate if a line is inbound or outbound? We know that "__I_F00" means inbound, but users might not. A translation would help.
* Is there a way to sort routes by longest travel time? That might be a common use case to figure out if there are any anomalously long or short routes. If this is too much work to implement now, we can just raise an issue and do it later.
* I see that the axes of the scatter plot don't actually cross at (0,0) -- and they shouldn't, given how all numbers here are positive. Could you make the axes positioned so that they cross at (0,0)? In other words, can you move the axes to the far edges of the scatter plot so that the user doesn't  mistakenly think they're looking at negative numbers?
* Is there a way for users to export this data? Users may want to export this to Google Sheets and do more advanced analysis there.
* Is there a way for users to dig deeper into the numbers behind a graph -- for instance, is there some more in-depth information they could download if they clicked on the graph? In your example, it's easy to tell that something's wrong with our numbers for certain routes, but we'd have to dig into our database to figure out exactly what's wrong.

exxonvaldez:
Keep in mind this page is intended for OpenTransit developers, to help us look for bugs and refine our algorithms.  We might not link it in our navigation if we don't want to.  The problem is that we currently report buses averaging 50mph end to end and multihour waits for routes that we know don't run midday.

>     * Does the scatter plot show both inbound and outbound routes? Are they separate or merged? If they're merged, it might help to separate them out to keep it consistent with the line graphs, which separate inbound and outbound.

It's merged.  Inbound and outbound wait times shouldn't differ enough to matter, and travel times in each direction are pretty close, I just checked.  Plus there's the 38 with four directions.  The quadrant chart is pretty dense as it is, and it's meant to help developers know which routes they should prioritize.

For your other comments, I don't think they apply to OpenTransit developers.  If you want to "productize" the page, we should brainstorm a bit about what problem it solves. 

We can have a feature that makes it easier to download the existing precomputed wait times and trip times files.  Or we get fancier about it like Uber Movement, where you are inputting filters and getting custom data dumps back.

exxonvaldez:
Added a semi-hidden link in the navigation to the data diagnostic page (a dot at the end), and made DataDiagnostic a functional component.

I can start another PR for making the x-axis correspond to distance along a route.  Should be interesting to see how that looks.

PR #221, opened 2019-08-01
Fix links in the spider map by using routing.

exxonvaldez:
There were click handlers in `MapSpider` that were calling handleGraphParams followed by push('/route').  After Chris' merge, these were returning a 404 page.  Looks like the quickest fix for now is to push a fully parameterized URI.  There may be a better way (like getting the redux state change and the routing change done just by dispatching an action), but this works for now.


hathix:
This LGTM, but @ckrajewski should sign off on it in case there's some nuance I'm missing.

exxonvaldez:
@hathix I'd like to merge this is now.  The 404 pages look bad and interfere with my testing.  It's a small change to go back to the old broken code if needed, and this is easy to modify later if there's a better way going forward.

PR #222, opened 2019-08-02
Add date/time picker to app bar

### LINKED ISSUE #213: Add Material date picker
#178 added a material timepicker. Let's add a material datepicker too; we don't currently have one that works. Looks at Uber Movement's UI and steal that, ideally.

### LINKED ISSUE #214: Make the Material pickers more compact
In #178, we add material ui pickers for direction/route/stops. It's not very compact though. We should move these pickers to all live in one or two lines instead of 4 separate lines.

### LINKED ISSUE #215: Add a floating date/time picker to the corner of the UI
As you can see from our mockup, the date/timer picker should be a consistent element always floating in the corner of the UI. This is because every view could benefit from a date/time picker, and it's good to keep it in a consistent place.

2019_06_05 9_22 PM Office Lens (1)

exxonvaldez:
Screen Shot 2019-08-01 at 7 30 55 PM

Screen Shot 2019-08-01 at 7 31 19 PM

This PR adds a date and time picker that appear via clickable text (a button) to the app bar for the `Dashboard` and `RouteScreen` screens.  I've also tweaked the layout of the remaining `Select`s in `ControlPanel` so that route and direction appear on one line if there's room.  Same for from stop and to stop.

Note that the date picker doesn't do any validation to see if we actually have data for the given date, and the display of the current date/time is bare bones (it will say "All day" by default, but otherwise it is just the raw start and end times).  The date and time picker appear in a `Popper`, which gives us room to add Uber Movement-style date controls when we can support it.

- index.js: fetch precomputed data if we have a date
- ControlPanel: move time picker out.  Place remaining Selects in two Grids so that we get a 2 x 2 layout that can shrink to one column as needed.  Use second stop list.  Fix htmlFor attributes to point at actual inputs.
-DateTimePanel:   new button that opens a Popper with a date picker and time picker.
- RouteSummary: fix warnings for null Tooltip title.
- RouteTable: fix warnings, removed unused code for selection counts.
- page.js:  remove routing for dispatch of RECEIVED_GRAPH_PARAMS, as the time/date pickers need to set graphParams without changing the current screen.  Screen changes are accomplished through Links or by pushing a new URI.
- Dashboard:  convert to functional component, add DateTimePanel.
- RouteScreen:  convert to functional component, add DateTimePanel.

Fixes #213; fixes #214; fixes #215 .


exxonvaldez:
Fixes #213 and #214.

hathix:
Awesome! A few functionality-related questions and comments:

* Will the Popper support the relatively complex UI elements needed for the Uber Movement-style UI, such as selectors for the day of the week? And will the gray bar be able to show extra details like the chosen days of the week?
* Is there a way to clear the date part and look at data for *all* dates? If not, it would probably help to add a "clear" or "reset" button next to the "close" button on the popper. It's probably good UX to add a clear button anyway, since most filtering UIs offer a clear button, and it'll get more and more important as the Popper adds more and more functionality.
* What would happen if the window got narrower? Would the route name get cut off, or would the gray bar be hidden? How should we address mobile, where you definitely can't show both the route name and the gray bar?
* The time display is a bit inconsistent: in the dropdown it says "7 am - 10 am" but in the gray bar it says "07:00-10:00", which is military time. This might get confusing if the user chooses a time in the afternoon, when military and normal times would look very different. Can we use the same time formatting in both places?

exxonvaldez:
Screen Shot 2019-08-02 at 6 58 28 PM

Screen Shot 2019-08-02 at 6 59 44 PM

>     * Will the Popper support the relatively complex UI elements needed for the Uber Movement-style UI, such as selectors for the day of the week? And will the gray bar be able to show extra details like the chosen days of the week?

First screenshot shows the widgets needed, they will fit in the Popper (which is just a container) no problem.

Second screenshot shows Uber's gray bar, they use a really tiny font.  Time range is just words, no numbers.

>     * Is there a way to clear the date part and look at data for _all_ dates? If not, it would probably help to add a "clear" or "reset" button next to the "close" button on the popper. It's probably good UX to add a clear button anyway, since most filtering UIs offer a clear button, and it'll get more and more important as the Popper adds more and more functionality.

For now, the user must pick one day, and this PR enforces that by ignoring an empty date. I can add a "reset" in this PR.

>     * What would happen if the window got narrower? Would the route name get cut off, or would the gray bar be hidden? How should we address mobile, where you definitely can't show both the route name and the gray bar?

The blue bar can wrap so the breadcrumbs becomes multiple lines.  We need to decide if we want to keep the breadcrumbs, and be like nyc busstats.  The Select menus themselves can act as breadcrumbs and the inputs go horizontally along the top, wrapping as needed.  We can also be more like Uber Movement, and keep options in a lefthand drawer that sits over the map.  They punt on small displays, but basically you need lots of drawers for all the content.  We can also hide the gray bar as a small floating button that you tap to expand into the bar, and tap again for the Popper.

>     * The time display is a bit inconsistent: in the dropdown it says "7 am - 10 am" but in the gray bar it says "07:00-10:00", which is military time. This might get confusing if the user chooses a time in the afternoon, when military and normal times would look very different. Can we use the same time formatting in both places?

Yeah, we just hadn't gotten around to creating constants to represent the time ranges, I can do that in this PR.

youngj:
The top bar seems like a good place for this to make it easy to reuse across different screens. 

The date format in the top bar is different from the date format in the panel.

The time range dropdown is blank when it should say "All day".

It seems redundant for the user to click and open a panel that has the same information that they could already see, and then click again to actually change the date or time. Could we put the date / time controls directly in the top bar so the user could change them in one click? 

The text "Date-Time Range" and the following padding uses a lot of space. If the above suggestion is applied, the Date and Time Range controls could be labeled individually (labels could be above the controls since the top bar seems tall enough) or perhaps not at all.

exxonvaldez:
> The date format in the top bar is different from the date format in the panel.

Fixed.
> The time range dropdown is blank when it should say "All day".

Fixed.

> It seems redundant for the user to click and open a panel that has the same information that they could already see, and then click again to actually change the date or time. Could we put the date / time controls directly in the top bar so the user could change them in one click?

That's a good point.  We expect the panel to have the same options as Uber Movement (see screenshot above), so eventually the panel will get a lot bigger.

> The text "Date-Time Range" and the following padding uses a lot of space. If the above suggestion is applied, the Date and Time Range controls could be labeled individually (labels could be above the controls since the top bar seems tall enough) or perhaps not at all.

The padding is there to show where the second date of the date range will go once we support date ranges.  As discussed with Neel above, this may be too wide for some screens.  So the gray bar may have to support a compact form (like just an icon only) or float elsewhere, like bottom right corner.  In the long run, we need to decide if we want input controls laid out horizontally along the top, or vertically on the left, or stashed in the corners like they are now.

exxonvaldez:
Cleaned up two things:  made the DateTimePanel's TimeRange select wider.  Before it was changing width as you chose things, pushing the popper's right edge off screen.

And removed ControlPanel.css, this contained some Bootstrap-specific css plus the fix for the map that was showing through the Bootstrap selects.  Don't need either of those any more.

exxonvaldez:
Changed to slashes, removed fixed width from button and added nowrap to part of the button text (date, time range, and icon).

I'm going to keep the label for now for consistency with the labeled `Select`s on the route screen.  I think it'll help us visualize the final layout (inputs on the left vs inputs at the top vs corners).

PR #224, opened 2019-08-03
fix loading URLs with slashes in production environment

youngj:
Currently URLs like http://muni.opentransit.city/route/10 with a slash in the path are broken because create-react-app replaces %PUBLIC_URL% with `frontend/build`, not `/frontend/build` . This PR adds a `base` tag to the HTML so that all relative URLs are loaded relative to the domain root instead of the current "directory" of the URL path.


hathix:
Nice! Ship it

PR #226, opened 2019-08-04
add the app bar to the isochrone map and use its date/time selector; add instructions to isochrone map

youngj:
Instructions for the isochrone map are shown in the top right when not showing trip information.

image

image

image

image

### DIFF
--- a/frontend/public/isochrone-worker.js
+++ b/frontend/public/isochrone-worker.js
@@ -129,7 +129,7 @@ async function getTripTimesFromStop(routeId, directionId, startStopId, dateStr, 
 
         let s3Url = 'https://opentransit-precomputed-stats.s3.amazonaws.com/trip-times/v1/sf-muni/'+
             dateStr.replace(/\-/g, '/')+
-            '/trip-times_v1_sf-muni_'+dateStr+'_'+statPath+timePath+'.json.gz?v2';
+            '/trip-times_v1_sf-muni_'+dateStr+'_'+statPath+timePath+'.json.gz';
 
         tripTimes = tripTimesCache[dateStr + timeStr + stat] = await loadJson(s3Url).catch(function(e) {
             sendError("error loading trip times: " + e);
@@ -203,7 +203,7 @@ async function getWaitTimeAtStop(routeId, directionId, stopId, dateStr, timeStr,
 
         let s3Url = 'https://opentransit-precomputed-stats.s3.amazonaws.com/wait-times/v1/sf-muni/'+
             dateStr.replace(/\-/g, '/')+
-            '/wait-times_v1_sf-muni_'+dateStr+'_'+statPath+timePath+'.json.gz?v2';
+            '/wait-times_v1_sf-muni_'+dateStr+'_'+statPath+timePath+'.json.gz';
 
         //console.log(s3Url);
 
--- a/frontend/src/App.css
+++ b/frontend/src/App.css
@@ -36,3 +36,15 @@
 .largeMarginTop {
   margin-top:4rem;
 }
+
+.page-title
+{
+    flex-grow:1;
+}
+
+.flex-screen
+{
+    height:100%;
+    display:flex;
+    flex-direction:column;
+}
--- a/frontend/src/screens/Dashboard.jsx
+++ b/frontend/src/screens/Dashboard.jsx
@@ -19,33 +19,24 @@ import {
   resetIntervalData,
 } from '../actions';
 
-const useStyles = makeStyles({
-  title: {
-    flexGrow: 1,
-  },
-});
-
 function Dashboard(props) {
-  
+
   useEffect(() => {
     if (!props.routes) {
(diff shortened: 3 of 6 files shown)


exxonvaldez:
Looks nice.  I see you removed the useStyles hook in favor of css, but wouldn’t it be more consistent with our decisions in the readme.md to make Isochrone a functional component?

youngj:
I changed setting the style of the page title from useStyles to the page-title CSS class to reduce code duplication. Each component should not need to separately specify the same style rules for the page title. After removing this style from those components, there were no longer any styles that needed to be set via useStyles so I removed it, but it could be added again in the future if there are any styles that need to be set this way.

Making Isochrone a functional component does not seem like it would be a productive use of time. Functional components seem convenient for simple components, but I don't see any reason to force everything to use functional components, let alone spend time rewriting working code, when component classes work perfectly well.

exxonvaldez:
> I changed setting the style of the page title from useStyles to the page-title CSS class to reduce code duplication. 

Gotcha.  I was thinking we'd refactor the common UI elements into its own navbar component, and then you can extract the style stuff into that navbar.  

> Making Isochrone a functional component does not seem like it would be a productive use of time. Functional components seem convenient for simple components, but I don't see any reason to force everything to use functional components, let alone spend time rewriting working code, when component classes work perfectly well.

I made a functional component version of Isochrone.jsx with hooks based on this branch, you can see it at https://github.com/trynmaps/metrics-mvp/compare/isochrone-app-bar...tc-isochrone-hooks?expand=1 .  Seems to work, and is actually I think less verbose, and easier to read.  It's not a particularly dramatic change, let me know what you think.  I'm not particularly attached to hooks myself, but it'll make working on different screens easier for developers.

youngj:
Personally I think the hooks version is less easy to read. 

Compare:
```
        this.state = {
            stat: 'median',
            maxTripMin: 90,
            computedMaxTripMin: null,
            computeId: null,
            computing: false,
            latLng: null,
            endLatLng: null,
            tripInfo: null,
            enabledRoutes: {}
        };
```
with:
```
        const [stat, setStat] = useState('median');
        const [maxTripMin, setMaxTripMin] = useState(90);
        const [computedMaxTripMin, setComputedMaxTripMin] = useState(null);
        const [computeId, setComputeId] = useState(null);
        const [computing, setComputing] = useState(false);
        const [latLng, setLatLng] = useState(null);
        const [endLatLng, setEndLatLng] = useState(null);
        const [tripInfo, setTripInfo] = useState(null);
        const [enabledRoutes, setEnabledRoutes] = useState({});
```

The class version uses idiomatic JavaScript that any JavaScript developer can understand without knowing React or what the function useState does. It also facilitates debugging because you could just go `console.log(this.state)` to see everything.

The hooks version requires separate boilerplate for each state variable. The function `useState` does not act like a normal JavaScript function at all. For example, the expression `useState(null)` creates state associated with that particular component instance despite the component instance not being an argument to useState. Also, each different call to `useState(null)` returns different state (compare `const [computedMaxTripMin, setComputedMaxTripMin] = useState(null);` vs `const [computeId, setComputeId] = useState(null);` vs ) even though useState has the same arguments. This is not how JavaScript functions normally behave. In order to understand what's going on with hooks, the reader needs to read the documentation for React Hooks and that `useState` and `useEffect` have special magic behavior.

Compare also the readability of `useEffect(fn,[])` vs componentDidMount.

Since React class components are just idiomatic JavaScript and lifecycle methods like componentDidMount, componentDidUpdate, and render are fairly self-explanatory, a reader who understands JavaScript doesn't really need to know React to understand what's going on.

Also, since class components have been around much longer than hooks, there are going to be a lot more developers who have used React before who are familiar with class components already.

One of the main motivations why the React developers introduced hooks was that some people were apparently confused by JavaScript's `this` variable, so they came up with a clever way to avoid users ever needing to see `this` by instead creating a bunch of closures. I don't have a problem understanding `this`, so this change doesn't help me. Instead, creating closures adds some cognitive load because inner functions may depend on variables in the outer function which look the same as local variables. When using classes, long-lived instance variables (`this.foo`) are easily distinguished from local variables (`foo`). 

In the hooks version of the Isochrone code, long-lived instance variables were moved to global variables, which are not really equivalent since there could potentially be more than one Isochrone component. Apparently the Hooks equivalent of instance variables is useRef (https://reactjs.org/docs/hooks-faq.html#is-there-something-like-instance-variables). React's example code for useRef seems more complex than using instance variables like `this.foo`.

exxonvaldez:
I admire the depth of your anti-Hooks conviction.  I'm good with the Isochrone component as a class.

We should have invited you to the discussion we had a few months back about whether to use functional components going forward.  There was a lot of wheel spinning because we all agreed to switch to Material UI, but we weren't sure at the time if that meant using Hooks was required (now we know the answer is no, except for Hooks-based stuff like useStyles).  We decided to use Hooks going forward but migrate the existing class-based components later.

Issue #164, opened 2019-07-11
Improve route sorting


hathix:
Pretty small issue but I'll flag it nonetheless.

Currently the routes are sorted like 1, 10, 11, ... 19, 1AX, 1BX, 2, 20, 21, etc. Ideally it would be sorted like 1, 1AX, 1BX, 2, 3, ... 9, 10, 11, ... 20, 21, etc. That is, we should sort by route number, not lexically.

![Screenshot 2019-07-10 at 8 38 57 PM](https://user-images.githubusercontent.com/728084/61020469-c2103a80-a352-11e9-982b-dcecbdceeb4e.png)

hathix:
This should be done in https://github.com/trynmaps/metrics-mvp/blob/master/frontend/src/components/RouteTable.jsx

hathix:
basically, grab the sorting function from this stackoverflow answer: https://stackoverflow.com/a/8107614.

then in the `desc` sort function, if the `orderBy` happens to be the Name (and is a string), use that stackoverflow function instead. the current sorting function works great for numerical things but not these strings (and this string thing doesn't work for numerical values).

exxonvaldez:
We could also order routes the same as they are in the route picker.  This order comes from the Nextbus API that lists all the routes.  If the `orderBy` is Name, just use routes array unsorted (and reversed if `order` is 'desc').

PR #228, opened 2019-08-07
Improve spider map and stops map by using GTFS geometry.

### LINKED ISSUE #167: When you click on the map, put a pin at your chosen location
Currently we show all the routes that run near the spot you chose, which is great. But it's confusing to new users what just happened. It would help to show a pin in the spot they chose to make it clear that we're showing routes that go near their location.

Screenshot 2019-07-10 at 8 49 02 PM

### LINKED ISSUE #196: Show your location on the map with a blue dot
Google Maps and other apps do this. It's currently not clear where you are or why we're showing you these routes. Showing your location with a blue dot would help.

exxonvaldez:
Screen Shot 2019-08-06 at 8 13 31 PM

Screen Shot 2019-08-06 at 8 14 55 PM

- MapSpider: use GTFS geometry.  Also put a draggable blue Marker at spider's origin location (Fixes #167).  Minor adjustment for map height.
- MapStops: use GTFS geometry for plotting and calculating speed between stops.
- mapGeometry: new helper file for getting GTFS geometry between stops (`getTripPoints`) and for getting GTFS-based distance between stops (`getDistanceInMiles`).
- Isochrone:  extracted helper method `getTripPoints` for GTFS geometry.

Note:   we're still using precomputed GTFS-based distances for route lengths.  Calculating all the route lengths on the fly gets a little messy when the stop geometry is missing, since the fallback would be haversine distance between stops.  Maybe we could add the full GTFS shape length to our `routes` object?


exxonvaldez:
> The route direction object from S3 should already have a `distance` property that is the length of the entire GTFS shape in meters.

Oh, right, didn't see that the first time, thanks for pointing it out.  Switched the `getAllDistances()` method to use that.

One small thing I wanted to mention about the geometry:  at very high magnification, you can see the route shape going down the centerline of  the streets and then veer onto the sidewalk to get to a stop.  It would be nice to have a `stop_geometry` coordinate that represents the closest point on the shape to the stop, so that we can plot routes that are always on the shape and go near stops instead of into them.

frontend/src/components/RouteSummary.jsx
```diff
@@ -52,12 +52,12 @@ function RouteSummary(props) {
 
     allWaits = getAllWaits(props);
     
-    const allDistances = getAllDistances();
+    const allDistances = getAllDistances(props);
```
youngj: Would be clearer to use `routes` as argument instead of `props` object that contains routes property.

frontend/src/components/RouteSummary.jsx
```diff
@@ -52,12 +52,12 @@ function RouteSummary(props) {
...
     allSpeeds = getAllSpeeds(props, allDistances);
```
youngj: To calculate speeds for a route, I think it would be better to take the average of (distance for direction / median trip time for direction), instead of (average distance for all directions) / (average of median trip times for all directions). 

getAllDistances may be unneeded in that case.

hathix:
I think this fixes #167 (putting the pin on the map) not #164 (putting a blue dot at the GPS location), right?

exxonvaldez:
> I think this fixes #167 (putting the pin at the origin location) not #196 (putting a blue dot at the GPS location), right?

Good point, I updated the description.

PR #238, opened 2019-08-09
script to compute new arrival times, wait time stats, and trip time stats

youngj:
compute_new.py checks S3 to see the last date that has been fully computed, and starts computing arrival times, wait time stats, and trip time stats on the next day and advances until it reaches the current day. It could be run in Kubernetes as a cron job (not in this PR) to automatically compute the necessary stats. It could also be run more than once a day in order to provide near-real-time stats for the current day.


youngj:
When computing stats for the current day, it currently re-downloads the current day's GPS data from trynapi every time, which can take several minutes and will increase costs for data transfer. In the future we could optimize it to cache GPS data locally and only download new GPS data from trynapi.

PR #241, opened 2019-08-15
add cron job to automatically compute statistics once an hour

youngj:
It normally takes about 10 minutes.

### DIFF
--- a/kubernetes/compute-new-cronjob.yaml
+++ b/kubernetes/compute-new-cronjob.yaml
@@ -1,0 +1,28 @@
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: compute-new
+spec:
+  schedule: "0 * * * *"
+  concurrencyPolicy: Forbid
+  jobTemplate:
+    spec:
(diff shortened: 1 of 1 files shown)


hathix:
How much in gcp credits will this cost us?

youngj:
It doesn't seem to have made a noticeable change in the GCP cost. 

@EddyIonescu - has there been an increase in AWS cost in the past week since the cron job has been running, compared to >3 weeks ago?

It currently starts on the hour, but it takes a few minutes to download all the GPS state from trynapi so the data will probably include a few minutes after the hour.

EddyIonescu:
@youngj Thanks for letting me know! Our AWS costs are about $15/month, but has already gotten to about $20 this month, so about $25 for the month. It's been running on credits for a while now, but this would make them run out a few months sooner.

The increased cost-segments, in descending order, are S3 POST, data-transfer, and S3 GET costs. Given that the API is running on GCP, would it make sense to store the Cron job results on Google Cloud instead? Also, what's our status on GCloud credits?

PR #243, opened 2019-08-15
Codeowners to auto suggest reviewers

akgupta89:
This will help with auto suggesting reviewers

It'd be better if we moved the python stuff into a `/backend/` folder so we can maintain this cleaner as a mono repo and make the file structure of this repo suggest code owners properly. Ideally better than maintaining a document that suggests code owners to tag


hathix:
LGTM, I made a note that we should move stuff into the backend folder

PR #173, opened 2019-07-11
[WIP] Renders map stop legend for inbound/outbound

juanmanual:
- [x] Add legend for inbound stops
- [x] Add legend for outbound stops
- [ ] Add legend for start stops
- [ ] Add legend for end stops

PR #248, opened 2019-08-21
First pass at Marey chart

exxonvaldez:
## Proposed changes

- New component added temporarily to `RouteSummary` screen that shows the arrival history for the selected day and route as a Marey chart.
- The arrival history comes from S3, currently the same json file consumed by the back end.
- Arrivals are reorganized by vehicle id and sorted by distance along route, using precomputed GTFS geometry distances to convert stop ids to distances.
- ~~There is an apparent bug with the hover functionality when multiple Line series are present, so if hover was enabled, it would show a Hint for the wrong series.  There may be some complicated ways to work around this.  Hover functionality would help greatly with interpretation of dots (could show stop title, exact time, etc.)~~ Edit: Worked around onNearestXY bug with onValueMouseOver (requiring putting the pointer directly over a point) to include a basic hover Hint with stop title and time.

## Screenshot

These screenshots don't show the full chart, so you can't see the x-axis at the bottom.  The x-axis is inbound distance in miles, so inbound vehicles move down and the right.  Outbound vehicles move down and to the left.  Lines are colored by vehicle ID, so you can usually see vehicles reversing after reaching a terminal -- there may be a slight jump in the x direction.

Screen shot below is the evening that Jesse missed hack night due to the 27.

Screen Shot 2019-08-20 at 5 47 19 PM

.
.
.

Screen Shot 2019-08-20 at 5 55 57 PM

## Link to demo, if any

Try it out here: https://ot-terence.herokuapp.com/  Choose a route from the dashboard route table.  Some routes are cleaner than others (55 is pretty good, KT and M are ugly -- maybe need to hide arrivals of vehicles not in service?).  Also this chart is heavy on the client CPU so it may take some time to render.


exxonvaldez:
PR description edited to reflect addition of a basic working mouse hover (`Hint`).

Screen Shot 2019-08-21 at 2 25 33 AM

hathix:
Could you add checkboxes to your pr's body showing the steps that are finished and those that are remaining?

exxonvaldez:
> Could you add checkboxes to your pr's body showing the steps that are finished and those that are remaining? #173 has an example of it

- Biggest open question is whether this chart should be its own screen due to lengthy loading time.
- Chart could load faster if arrival history data was optimized for a Marey chart.

Future enhancements (not for this PR):

- Show scheduled trips on the chart
- Options to show only inbound or outbound
- Options to show only part of a day or part of a route
- Multi-day Marey chart as a "heatmap" to show bunching and gaps.

hathix:
How could we optimize arrival history data for a Marey chart? Unless this is a simple optimization, my gut reaction is that it isn't worth optimizing this early in the process; we should create a proof of concept and only optimize if there's user demand.

As for having the chart on its own screen, could you show just a small slice of the chart on the trip summary page (hopefully a slice that's fast to load) and only show the marey diagram when the user clicks in? I agree that the loading time might be long, but the marey chart doesn't make much sense as a standalone page the same way the isochrone does.

exxonvaldez:
> How could we optimize arrival history data for a Marey chart? Unless this is a simple optimization, my gut reaction is that it isn't worth optimizing this early in the process; we should create a proof of concept and only optimize if there's user demand.

The arrivals (say 5,000 of them) are currently organized by stop.  The client has to go through all of them and group them by trip id.  It doesn't take too long, maybe a few seconds on my laptop.  So we can skip optimizing this for now.

> As for having the chart on its own screen, could you show just a small slice of the chart on the trip summary page (hopefully a slice that's fast to load) and only show the marey diagram when the user clicks in? I agree that the loading time might be long, but the marey chart doesn't make much sense as a standalone page the same way the isochrone does.

Since the data for the day isn't sliced in any way, it all has to be loaded and processed and plotted.  So I don't see an easy way to "preview" it.

Agreed that it's not standalone.  It works well as part of the route summary.  Potentially it could go on the already lengthy trip-to-trip page also.

I've incorporated some early feedback.  There's now a highlight (thicker line) when you hover over a trip, you can filter the chart to inbound only or outbound only, and you can use the time range picker to show just part of the day.  Marking as ready for review.

Restricting the chart to a range of stops is actually kind of complicated, for routes with more than just one inbound and outbound direction.  Right now we are just flattening all the directions together.  So I'm going to hold off on doing a range of stops for now.

frontend/src/components/MareyChart.jsx
```diff
@@ -0,0 +1,348 @@
...
+    return tripSeriesArray;
+  }
+
+  let series = null;
+  if (processedArrivals) {
+    series = createSeries(processedArrivals);
+  }
+  
+  const graphParams = props.graphParams;
+  const startHour = graphParams.start_time ? parseInt(graphParams.start_time) : 3;
```
youngj: if graphParams.start_time is not defined, could the default startHour be the hour of the start_time timestamp in the arrivals file?

frontend/src/components/MareyChart.jsx
```diff
@@ -0,0 +1,348 @@
...
+            yPercent={1.0 - 85.0/((endHour-startHour)*100.0) }
+            style={{
+              textAnchor: 'end'
+            }}       
+          />
+          {hintValue ?
+            <Hint
+              value={hintValue}
+              format={ hintValue => [{title: 'Stop', value: hintValue.title },
+                                     {title: 'Time', value: `${(Math.floor(hintValue.minutes / 60) + 3)}:${Math.round(hintValue.minutes % 60).toString().padStart(2, '0')}`},
```
youngj: Could hintValue store the time string so it doesn't need to be recreated from hintValue.minutes?

frontend/src/components/MareyChart.jsx
```diff
@@ -0,0 +1,348 @@
...
+            />}
+            label={'Outbound only'}
+          /> 
+          
+            
+        </div>
+        </FormControl>  
+  
+        <XYPlot height={(endHour-startHour) * 100} width={600}
+          yDomain={[endHour, startHour] /* 3am the next day at the bottom, 3am for this day at the top */}
```
youngj: For some routes there are big empty spaces at the top/bottom of the chart. Could the y axis start at the start of the hour where the first bus was observed and end at the end of the hour when the last bus was observed?

frontend/src/components/MareyChart.jsx
```diff
@@ -0,0 +1,348 @@
...
+      // This is a little clunky -- for all outbound routes, we restate the distance
+      // as distance in the inbound direction by subtracting the stop's distance from
+      // the length of the outbound direction.  This does not line up exactly with the
+      // inbound direction length.
+      
+      if (directionInfo.name === 'Outbound') {
+        distance = directionInfo.distance - distance;
+      }
+      distance = metersToMiles(distance);
+      tripData[tripID].series.push({
```
youngj: For vehicles that remain at one stop for a significant amount of time (maybe ~2 minutes or longer?) would it be possible to add a second point with the y coordinate based on arrival.e so that we can visualize the dwell time?

frontend/src/components/MareyChart.jsx
```diff
@@ -0,0 +1,348 @@
...
+        stopID: stopID,
+        title: route.stops[stopID].title,
+        minutes: (arrival.t - start_time)/60,
+        vehicleID: vehicleID,
+        x: distance,
+        y: (arrival.t - start_time)/60/60 + 3.0, // convert to number of hours since midnight, assume 3am start time for now
```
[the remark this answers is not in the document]
exxonvaldez: Good catch, I made a bad assumption.  The +3 was a quick and dirty conversion from hours since start_time to time of day.  Showing the time of day on the chart makes it easier to find a known anomaly.  Any thoughts on how to make this conversion without dealing with time zones?  How does the backend know what the start times are?

frontend/src/components/MareyChart.jsx
exxonvaldez: Do you mean creating a time string for each data point (line 145)?  Basically move this calculation to there?

frontend/src/components/MareyChart.jsx
exxonvaldez: Good idea, I can work on that.

frontend/src/components/MareyChart.jsx
exxonvaldez: Also a good idea.

frontend/src/components/MareyChart.jsx
youngj: yeah

frontend/src/components/MareyChart.jsx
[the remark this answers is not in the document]
youngj: I'd use a time zone for this, could use https://momentjs.com/timezone/docs/ . The backend sets the start time for each route in compute_arrivals.py (around line 17)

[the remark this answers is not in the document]
exxonvaldez: Now using Moment to get the offset from "hours since start_time" to "time of day".

frontend/src/components/MareyChart.jsx
exxonvaldez: I went with your later suggestion of using the hour of the earliest arrival seen.

frontend/src/components/MareyChart.jsx
exxonvaldez: Ok, using Moment to store a time string in the data point.

frontend/src/components/MareyChart.jsx
exxonvaldez: Okay, this is done -- for the "All Day" time range.   Otherwise time range sets the bounds of the chart.

frontend/src/components/MareyChart.jsx
exxonvaldez: Okay, added second data point for dwells exceeding 2 minutes.  Main effect is at the terminals, where you can see overlapping inbound and outbound dwells.

PR #259, opened 2019-08-22
Feature/genUrls

### LINKED ISSUE #217: Generate URL Path When Using Map or DropDowns
Currently, no URL is generated when populating data through the use of the map or dropdowns.
As of PR #211 , we now support GET operations.
In order for users to share and use these URLs, we should auto-populate the URL parameters with the appropriate ids (i.e. /direction/14___O_F00) when selecting an inbound direction for route 14

ckrajewski:
URLs now generated off of drop down options
Supports selecting options from the map too :)
Fixes #217


frontend/src/components/ControlPanel.jsx
```diff
@@ -54,6 +64,29 @@ function ControlPanel(props) {
       : 0;
     return secondStopInfo.stops.slice(secondStopListIndex + 1);
   }
+  
+  /*
+    sets the URL based on dropDown values
+  */
+  const setPath = (pathParam,id,path = document.location.pathname) => {
```
ckrajewski: maybe we should move this method to a util?
So we can reference it from map eventually

frontend/src/components/ControlPanel.jsx
```diff
@@ -54,6 +64,29 @@ function ControlPanel(props) {
...
+    if(endingPathIndex === -1){
+      return `${path}/${pathParam}/${id}`;
+       
+    }
+    //otherwise, we need to cut off the URL and add latest parameter
+    pathArray[endingPathIndex+1]=id;
+    return pathArray.slice(0,endingPathIndex+2).join('/');
+
+  }
+  const commitPath = path => push(path);
```
ckrajewski: same with this small helper method

frontend/src/routesMap.js
```diff
@@ -15,7 +15,7 @@ export default {
     the ? after the : means an optional paramter variable
     ()* shows am optional parameter label
     */
-    path: '/route/:route_id/(direction)*/:direction_id?/(start_stop)*/:start_stop_id?/(end_stop)*/:end_stop_id?',
+    path: '/route/:route_id/(direction)*/:direction_id?/(from_stop)*/:start_stop_id?/(to_stop)*/:end_stop_id?',
```
exxonvaldez: Since this path is just a String, can you use the constants in `routeUtil` to build the string.  Maybe like:

 ```
  path: `/${ROUTE}/:route_id/(${DIRECTION})*/:direction_id?/...`,
```

ckrajewski: @exxonvaldez 
Added util and fixed the param names, which caused the back end calls to not work before.
Also incorporated a builder pattern for constructing URLs, so feel free to take a look

Issue #252, opened 2019-08-22
UI for "Bus Stop Metrics Table" could some clarity enhancements


professionalzack:
The "table" on the top of the right-hand side when you select a bus stop is not easy to read. An actual table might help, maybe a little bit of color and/or borders for the columns/rows. This should not be too difficult in Material UI
![Screen Shot 2019-08-21 at 7 47 28 PM](https://user-images.githubusercontent.com/26235535/63482261-1bd54b80-c44d-11e9-9618-27123bef6b87.png)

PR #260, opened 2019-08-23
imports and implements Table component from material ui

professionalzack:
Addresses #252 
coded alongside @katjone 

## Proposed changes
Still ugly, but it is more readable. More importantly, we transitioned the table from an HTML table to a material-ui-react Table component. The code is a lot easier to look at imo, and also easier to edit.
The table row text is written into an array above the return, and then mapped in. The table, and various table elements can be customized using the [material-ui Table APIs](https://material-ui.com/components/tables/#api). I would suggest a smaller font-size, which we well probably check out later, but we wanted to make sure to at least get this in today. Okay, thanks for being so welcoming last night, we both had a great time! 
Table-Row generation:
Current State of Tablehood

### DIFF
--- a/frontend/src/components/Info.jsx
+++ b/frontend/src/components/Info.jsx
@@ -209,16 +209,11 @@ class Info extends Component {
     const speed = tripTimes && distance ? (distance / (tripTimes.avg / 60.0)).toFixed(1) : 0; // convert avg trip time to hours for mph
     const grades = speed ? this.computeGrades(headwayMin, waitTimes, tripTimes, speed) : null;
 
-    
-    const addNewRow = (metric, value, grade, score) =>{ 
-      return {metric, value, grade, score}
-    };
-
     const tableRows = !grades && [] || [
-      addNewRow('Average Wait', (`${Math.round(waitTimes.avg)} minutes`), grades.averageWaitGrade, grades.averageWaitScore),
-      addNewRow('20 min wait probability', `${Math.round(grades.longWaitProbability * 100)}% ${ grades.longWaitProbability > 0 ? `(1 time out of ${Math.round(1 / grades.longWaitProbability)})` : ''}`, grades.longWaitGrade, grades.longWaitScore),
-      addNewRow('Travel Time', `Average time ${Math.round(tripTimes.avg)} minutes (${ speed } mph)`, grades.speedGrade, grades.speedScore),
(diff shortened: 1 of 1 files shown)


frontend/src/components/Info.jsx
```diff
@@ -209,6 +209,18 @@ class Info extends Component {
     const speed = tripTimes && distance ? (distance / (tripTimes.avg / 60.0)).toFixed(1) : 0; // convert avg trip time to hours for mph
     const grades = speed ? this.computeGrades(headwayMin, waitTimes, tripTimes, speed) : null;
 
+    
+    const addNewRow = (metric, value, grade, score) => {  
+      return {metric, value, grade, score}
+    }
+
+    const tableRows = [
+      grades && addNewRow('Average Wait', (<><div>{`${Math.round(waitTimes.avg)} minutes`}</div><div>{`${PLANNING_PERCENTILE}% of waits under ${ Math.round(getPercentileValue(waitTimes, PLANNING_PERCENTILE)) } minutes`}</div></>), grades.averageWaitGrade, grades.averageWaitScore),
```
akgupta89: Great job and excellent first contribution!

Some small fixes I noticed:
1. You have an empty <> and corresponding </> tag which aren't needed
2. Instead of using addNewRow, can you nest the object directly inside? e.g.
`{'Average Wait', (<div>....</div>),   grades.averageWaitGrade, grades.averageWaitScore}` 
3. I'm not sure what the 'grades && ' are for. It seems unnecessary. Could you try removing it?

This massively cleans up our UI, great work!

professionalzack: 1. The `<>...</>` jsx fragment was necessary to avoid element adjacency, but this issue is solved by taking @exxonvaldez 's suggestion that we pull "90% of waits" out of the table altogether. :)
2. The reason for the addNewRow vs 'nesting directly' is just to take advantage of shorthand property names. In your suggestion, a row would actually have to look like this: `{metric: 'Average Wait', value: (<div>....</div>), grade: grades.averageWaitGrade, score: grades.averageWaitScore}`. But, yeah, I am happy to switch it to that.
3. As @exxonvaldez said, the guards are necessary, but after looking it over, we were able to limit it to one guard: `const tableRows = !grades && [] || [...]`. I am quite happy w that. 

Okay, these are simple changes, I'll just do them right now. Thanks!

PR #262, opened 2019-08-26
Warn user if searching routes outside of service area

### LINKED ISSUE #163: Geolocation fails silently if you aren't in SF
If you hit the "routes near me" button and your geolocation isn't in SF, nothing happens -- nothing is shown on the map. We should show an error if your lat/long is outside some bounding box around san francisco.

TheLordHighExecutioner:
Fixes #163

## Proposed changes
Currently, the application silently fails if the user selects a coordinate outside of the service area (include when the user's geolocation is outside of the service area).

The proposed change alerts the user when they are outside of the service area.
...

## Screenshot

Kapture 2019-08-26 at 16 37 16

...

## Link to demo, if any

(I can't seem to get Heroku to work, here is where the demo should be:)
https://metrics-mvp-service-area.herokuapp.com/
...

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -1098,6 +1098,39 @@
       "resolved": "https://registry.npmjs.org/@nodelib/fs.stat/-/fs.stat-1.1.3.tgz",
       "integrity": "sha512-shAmDyaQC4H92APFoIaVDHCx5bStIocgvbwQyxPRrbUY20V1EYTbSDchWbuwlMG3V17cprZhA6+78JfB+3DTPw=="
     },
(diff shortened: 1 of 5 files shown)


TheLordHighExecutioner:
> Looks good.
> 
> I saw the autobind decorator came and went, just out of curiosity, any particular reason for not using it? In the long run, we are headed towards using functional components instead of class components, which may make how binding is done moot.

Since we were moving towards functional components, I decided that it wasn't worth the hassle to add a new dependency.

> Edit: also, what did you use to make the nifty screen capture?

I used Kap: https://getkap.co/

PR #274, opened 2019-08-29
[frontend] Updated security vulnerabilities in frontend packages

TheLordHighExecutioner:
Ran `npm audit fix`

## Proposed changes
Frontend had over a thousand NPM security warnings. I ran `npm audit fix` to address them.
...

## Link to demo, if any

https://metrics-mvp-audit-fix.herokuapp.com/

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -5,34 +5,44 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.0.0",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0.tgz",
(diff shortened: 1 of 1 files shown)


akgupta89:
I'm worried that this won't help enforce consistency until everyone uses `npm ci` instead of `npm install` to use package-lock.json while building dependencies.

hathix:
@TheLordHighExecutioner what do you want to do about this?

hathix:
I'm going to abandon this because there are too many merge conflicts & it will keep getting worse. The problem is that package-lock keeps getting regenerated each time someone installs something, so our changes on this will keep getting clobbered. As good as it is to fix the NPM audit, I don't think it's worth it for now.

PR #276, opened 2019-08-30
[frontend] Move SF-specific strings to `locationConstants.js`

### LINKED ISSUE #254: Refactor the codebase to pull out all SF-specific strings in constants files
Part of the OT For Any City project; [see the spec](https://docs.google.com/document/d/1VoGpaReLsnudHk2LOR-BaGZAAb1p_jI1Le8cugpSj1U/edit#).

Basically, make it so that we could easily swap this to any other city with little more than some string changes.

TheLordHighExecutioner:
## Proposed changes

Makes the product more generic by moving location specific content to a single file

## Link to demo, if any

https://metrics-mvp-sf-strings.herokuapp.com/


exxonvaldez:
Looks like a good start.  A couple of things:

- instead of transitName, can you use agencyName?  This is the nomenclature that Nextbus uses (specifically the string "sf-muni" is to match the agency argument in their APIs: https://www.nextbus.com/xmlFeedDocs/NextBusXMLFeed.pdf ).  Also it'd be good to have two variables extracted out, perhaps agency (sf-muni) and agencyName (SF Muni) or like agencyCode and agencyDisplayName.

- In the `Map`-based components (three of them), there are default starting coordinates that need to be moved to this file.  This is a good chance to refactor so all the components share the same starting coordinates.

- If/when we do similar changes for the backend files, how should the frontend and backend share constants?

hathix:
@exxonvaldez , do the new commits address your comments?

hathix:
Boy, lots of merge conflicts. Who wants to take on the task of resolving them?

TheLordHighExecutioner:
Okay, I'm having trouble with `isochrone-worker.js` and the Web Worker pattern. For some reasoning, I am unable to import from `locationConstants` into `isochrone-worker.js`. Does anyone know how to fix this?

exxonvaldez:
> Okay, I'm having trouble with `isochrone-worker.js` and the Web Worker pattern. For some reasoning, I am unable to import from `locationConstants` into `isochrone-worker.js`. Does anyone know how to fix this?

Looks like React doesn’t support webpacking or compilation for worker files, I’m not sure.  See this React feature request:  https://github.com/facebook/create-react-app/issues/3660 .  There are some hairy workarounds at the end of that, I guess the idea being that you pass javascript functions from the React side (from source files that are using imports and were transpiled) to the worker.

Given that we don’t have a fully thought out plan for making OpenTransit support multiple cities or agencies yet, we can leave the worker alone for now and hope that the framework supports this later.  The workarounds seem kind of brittle and not promising.

exxonvaldez:
Alternatively, you can refactor the app to worker code to pass all the locale-specific variables from the React app to the worker as messages.

Issue #200, opened 2019-07-25
Make the trip info table look more like the RouteSummary component


hathix:
#176 made a great new component for the route summary:

![Screenshot 2019-07-24 at 7 21 38 PM](https://user-images.githubusercontent.com/728084/61841078-534dd980-ae48-11e9-96b5-93399f88be51.png)

But the trip info table is still pretty old-school:

![Screenshot 2019-07-24 at 7 21 50 PM](https://user-images.githubusercontent.com/728084/61841077-534dd980-ae48-11e9-913f-3bedd38e0bd0.png)

It would be good to make this one look more like the fancy RouteSummary component. (To be clear - it doesn't need to be a table! Just some good way to visualize the data).

exxonvaldez:
Work in progress:

![Screen Shot 2019-09-13 at 11 03 01 AM](https://user-images.githubusercontent.com/44861283/64886426-0520af80-d61b-11e9-8a46-1bc99abb2395.png)

PR #281, opened 2019-09-05
UI enhancements

exxonvaldez:
## Proposed changes

For reporting of stop to stop data, Jesse suggested using tabs to make the display more compact and require less scrolling.  In addition, this allows the `ControlPanel` to go on the right column, letting `MapStops` take the full height on left.

- Info.jsx: move stop to stop score table, time of day chart, and histograms behind a new row of 5 tabs.  Remove Cards that were wrapping these items.
- InfoIntervalsOfDay: remove Card wrapping the chart.
- MapSpider: update resizing code to avoid componentWillMount().
- MapStops: add resizing code so that map is full height of window, like MapSpider.
- RouteSummary:  tweak Grid of cards so that all cards are rendered at the same height regardless of contents.
- RouteScreen:  move ControlPanel to right hand column so that map has the entire left column.

## Screenshot

At route summary level, `ControlPanel` on right and cards are same height:

Screen Shot 2019-09-05 at 12 28 52 PM

New tab bar:

Screen Shot 2019-09-05 at 12 29 09 PM

Another tab shown:

Screen Shot 2019-09-05 at 12 29 25 PM

For windows less than 640px wide, note scrolling in tab bar:

Screen Shot 2019-09-05 at 12 29 49 PM

## Link to demo, if any

https://ot-terence.herokuapp.com (if not demoing Marey chart branch)


frontend/src/components/MapStops.jsx
```diff
@@ -13,6 +13,41 @@ const SF_COORDINATES = { lat: 37.7793, lng: -122.419 };
...
+  // case make the map half height.
+  //
+  // TODO: Need to convert this component to a functional component.  Then we can use the useTheme
+  // hook to programatically access the breakpoint widths.
+  //
+  // Note: This code has to be adjusted to be kept in sync with the UI layout.
+  //
+  
+  computeHeight() {
+    return (window.innerWidth >= 640 ? window.innerHeight : window.innerHeight/2) - 64 /* blue app bar */;
```
akgupta89: 🔥 LGTM, much needed addition.

It'd be nice if this was done in the CSS rather than JS, but that would be a good future state target, because I know we need a whole CSS reworking and probably start using a CSS pre processor.

akgupta89:
Also you should point out that you introduced aria labels which is a move in the right direction but probably something we should limit to a specific ticket which should be introduced in all files in a standard way.

frontend/src/components/Info.jsx
```diff
@@ -307,26 +314,48 @@ class Info extends Component {
...
+        <br/>
+        <AppBar position="static" color="default">
+          <Tabs value={this.state.tabValue} onChange={this.handleTabChange.bind(this)}
+             aria-label="tab bar"
+             variant="scrollable"
+             scrollButtons="on">
+            <Tab style={{ minWidth: 72 }} label="Summary" {...a11yProps(SUMMARY)} />
+            <Tab style={{ minWidth: 72 }} label="Time of Day" {...a11yProps(TIME_OF_DAY)} />
+            <Tab style={{ minWidth: 72 }} label="Headways" {...a11yProps(HEADWAYS)} />
+            <Tab style={{ minWidth: 72 }} label="Waits" {...a11yProps(WAITS)} />
```
youngj: I think the labels would be clearer as "Wait Times" and "Trip Times"

frontend/src/components/Info.jsx
```diff
@@ -307,26 +314,48 @@ class Info extends Component {
...
-            <Card>
-              <CardContent>
-                <span className="h4">Overall Grade: </span>
-                <span className="h1">{grades.totalGrade}</span> ({' '}
-                {grades.totalScore} / {grades.highestPossibleScore} )
-                <div>
+            <Box p={2} hidden={this.state.tabValue !== SUMMARY} >
+            
+                <Typography variant="h5" display="inline">Trip Grade:{' '}
+                {grades.totalGrade}{' '}
```
youngj: What do you think about making the Summary tab look more like the cards for the overall route grade (with colored background / stars), instead of a table with just numbers and text? (Could be separate change)

frontend/src/components/MapStops.jsx
exxonvaldez: I agree, although my Google searches led me to think this might be a JS-only thing.  You can use css to specify a percentage of viewport height, but I don't think you can do the "-64" part to manually adjust for the presence of other UI elements.

exxonvaldez:
> In the RouteTable, what do you think about adding rectangles of varying color/width next to each number to make it more visually interesting and easier to see good/bad wait times/trip times/scores at a glance?

Generally sounds good as a separate issue.  One thing I meant to get to with the route table/summary was adding the other two sub scores, long wait probability (as a placeholder for on time performance) and travel time variability (rarely an issue but interesting when it is).  I want to see if adding these subscores clarifies or muddies the picture.  With 4 subscores, we might need a  subtle and compact way of highlighting values of interest, like text/background colors.  Horizontal bars could still work.

One example:  https://www.metacritic.com/browse/games/release-date/new-releases/ps4/metascore but more subdued.

frontend/src/components/Info.jsx
exxonvaldez: > What do you think about making the Summary tab look more like the cards for the overall route grade (with colored background / stars), instead of a table with just numbers and text? (Could be separate change)

Yeah, definitely want this.  Neel filed this as #200 .

frontend/src/components/Info.jsx
exxonvaldez: > I think the labels would be clearer as "Wait Times" and "Trip Times"

Okay, done.

PR #282, opened 2019-09-06
Beautify Round #4 into Beautify Round #5

akgupta89:
Beautify Round#4 work done on top of #279, so either merge this into Round#5 and then into master or merge this into master after merging #279 . No modification to code or logic, just another round of beautification

- Moved eslint related dependencies into devDependencies
- Removed unused prop-types in Functional Components
- Fixed used before defined for functions
- Fixed shadowing of variables (sub-scoped variables with same names)
- Converted for loops to forEach, map, and find iterators
- Converted unary operators to `+= 1;`
- Resolved no-unused-vars where possible
- Removed disabled eslint statement
- Set button prop `type='button'`

Down to 28 linting errors remaining! (assuming none are hiding that will reveal themselves after this round is finished)


frontend/src/components/Home.jsx
```diff
@@ -99,17 +96,13 @@ const mapStateToProps = state => ({
 const mapDispatchToProps = dispatch => ({
   fetchData: (graphParams, intervalParams) =>
     dispatch(fetchData(graphParams, intervalParams)),
-  resetGraphData: params => dispatch(resetGraphData()),
+  resetGraphData: params => dispatch(resetGraphData(params)),
```
exxonvaldez: Seems odd given the effect it has on actions/index.js.  Couldn't params be removed here like:

`resetGraphData: () => dispatch(resetGraphData())`

frontend/src/components/MapStops.jsx
```diff
@@ -314,8 +313,8 @@ class MapStops extends Component {
               ),
             );
           }
-          index++; // use a loop keeps the index consistent with direction and also stop color
-        }
+          index += 1; // use a loop keeps the index consistent with direction and also stop color
```
exxonvaldez: Do you have a reference that explains why `+= 1` is preferred over `++`?  Also, here it looks like here you can get `index` for free as part the `forEach` callback:

`selectedRoute.directions.forEach((direction, index) => { `...

frontend/src/components/Navigation.jsx
```diff
@@ -1,10 +1,11 @@
 import React, { Fragment } from 'react';
 import { NavLink } from 'redux-first-router-link';
 
+// eslint-disable-next-line no-unused-vars
 export default props => {
   return (
     <Fragment>
-      <button>
+      <button type="button">
```
exxonvaldez: This file is now dead code, let's just remove it instead of modifying it.

frontend/src/screens/Dashboard.jsx
```diff
@@ -62,9 +62,9 @@ const mapStateToProps = state => ({
 const mapDispatchToProps = dispatch => ({
   fetchData: (graphParams, intervalParams) =>
     dispatch(fetchData(graphParams, intervalParams)),
-  resetGraphData: params => dispatch(resetGraphData()),
+  resetGraphData: params => dispatch(resetGraphData(params)),
   fetchGraphData: params => dispatch(fetchGraphData(params)),
-  resetIntervalData: params => dispatch(resetIntervalData()),
+  resetIntervalData: params => dispatch(resetIntervalData(params)),
```
exxonvaldez: Note that `fetchRoutes` doesn't take any params, so that reinforces the idea that `resetGraphData` and `resetIntervalData` should not either.

frontend/src/screens/Isochrone.jsx
```diff
@@ -475,14 +469,17 @@ class Isochrone extends React.Component {
     );
   }
 
+  // eslint-disable-next-line no-unused-vars
```
exxonvaldez: Seems like it's better to remove or comment out `event` than adding an eslint rule here?

frontend/src/components/MapStops.jsx
akgupta89: https://eslint.org/docs/rules/no-plusplus
as well as
https://github.com/airbnb/javascript/issues/540

I've seen people use the unary operator in hacky ways which hinders code understandability, and more modern languages like Python do not even have the operator. I tend to be reliant on it myself sometimes, but the one area I changed the linting rules was inside the for loop, which seems like a standard place to use it, just not outside loop.

frontend/src/components/Navigation.jsx
akgupta89: Got it, deleted.

frontend/src/components/Home.jsx
akgupta89: I didn't want to risk there being some unintended consequences by modifying the function input that calls the event, so I thought it'd be better just to use pass the params to the function. I removed it as you have suggested.

frontend/src/screens/Isochrone.jsx
akgupta89: The event that calls it mandatorily passes in an event to the method,

frontend/src/components/MapStops.jsx
exxonvaldez: Arun, what are your thoughts on using forEach((direction, index) => ...? instead of index += 1?

frontend/src/screens/Isochrone.jsx
exxonvaldez: True, but do we/ESLint care about the event argument being passed in?   https://stackoverflow.com/questions/48639264/check-for-correct-number-of-parameters-in-eslint 

You can have the "event" parameter here commented out just as a reminder that an event gets passed in.  Seems clumsier to have the parameter and then turn off no-unused-vars for the entire line.

frontend/package.json
```diff
@@ -5,15 +5,14 @@
   "dependencies": {
     "@date-io/date-fns": "^1.1.0",
```
exxonvaldez: As long as you're in here, please remove the following unused dependencies:

- @date-io/date-fns
- bootstrap
- material-ui-pickers
- rc-time-picker
- react-date-picker
- react-gradient-timepicker
- react-timekeeper
- react-times

Also update moment-timezone to ^0.5.26 (doing the same in the Marey chart branch).

frontend/src/screens/Isochrone.jsx
akgupta89: Hmm, it was causing issues last time I removed it, I must have did something else that did it. Removed.

frontend/src/components/MapStops.jsx
akgupta89: Good idea!

exxonvaldez:
Oh, seems like it would be a good practice to deploy to heroku first when there are build-related changes.  If you're able to finish getting it set up, it is well worth it.

frontend/src/components/MapSpider.jsx
```diff
@@ -54,21 +57,6 @@ class MapSpider extends Component {
...
-  //
-  // Note: This code has to be adjusted to be kept in sync with the UI layout.
-  //
-  updateDimensions() {
-    const height =
-      (window.innerWidth >= 640 ? window.innerHeight : window.innerHeight / 2) -
-      64; /* blue app bar */
-    this.setState({ height });
-  }
-
```
exxonvaldez: I'm not sure I get the reordering of these methods... moving this will cause a conflict with my UI tabs PR, where I am reimplementing this resize code slightly. componentWillMount's body moves into componentDidMount due a new warning against the use of componentWillMount in React 16.9.  Shouldn't updateDimensions stay above its caller?

hathix:
Arun, make sure you handle merge conflicts and any lingering feedback from Terence, then you can merge it.

PR #314, opened 2019-09-19
Enhancements to route summary, trip summary, and route table. 

### LINKED ISSUE #197: Making rating system accessible
In #176 we are using colored backgrounds to show the quartiles of scores (instead of the placeholder A/B/C/D we had before).  This is not fully accessible for color blind users, so we should use something like https://material-ui.com/components/rating/ (on a scale of 1 to 4 stars, or switch to pentiles for 5 stars) or a set of 4 icons or symbols to represent the quartiles.  Alternatively, we could have a graphical indicator of score, such as a "partial ring" around the score or a tiny horizontal bar that is partially filled in.

Another thing that can be split out as another issue:  The scores need some way of showing how they were calculated, possibly a Material UI Popper widget.

### LINKED ISSUE #200: Make the trip info table look more like the RouteSummary component

### LINKED ISSUE #201: Surface grading criteria when you click route scores
#176 added some cool cards to grade the routes based on median wait, median speed, and route score. But it's not immediately clear *why* they got that score.

Screenshot 2019-07-24 at 7 21 38 PM

There's actually [a very well-reasoned rubric for grading](https://docs.google.com/document/d/19eMnShDf8f6Qitq175uYil7StwzSkuVXSNDUfyeVCMw/edit) made by @exxonvaldez, but that rubric isn't surfaced.

Let's show that grading rubric in a modal dialog when you click on the colored grading cards. Just have a dialog that has some paragraphs of text from [that doc](https://docs.google.com/document/d/19eMnShDf8f6Qitq175uYil7StwzSkuVXSNDUfyeVCMw/edit) and perhaps a table where the appropriate star/grade rating is highlighted.

### LINKED ISSUE #286: RouteTable: Try long wait prob and travel variability subscores.  Make it more visually interesting and easier to see good/bad values at a glance
> In the RouteTable, what do you think about adding rectangles of varying color/width next to each number to make it more visually interesting and easier to see good/bad wait times/trip times/scores at a glance?

Generally sounds good as a separate issue.  One thing I meant to get to with the route table/summary was adding the other two sub scores, long wait probability (as a placeholder for on time performance) and travel time variability (rarely an issue but interesting when it is).  I want to see if adding these subscores clarifies or muddies the picture.  With 4 subscores, we might need a  subtle and compact way of highlighting values of interest, like text/background colors.  Horizontal bars could still work.

One example:  https://www.metacritic.com/browse/games/release-date/new-releases/ps4/metascore but more subdued.

_Originally posted by @exxonvaldez in https://github.com/trynmaps/metrics-mvp/pull/281#issuecomment-529812170_

### LINKED ISSUE #307: Show overall trip time on trip summary page
On [a trip summary page like this](http://muni.opentransit.city/route/12/direction/12___I_F00/fromStop/4677/toStop/4664), you can see the average wait and travel times, but not the overall total time (in this case, 8 + 12 = 20 minutes). I feel like this is the #1 thing people want to see, so we should highlight that number in the summary.

Screen Shot 2019-09-11 at 9 12 14 PM

exxonvaldez:
Fixes #307. Fixes #286. Fixes #201.  Fixes #200. Fixes #197.

## Proposed changes

- Stop to stop trip summary now shows total (wait + onboard) time first, for both planning and median cases.

- Route table now includes long wait probability and travel time variability.  Scores are shown using a colored background to emphasize them and make it easier to look for good/bad routes.

- Route summary also includes long wait and variability.

- Trip summary now uses cards like the route summary.

- Info bubbles on the cards provide explanations of how scores and subscores are calculated.

- The yellow for the card was adjusted to have better contrast with the rating stars.  The minimum star rating is half a star, because zero stars was confused with five stars.

Technical discussion:

- index.js: fetch larger precomputed wait/trip time files to get 20 min wait probability for route and 90% travel time for route
- InfoIntervalsOfDay: avoid console errors with NaNs, fixes some variables left behind by Arun's snake to camel case transition
- InfoScoreCard: new component that represents a single metric with a score, supports displaying a Popover that explains the score.
- InfoScoreLegend: new helper component for explaining scoring (a table with colored rows)
- InfoTripSummary: New component extracted from trip summary code formerly in `Info.jsx`.  Use new `InfoScoreCard` component.  Use common computeGrades method instead of separate one.  Use stop geometry to compute distances.
- MapStops: comment fix
- MareyChart: avoid crash when directionInfo is not present
- RouteSummary:  use new InfoScoreCard component instead of previous cards
- RouteTable:  Add columns for 20 min wait probability and travel variance minutes.  Color the score column for emphasis and ease of scanning.
- precomputed: Changes to use the larger wait/time precomputed files.
- routeCalculations: Changes to use the larger wait/time precomputed files.  Update the common method for computing grades to support four metrics rather than just two.

## Screenshot

Route table with new columns and colored score background:

Screen Shot 2019-09-18 at 8 33 10 PM

Enhanced route summary:

Screen Shot 2019-09-18 at 8 25 46 PM

With a Popover opened:

Screen Shot 2019-09-18 at 8 30 39 PM

Enhanced trip summary:

Screen Shot 2019-09-18 at 8 31 38 PM

## Link to demo, if any


exxonvaldez:
Was doing some spot testing, I just found and fixed a bug with fetching of routes in RouteScreen -- fell victim to shadowing of `fetchRoutes`.  No longer shadowing in that file.  Can somebody do some sanity testing on the demo site (ot-terence), and merge in if it works?  I know Arun's PR is waiting on this one.  Then we can iterate on this one some more.

youngj:
Thanks for doing this, I think the Summary tab looks much better.

- On the Summary tab, "Typical Trip" seems like the most important stat, so it seems like it should be first. Maybe this stat shouldn't include the wait time, though. A lot of people would plan their trips based on the predicted arrival time so they wouldn't actually be waiting the median wait time. This would also avoid needing to show how the typical trip stat is calculated as the sum of wait time and trip time (e.g. "14 min wait + 8 min"), which could be confusing to users 
- I think people would be confused by the stat labeled "90% of trips". Maybe instead of having a different card for this, the Typical Trip card could show something like "90% of trips less than XX minutes" underneath the median trip time.
- The median wait stat info bubble could explain that this is if you arrive at the origin stop at a random time (ignoring schedule/predictions)
- It's probably unclear to users what "Extra travel" card means .... took me a while to figure out that this is the 90th percentile trip time minus the median trip time. It seems like it would be hard to explain this in a way that does not confuse or overwhelm people. I'd prefer to remove Extra Travel from the summary and the route table to keep it simple and reduce the information overload. Maybe there could be an advanced stats tab for things like this, or it could just be in the Trip Score info bubble. 
- "20 Min Wait" also seems like it could be confusing to people (probability of having to wait 20 minutes if you show up at a random time ignoring schedule/predictions). Maybe this could go in an advanced tab rather than being in the summary tab and route table. 

Maybe the cards on the Summary tab could be Trip Time, Wait Time, Speed, Trip Score? I think reducing the number of top line stats like this would make it less overwhelming for users.

hathix:
I tested the demo site and the fetching seems to work normally.

hathix:
This PR is blocking some other work, so I'm going to take Jesse's comments and put them in a new issue.

PR #315, opened 2019-09-19
Beautify Round #7

akgupta89:
- Reordered package.json
- Installed cslx package to package.json
- Converted Info.jsx to functional component
  - Removed states in favor of react states
  - Reordered functional component functions
  - Removed arguments shadowing function variables
- camelCased variables


hathix:
I just tested the functionality manually and it works just fine. Haven't looked too hard at the actual code though.

frontend/src/components/MareyChart.jsx
```diff
@@ -68,15 +68,15 @@ function MareyChart(props) {
...
   const OUTBOUND = 'Outbound'; // same as directionInfo name
 
-  const { graphParams, fetchArrivals, arrivals, routes } = props;
+  const { graphParams, thisFetchArrivals, arrivals, routes } = props;
 
   // On first load, get the raw arrival history corresponding to graphParams.
 
   useEffect(() => {
     if (graphParams.routeId) {
       fetchArrivals(graphParams);
```
exxonvaldez: I think this should be `thisFetchArrivals` not `fetchArrivals`.

frontend/src/components/Info.jsx
```diff
@@ -250,92 +274,78 @@ class Info extends Component {
         {waitTimes ? (
           <Box p={2} hidden={this.state.tabValue !== WAITS}>
             <Typography variant="h5" display="inline">
```
exxonvaldez: Strange merge result here, or something.  Trip grade code here is under waits, but in latest code is now a separate component.

PR #322, opened 2019-09-30
count arrivals as belonging to separate trips if there is a gap of 3+ stops, or 30+ minutes of missing GPS observations

### LINKED ISSUE #285: Better handling of vehicles that take multiple hours to move to the next stop
As discussed wth @youngj @jtanquil , when we look at Marey charts for some of the lines (in particular light rail), there are many "trips" where a vehicle takes multiple hours to go from one stop to another or multiple hours to turn around.  This results in inaccurate travel time stats (like the KT averaging 1.6 mph on 9/4 am peak).  We should either detect that the vehicle is not in revenue service, or terminate the trip after a certain amount of time where there is no progress.

Screen Shot 2019-09-09 at 10 51 04 AM

Screen Shot 2019-09-09 at 10 50 18 AM

youngj:
The Marey chart highlighted some issues with the arrival time calculation, especially on Muni Metro routes, where there were many long gaps in arrival times for one vehicle that were being counted as part of a single trip (when most likely that vehicle was out of service during the gap):

Example (KT on 9/27):
image

To fix this issue, the arrival time algorithm now counts arrivals as belonging to separate trips if there is a gap of 3 or more stops, or if there is a gap of 30 or more minutes without any GPS observations from that vehicle (indicating that the vehicle was likely not in continuous service during that time).

The algorithm does not have any limit on the travel time between stops (e.g. 1 hour). It is possible that a bus may actually be in service between those two stops but waiting a long time in traffic. With some transit systems it may also be normal for a vehicle to take a long time between stops. This makes it hard to come up with a heuristic for splitting trips based on the travel time between stops.

With this change, the Marey chart looks much more reasonable, although there are still occasional cases with a large travel time between 2 stops (usually between the 1st and 2nd stop or 2nd-to-last and last stop for a vehicle in a particular direction, due to the vehicle waiting in between those two stops before starting the next run).

Example (KT on 9/27):
image

In order to avoid loading old arrival times and statistics from the cache, the version codes for the arrival times and precomputed wait time/trip time stats have been changed, and the historical stats will need to be regenerated.

Mostly fixes #285

### DIFF
--- a/backend/models/eclipses.py
+++ b/backend/models/eclipses.py
@@ -35,6 +35,22 @@ def resample_bus(bus: pd.DataFrame) -> pd.DataFrame:
 
     new_rows = []
 
+    def make_separator_row(vid):
+        # adding a separator row at the end of each vehicle's observations allows simplifying
+        # get_possible_arrivals_for_stop() (and making it slightly faster).
+        # this row will always be filtered out by find_arrivals
+        # so the row index will always have a gap in it even if two vehicles
+        # adjacent in the buses frame both happen to be near the same stop
+        return (
+            vid,
+            '',
+            0,
+            0,
+            0,
+            0,
+            # 0 # uncomment for debugging
+        )
+
     if not bus.empty:
         # faster to use numpy arrays instead of pandas series
         time_values = bus['TIME'].values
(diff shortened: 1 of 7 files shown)


frontend/src/actions/index.js
```diff
@@ -144,10 +144,12 @@ export function fetchArrivals(params) {
...
     const dateStr = params.date;
 
-    const s3Url = `https://opentransit-stop-arrivals.s3.amazonaws.com/v4/sf-muni/${dateStr.replace(
+    const arrivalsVersion = 'v4a';
+
+    const s3Url = `https://opentransit-stop-arrivals.s3.amazonaws.com/${arrivalsVersion}/sf-muni/${dateStr.replace(
       /-/g,
       '/',
-    )}/arrivals_v4_sf-muni_${dateStr}_${params.routeId}.json.gz`;
+    )}/arrivals_${arrivalsVersion}_sf-muni_${dateStr}_${params.routeId}.json.gz`;
```
exxonvaldez: Thanks for making all the necessary front end updates.  Just noting for myself that this URL should live in locationConstants.js, will create an issue for it.

backend/models/eclipses.py
```diff
@@ -479,7 +504,8 @@ def set_trip(dir_arrivals):
 
         stop_index_diff = np.diff(stop_index_values, prepend=-999999)
 
-        new_trip = stop_index_diff <= 0
+        # if vehicle not observed at 3 or more stops in a row, count it as starting a new trip
+        new_trip = (stop_index_diff <= 0) | (stop_index_diff > 3)
```
EddyIonescu: Just wondering, from the sample it looks like the 30-minute limit would be enough? Also, given Nextbus' flakyness I'm wondering if this could end up causing a bunch of new/split trips, especially in cases of detours where buses usually disappear from Nextbus until they're back on route.

youngj: I first tested with only the 30 minute limit and it fixed most cases, but there were still some cases where a vehicle didn't have a 30 minute gap in GPS observations but still had a large gap in the stop_index of adjacent arrivals. Splitting trips that skipped 3 stops fixed most of these remaining cases. I guessed that in most cases of detours a vehicle might skip 1 or 2 stops but is unlikely to skip 3 or more stops (let me know if you have any counter-examples).

backend/models/eclipses.py
```diff
@@ -64,52 +80,53 @@ def resample_bus(bus: pd.DataFrame) -> pd.DataFrame:
         vid = bus['VID'].values[0]
         did_values = bus['DID'].values
 
+        obs_group = 1
```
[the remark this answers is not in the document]
youngj: OK, added a comment

PR #324, opened 2019-10-02
Dashboard: Fix sorting of routes by name (title).

### LINKED ISSUE #164: Improve route sorting

exxonvaldez:
Fixes #164.

## Proposed changes

When sorting by route name (internally `title`), use the natural order of the routes array, which comes from Nextbus. It is the simplest way to get a "good" ordering for the routes without having to implement a hybrid numeric/lexical sort.

## Screenshot

Screen Shot 2019-10-01 at 9 30 28 PM

## Link to demo, if any

n/a

### DIFF
--- a/frontend/src/components/RouteTable.jsx
+++ b/frontend/src/components/RouteTable.jsx
@@ -27,6 +27,12 @@ import {
 
 import { handleGraphParams, fetchPrecomputedWaitAndTripData } from '../actions';
 
+/**
+ * 
+ * @param {any} a
+ * @param {any} b
+ * @param {any} orderBy
+ */
 function desc(a, b, orderBy) {
   // Treat NaN as infinity, so that it goes to the bottom of the table in an ascending sort.
   // NaN needs special handling because NaN < 3 is false as is Nan > 3.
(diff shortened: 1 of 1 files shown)


frontend/src/components/RouteTable.jsx
```diff
@@ -50,7 +50,31 @@ function desc(a, b, orderBy) {
...
+ * 
+ * Sorting by title is a special case because the original order of the routes array is
+ * better than sorting route title alphabetically.  For example, 1 should be followed by
+ * 1AX rather than 10 and 12.
+ *  
+ * @param {Array} array      Array to sort
+ * @param {Function} cmp     Comparator to use
+ * @param {String} sortOrder Either 'desc' or 'asc'
+ * @param {String} orderBy   Column to sort by
+ */
```
akgupta89: It's usually good to include a `@return` definition here

exxonvaldez: Ok, @returns added.

PR #325, opened 2019-10-02
Refactor static  colors codes to variables tha can be use in an easie…

### LINKED ISSUE #209: Refactor out static color codes
We hardcode in several colors into the UI, [such as purple](https://github.com/trynmaps/metrics-mvp/search?utf8=%E2%9C%93&q=purple&type=). We should factor that out and keep all the branding colors we use in the UI in some CSS/JS constants files. This way we can easily update our brand colors.

DanielAguirre:
Refactor colors static codes to variables that can be used in an easier way thrown the application

Fixes #209

## Proposed changes
Set colors code as variable to make easier reuse it or change it
```
export const Colors = {
  GRAY: '#a4a6a9',
  PURPLE: '#aa82c5',
  BLUE: 'blue',
  RED: 'red',
  GREEN: 'green'
}
```

### DIFF
--- a/frontend/src/UIConstants.js
+++ b/frontend/src/UIConstants.js
@@ -2,8 +2,17 @@
  * Constants for the UI that allow for reconfiguration.
  */
 
+// Colors definition:
+// This section its should be use to declare the colors
+// used in the application
+export const GRAY = '#a4a6a9';
+export const PURPLE = '#aa82c5';
+export const BLUE = 'blue';
+export const RED = 'RED';
+export const GREEN = 'green';
+
 // placeholder colors: gray and purple from nyc busstats
-export const CHART_COLORS = ['#a4a6a9', '#aa82c5'];
+export const CHART_COLORS = [GRAY, PURPLE];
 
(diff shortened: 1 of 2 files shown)


frontend/src/UIConstants.js
```diff
@@ -2,8 +2,17 @@
  * Constants for the UI that allow for reconfiguration.
  */
 
+// Colors definition:
+// This section its should be use to declare the colors
+// used in the application
+export const GRAY = '#a4a6a9';
```
akgupta89: The one thing I'd suggest is making it an Object e.g.:

```
export const Colors = {
   GRAY: ...  
};
```

So we only have to include one object and it's very clear what it is being used for? Other than that, this is a great contribution!

DanielAguirre: Thanks for the feedback, I'll start to work on this change :smile:

DanielAguirre: I just update the pull request with the  recommended changes

PR #330, opened 2019-10-08
MapStops improvements

### LINKED ISSUE #168: Put a legend on the trip information map
There are a lot of colors flying around -- blue and red for inbound/outbound stops, red/green for start/end stops. Can we add that to the legend (which just shows speed colors right now)?

Screenshot 2019-07-10 at 8 50 01 PM

### LINKED ISSUE #291: Finish #173 (map stop legend)
Some random person started PR #173 but never finished it. I'm just making this Issue to remind everyone we need to finish it.

### LINKED ISSUE #294: Make all stops the same color
The red/blue are a lot of color to take in (especially since there's color for the speed) and nobody really cares what's inbound or outbound.

### LINKED ISSUE #295: Add custom symbols/design for origin/destination stops
You can copy what Google maps does (put the red pin marker on the start and end stop, put the name there and bold it) or like put a square for start and triangle for end, or put a checkered flag at the end, or whatever. Just don't use the red/green fills because that clashes with the red/green we use for speeds.

exxonvaldez:
## Proposed changes

Changes the map of stops so that a larger map legend isn't needed:
- Make all stops the same indigo color
- Add "from stop" and "to stop" icons in black.  The model for this is the Google Maps UI, where these icons appear next to the corresponding form inputs as well on the map.

File level changes:
- `ControlPanel`: add "from stop" and "to stop" icons next to the pulldown menus
- `MapStops`:
  - add "to stop" and "from stop" icons to the map
  - make all other stops the same color (indigo)
  - rename `populateRouteDirection` to `populateStops`
  - move the call to `populateSpeed` out of `populateStops` so that the speed lines stay under all the stop markers of all directions
- `UIConstants`: new indigo color from the Material UI palette for stop markers.

## Screenshot

Screen Shot 2019-10-08 at 2 32 56 PM

## Link to demo, if any

### DIFF
--- a/frontend/src/UIConstants.js
+++ b/frontend/src/UIConstants.js
@@ -2,6 +2,8 @@
  * Constants for the UI that allow for reconfiguration.
  */
 
+import indigo from '@material-ui/core/colors/indigo';
+
 // Colors definition:
 // This section its should be use to declare an object color
 // that contain the colors used in the application
@@ -10,7 +12,8 @@ export const Colors = {
   PURPLE: '#aa82c5',
   BLUE: 'blue',
   RED: 'red',
-  GREEN: 'green'
+  GREEN: 'green',
+  INDIGO: indigo[400],
 }
 
 // placeholder colors: gray and purple from nyc busstats
--- a/frontend/src/components/ControlPanel.jsx
+++ b/frontend/src/components/ControlPanel.jsx
@@ -3,6 +3,7 @@ import React from 'react';
 import { connect } from 'react-redux';
 // import PropTypes from 'prop-types';
 import { makeStyles } from '@material-ui/core/styles';
+import Box from '@material-ui/core/Box';
 import Input from '@material-ui/core/Input';
 import InputLabel from '@material-ui/core/InputLabel';
 import MenuItem from '@material-ui/core/MenuItem';
@@ -11,6 +12,8 @@ import Select from '@material-ui/core/Select';
 import Grid from '@material-ui/core/Grid';
 import { handleGraphParams } from '../actions';
 import { ROUTE, DIRECTION, FROM_STOP, TO_STOP, Path } from '../routeUtil';
+import LocationOn from '@material-ui/icons/LocationOn';
+import TripOrigin from '@material-ui/icons/TripOrigin';
 
 const useStyles = makeStyles(theme => ({
   root: {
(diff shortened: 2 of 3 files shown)


akgupta89:
I'm curious to as why the direction colored stops were changed to one color, looks better but seems a bit more confusing?

Also what are you thoughts on using two drop pins (A & B) similar to Google Maps? I think since most people use Google Maps, it'd be easier to adapt to this.

hathix:
Overall, a good start! Some thoughts:

* I think the circle & pin design (for start/end stops) works in Google Maps because all the other UI elements for the route are lines. But in our app, each stop gets a circle as well. So when I looked at your screenshot, I was confused why there were 2 types of circles -- one type for stops and one type for my starting point. Having something like a pin and checkered flag for start and end, respectively, might make it more clear what's on your route and what isn't.
* Can we color the route between the chosen stops a different color than the rest of the route (or have some kind of pattern or other marking)? That way there's clearer visual feedback that the user has changed the stops.
* Could we use arrows or some other way to differentiate the 2 directions of stops? Otherwise people might get confused about why there are 2 parallel looking "routes."

exxonvaldez:
Thanks @akgupta89 and @hathix, good feedback.  Here's what'll I work on:

- Replace circles at stops with a vector graphics triangle that points in the direction of travel.
- Color the outer part of the line between selected stops (currently white) with a darker gray to show the selected range without obscuring the speed colors.
- See if better Material UI icons are available to indicate start/end stops.

exxonvaldez:
Screen Shot 2019-10-20 at 12 28 57 AM


1.  I've put a "v" inside the circles showing the direction of travel.  This also helps a lot when seeing stops in all directions. It makes overlapping stops and stops nearly on top of each other a little easier to work with.  The only way to separate overlapping stops is to zoom in, but at least you know which one is on top:

Screen Shot 2019-10-20 at 12 31 37 AM

2.  Note that the selected range of stops is highlighted in the first screenshot.

3.  Start stop is now a pin icon (instead of a circle) and the end stop is a flag icon.  And the start/end stop names are shown, to be more like Google Maps.  To make the icons easier to read, they are drawn on top of white circles (see first screenshot again).

On the coding side, I discovered that I had to use fully custom svg markers instead of the built in CircleMarker (the old circles).  This is because any custom stuff (icons, svg graphics, html) is always on top, with the Markers below that.  So overlapping stops came out looking illegible because you'd see overlapping icons instead of what is desired, one stop's circle overlapping another stop's circle with only the top stop's icon visible.

Going down the road of custom markers is kind of ugly compared to the simplicity of the built in CircleMarker, but I think it's our best option.  Writing our own svg is also kind of ugly but better than the Material UI icons, which are a single color and transparent and a fixed set of sizes, so you cannot specify background colors/shapes for them.

exxonvaldez:
> I'm curious to as why the direction colored stops were changed to one color, looks better but seems a bit more confusing?

So different colors for stops in different direction was hard to use, as you couldn't tell which way the red stops were going vs the blue stops.  Plus it would have meant adding a legend for stop colors, which I didn't want.  Lastly with the speeds being shown, there were too many colors going on at once.

> Also what are you thoughts on using two drop pins (A & B) similar to Google Maps? I think since most people use Google Maps, it'd be easier to adapt to this.

What I've been doing with the two icons and all the changes in general is based on the Google Maps web interface. I don't see A and B pins.

exxonvaldez:
Screen Shot 2019-10-20 at 2 39 50 PM

Agreed that the black was clashing.  Switched black icons and text to indigo.  I also switched the start icon to a "DirectionsTransit" icon.  This is used by Google Maps, but not directly on the map, only in the step by step directions.  It's a subway train, but they use it generically for transit  vs walking or driving.  Google Maps additionally has icons for buses vs subways vs historic streetcars, BART stations, etc.  I'm less sure about how the DirectionsTransit icon looks on the map, but I like it on the control panel.

PR #332, opened 2019-10-10
Beautify Round #8 (final)

### LINKED ISSUE #193: Run a giant round of beautification
Currently a lot of our PRs have a lot of unrelated code changes (that are just spacing/formatting) mixed in with the "real" changes. I think this is because some people have beautification hooks or their IDE does it automatically.

Can we have someone just run the beautifier on every file in the app, so that no future PRs get this problem?

akgupta89:
Fixes #193 

Linting now passes completely

## Proposed changes
 - Resolves shadowing 
 - Removes unused vars
 - Fixed directly modifying arguments


frontend/src/components/MareyChart.jsx
```diff
@@ -106,10 +110,11 @@ function MareyChart(props) {
       startTime,
       startHourOfDay,
     ) => {
+      const myTripData = tripData;
       const tripId = arrival.i;
       const vehicleId = arrival.v;
-      if (tripData.byTripId[tripId] === undefined) {
-        tripData.byTripId[tripId] = {
+      if (myTripData.byTripId[tripId] === undefined) {
```
hathix: What's the difference between `myVariable` and `thisVariable`? I see you using both throughout this PR.

frontend/src/components/QuadrantChart.jsx
```diff
@@ -55,7 +55,7 @@ function QuadrantChart(props) {
         className="custom-marking"
         customComponent={(row, positionInPixels) => {
           return (
-            <g className="inner-inner-component">
+            <g className="inner-inner-component" dummy-var={positionInPixels}>
```
hathix: Why's the dummy var needed? Add an HTML comment.

frontend/src/components/SidebarButton.jsx
```diff
@@ -7,85 +7,76 @@ import ChevronLeftIcon from '@material-ui/icons/ChevronLeft';
...
+  function toggleDrawer() {
+    setDrawer(!drawerOpen);
   }
 
-  handleToggleDrawer() {
-    this.setState(prevState => ({ drawerOpen: !prevState.drawerOpen }));
-  }
+  const activeStyle = {
+    fontWeight: 'bold',
+    color: '#3f51b5',
```
hathix: Can we pull this out into a constant?

hathix: I know this was already there before your PR, but might as well get it fixed while our attention is on it

frontend/src/components/QuadrantChart.jsx
akgupta89: positionInPixels needs read access to pass the lint, and a comment wouldn't count. I didn't want to remove the parameter because it sounded important that may be necessary in the future.

frontend/src/components/MareyChart.jsx
akgupta89: I was planning on migrating everything to `myX` instead of `thisX` to make things more readable, but I guess I missed some parts.

frontend/src/components/QuadrantChart.jsx
exxonvaldez: The parameter is not that important and should be removed.  It's part of the react-vis api, so it's well documented.

frontend/src/components/MapSpider.jsx
```diff
@@ -374,12 +381,14 @@ class MapSpider extends Component {
    * Append info about the downstream stops to the given stop object for plotting on the map.
    */
   addDownstreamStops(myStop) {
+    const thisMyStop = myStop;
```
exxonvaldez: Why is the addition of `thisMyStop` needed to pass linting?  Seems like doesn't really do anything useful from a code practices standpoint.  Is it because myStop is being modified?  If so, this assignment kind of just obscures that fact.

The name `thisMyStop` is hard to read.  Can the incoming parameter be something like `stop`, `stopObj`, or `targetStop`?

frontend/src/components/InfoTripSummary.jsx
```diff
@@ -35,32 +35,32 @@ export default function InfoTripSummary(props) {
   const waitTimes = graphData ? graphData.waitTimes : null;
   const tripTimes = graphData ? graphData.tripTimes : null;
 
-  const computeDistance = (graphParams, routes) => {
+  const computeDistance = (myGraphParams, myRoutes) => {
```
exxonvaldez: This seems to make the code harder to read.  If it's necessary to avoid shadowing, what if we just got rid of the parameters as in:

`const computeDistance = () => { `

and then let this code refer to the `graphParams` and `routes` objects in the enclosing scope?  The point of using the same variable names in the original code is to make it clear that it's the same object.

akgupta89: So the reason I didn't do that was because I figure really soon we are going to pull out all the calculation functions and hopefully to create a separate library that anyone can use, or at the very least move these functions outside of the view.

frontend/src/components/MapSpider.jsx
akgupta89: I agree, I thought about disabling this rule but I think in _practice_ the idea is good, theoretically any modification to a variable should result in a new variable. The best practice for coding is even if the variable name becomes quite longer, it should be a more descriptive name e.g `myStopDownstreamAtEndPoint` it helps with readability and developers to just read the variable name without having to know exactly what the function does or try to decipher it.

And by best practice functions shouldn't mutate arguments, they should return new ones. This was the quickest way to get around that without having to alter the code directly.

exxonvaldez: Okay, so let's get a more descriptive name than `thisMyStop`.  How about `stopToAugment`  or `targetStop`.

Issue #303, opened 2019-09-12
Show daily performance in hourly buckets


hathix:
Currently we only show performance in 3-4 hour buckets, like morning/afternoon/evening. Make each hour its own bucket so it's more granular.

PR #334, opened 2019-10-13
Convert backend API to GraphQL

youngj:
This is a continuation of https://github.com/trynmaps/metrics-mvp/pull/177 in a new branch (since the other branch is in jtanquil's personal repo). Most of the new changes are in the commit https://github.com/trynmaps/metrics-mvp/commit/ff44ff1d675e6104d517c28d818c3d3797f94c63 .

This PR updates the frontend to use the GraphQL API and removes the old API endpoints /api/metrics and /api/metrics_by_interval.

The frontend now requests the data that was previously returned by these two endpoints in one GraphQL query, without requesting data it doesn't need, using the query below:

```graphql
query($routeId:String, $startStopId:String, $endStopId:String,
    $directionId:String, $date:String, $startTime:String, $endTime:String) {
  routeMetrics(routeId:$routeId) {
    trip(startStopId:$startStopId, endStopId:$endStopId, directionId:$directionId) {
      interval(dates:[$date], startTime:$startTime, endTime:$endTime) {
        headways {
          count median max
          percentiles(percentiles:[90]) { percentile value }
          histogram { binStart binEnd count }
        }
        tripTimes {
          count median avg max
          percentiles(percentiles:[90]) { percentile value }
          histogram { binStart binEnd count }
        }
        waitTimes {
          median max
          percentiles(percentiles:[90]) { percentile value }
          histogram { binStart binEnd count }
        }
      }
      timeRanges(dates:[$date]) {
        startTime endTime
        waitTimes {
          percentiles(percentiles:[50,90]) { percentile value }
        }
        tripTimes {
          percentiles(percentiles:[50,90]) { percentile value }
        }
      }
    }
  }
}
```

### Changes to the GraphQL API

* URL changed to /api/graphql
* Query `routes` field simplified to return list of RouteInfo objects (instead of a RouteList object which only contains a list of RouteInfo objects)
* Query `routeInfo` field renamed to `routeConfig` to match terminology in nextbus.py, and returns RouteConfig object  ("config" meaning that it includes the stops/directions)
* Added `stopInfo` and `directionInfo` fields to RouteConfig object to allow clients to look up information about an individual stop or direction given its ID
* RouteInfo object has a `config` field which returns a RouteConfig object (so clients that only need id/name of routes can request routes property without fetching config)
* StopDictionary removed so that RouteInfo stops field is just a list of StopInfo objects (instead of a list of key-value pairs where the value is a StopInfo object -- the client could always make a dictionary from the returned data if needed)
* `stops` field of DirectionInfo object renamed to `stopIds` to indicate that they are strings and not StopInfo objects
* ArrayStats object renamed to BasicStats, and resolver methods now compute stats like avg, min, median, etc. only if requested (refactored logic from metrics.py)
* BasicStats `percentiles` and `histogram` fields have parameters that allow the client to customize the percentiles returned and the histogram bins
* String `value` field on HistogramBin object removed. Clients can use `binStart` and `binEnd` instead.
* routeMetrics field in top level query now only takes `routeId` parameter (no date range or stop related parameters), so that it corresponds to a metrics.RouteMetrics object that can be reused by multiple calculations for different stops/date ranges in order to improve efficiency
* RouteMetrics object contains a `trip` field with parameters `startStopId`, `endStopId`, `directionId` which returns a TripMetrics object. Date range parameters are not included here in order to make it easier for the client to compute stats over multiple date ranges by requesting multiple fields of the TripMetrics object
* `stopinfo` field removed from RouteMetrics object since the client can request stop information via the `routeConfig` field if necessary
* TripMetrics object contains an `interval` field with parameters `dates`, `startTime`, and `endTime` which returns an IntervalMetrics object. Having a `dates` field (array of date strings) will make it easier to support date ranges which may not include every date between a start and end date (although perhaps may become unwieldy if the client needs to request hundreds of dates)
* TripMetrics object contains a `timeRanges` field with parameter `dates` which returns a list of IntervalMetrics objects. This makes it easier for the client to request data from all of the standard time ranges within a day (early morning, AM peak, midday, PM peak, late evening). The timeRanges field doesn't allow the client to customize the time ranges, although clients can alternatively request multiple custom time ranges by requesting the `interval` field multiple times with different aliases.


hathix:
This might just be me, but I get this error when I start the app and try to visit `http://localhost:5000/api/graphql`:

Screen Shot 2019-10-13 at 8 43 49 PM

youngj:
@hathix - Probably need to run `docker-compose build` then `docker-compose up`

youngj:
GraphQL queries can be invoked via either GET or POST requests. For GET requests, the query in a parameter named `query` and optional variables in a JSON encoded parameter named `variables` (https://graphql.org/learn/serving-over-http/).  An example of calling the API via GET is in src/actions/index.js .

It would be available to use by third-party apps at https://muni.opentransit.city/api/graphql . 

Spacing in the GraphQL query doesn't matter.  

I'll add an explanation and example to metrics_api.md .

youngj:
Added documentation and example to https://github.com/trynmaps/metrics-mvp/blob/graphene/metrics_api.md . 

Note you will need to replace https://muni.opentransit.city with http://localhost:5000 to test the example URL before the PR is merged.

exxonvaldez:
Looks nice, seems to clean up the front end code quite a bit also.

There is an issue to make the intervals of the day chart into an hourly chart (#303).  Presumably the intervals would start at 03:00 and end at 03:00+1.  I'm not clear on how we'd go about doing that, as the timeRanges parameter only has a date argument?  Would we just change the default intervals in constants.py to be hourly?

youngj:
To implement different time ranges, one approach would be to an optional parameter to the `timeRanges` field that lets the client specify time ranges (could define a custom GraphQL input type that has a startTime and endTime string, and the parameter to `timeRanges` could be an array of those).

Alternatively, the client could request the `interval` field multiple times with different aliases like this:
```graphql
...
h3: interval(dates:[$date], startTime:"03:00", endTime:"04:00") { ...intervalFields }
h4: interval(dates:[$date], startTime:"04:00", endTime:"05:00") { ...intervalFields }
h5: interval(dates:[$date], startTime:"05:00", endTime:"06:00") { ...intervalFields }
...
```

youngj:
I'd like to merge this, does anyone want any changes before approving this?

PR #347, opened 2019-10-21
Adding Bread Crumb feature

### LINKED ISSUE #160: Make breadcrumbs clickable
In this case, "Muni", "18-46th Avenue", etc should be clickable and take you to the appropriate view. 

Screenshot 2019-07-10 at 7 11 16 PM

ckrajewski:
Fixes #160.

## Proposed changes

Adding breadcrumbs

We need to think of a better color though (hard to see link)

## Screenshot

image

## Link to demo, if any

...


akgupta89:
Links should be white IMO

frontend/src/screens/RouteScreen.jsx
```diff
@@ -44,11 +75,11 @@ function RouteScreen(props) {
       : null;
   const startStopInfo =
     direction && graphParams.startStopId
-      ? selectedRoute.stops[graphParams.startStopId]
+      ? Object.assign(selectedRoute.stops[graphParams.startStopId], {id:graphParams.startStopId})
```
exxonvaldez: This is modifying an element of `stops`, which might be confusing since these elements normally don't have ids, and this code would modify some but not others.  Since the breadcrumbs just need an id and a title, how about `startStopInfo = direction && graphParams.startStopId
?  { id: graphParams.startStopId, title: selectedRoute.stops[graphsParams.startStopId].title }
: null`

frontend/src/screens/RouteScreen.jsx
```diff
@@ -32,6 +35,34 @@ function RouteScreen(props) {
...
   }, [routes, myFetchRoutes]); // like componentDidMount, this runs only on first render
 
+  const breadCrumbs = paths => {
+    let link = {
+      type:'ROUTESCREEN'
+    }
+    const params = [ROUTE_ID, DIRECTION_ID, START_STOP_ID, END_STOP_ID];
+    const labels = (param, title) => {
+        let  specialLabels = {};
+        specialLabels[START_STOP_ID] = `(from ${title}`;
```
exxonvaldez: Let's remove the parentheses from the crumb labels now that there is a `>`  separator between them.

frontend/src/screens/RouteScreen.jsx
```diff
@@ -57,12 +88,8 @@ function RouteScreen(props) {
         <Toolbar>
           <SidebarButton />
           <div className="page-title">
-            {agencyTitle}
-            {selectedRoute ? ` > ${selectedRoute.title}` : null}
-            {direction ? ` > ${direction.title}` : null}
-            &nbsp;
-            {startStopInfo ? `(from ${startStopInfo.title}` : null}
-            {endStopInfo ? ` to ${endStopInfo.title})` : null}
+            <Link to="/">{agencyTitle}</Link>
```
exxonvaldez: I agree with Arun that the links should be styled white.

frontend/src/screens/RouteScreen.jsx
```diff
@@ -32,6 +35,34 @@ function RouteScreen(props) {
...
+    const params = [ROUTE_ID, DIRECTION_ID, START_STOP_ID, END_STOP_ID];
+    const labels = (param, title) => {
+        let  specialLabels = {};
+        specialLabels[START_STOP_ID] = `(from ${title}`;
+        specialLabels[END_STOP_ID] = `to ${title})`;
+        return specialLabels[param] ? specialLabels[param] : title;
+    }
+    return paths.filter(path => {
+      return  path ?  true : false;
+      }).map((path, index, paths) => {
```
exxonvaldez: I'd find this easier to read if the `.filter` and `.map` were separate statements.  Also for `nextValue`, since the paths are filtered, `paths[index+1]` will sometimes be an out of bounds index.  How replacing `nextValue` with `hasNextValue` where `hasNextValue = index+1 < paths.length` ?

hathix:
I agree we should use the Material Breadcrumb component if possible: https://material-ui.com/components/breadcrumbs/

ckrajewski:
tried looking into Material UI breadcrumbs, but it didn't seem to offer us much

frontend/src/screens/RouteScreen.jsx
```diff
@@ -51,18 +91,20 @@ function RouteScreen(props) {
...
-            {direction ? ` > ${direction.title}` : null}
-            &nbsp;
-            {startStopInfo ? `(from ${startStopInfo.title}` : null}
-            {endStopInfo ? ` to ${endStopInfo.title})` : null}
+        
+            <Link to="/" className={classes.whiteLinks}>{agencyTitle}</Link>
+
+            {breadCrumbs([selectedRoute,direction,
+              startStopInfo ? Object.assign(startStopInfo,{id: graphParams.startStopId }) : null,
+              endStopInfo ? Object.assign(endStopInfo,{id: graphParams.endStopInfo }) : null],classes.whiteLinks)}
```
ckrajewski: @exxonvaldez 
Moved Object.assign to when I call the method itself so as to not affect the original objects

akgupta89:
> tried looking into Material UI breadcrumbs, but it didn't seem to offer us much

Could you explain this comment a bit more?

ckrajewski:
@akgupta89 
Material UI was just a  UI shell; our core logic is still the same
And that UI shell involves having a pillbox around the link, plus using their link component
Ours uses redux route linking
https://material-ui.com/components/breadcrumbs/

ckrajewski:
@hathix @exxonvaldez 
image

ckrajewski:
Tried to improve styling
Used Breadcrumb, Typography, and NavigateNextIcon from Material UI

hathix:
Ready for us to review, @ckrajewski ?

exxonvaldez:
Styling is a nice improvement over original.  I also like the use of redux-first-router `Link`.  I did notice that the resulting hrefs work, but there is an issue where the hrefs break the UI.  Because they don't have the optional parts of the url pattern like "/fromStop/" in them, `buildPath` in routeUtil fails to parse the url correctly.  So trying to use the stop selectors doesn't work after clicking a breadcrumb link.

As @young noted in #408 , `buildPath` is brittle.  redux-first-router provides its own API do the equivalent, called `actionToPath` https://github.com/faceyspacey/redux-first-router/blob/master/docs/low-level-api.md .  Unfortunately the docs don't really explain how to use `actionToPath` but I found how `Link` uses it.  It's this utility method `toUrl` at https://github.com/faceyspacey/redux-first-router-link/blob/master/src/toUrl.js which is called by https://github.com/faceyspacey/redux-first-router-link/blob/master/src/Link.js .

I'd been thinking for a while that there had to be redux-first-router way to generate url paths and now we know what it is.  I think the best thing to do with this PR is add in the changes needed for #408 before merging, so that the issue I noted is fixed and this PR doesn't introduce any new defects into production.

exxonvaldez:
(Sorry it took 11 days to re-review this, my brain was out of action due to a cold.)

ckrajewski:
@exxonvaldez 
feel free to take another look :)
image

frontend/src/routeConstants.js
```diff
@@ -0,0 +1,14 @@
...
+export const TO_STOP = 'toStop';
+
+//route map constants
+export const ROUTESCREEN = 'ROUTESCREEN';
+
+//path param Id constants
+export const ROUTE_ID = 'routeId';
+export const DIRECTION_ID = 'directionId';
+export const START_STOP_ID = 'startStopId';
+export const END_STOP_ID = 'endStopId';
```
exxonvaldez: I'm not sure these constants provide much value or discoverability in the code.  Everywhere that `foo[ROUTE_ID]` is used, seems like it could just be `foo.routeId`.  And the value of the constants have to match the regexp and code in mapRoutes.js, which does not use these constants.

ckrajewski: @exxonvaldez 
I've updated the PR to be able to merge with master
I know this PR has been open for a while, so feel free to tackle some of the issues you were talking about after merge
Right now it's working, but I can't figure out how to include the optional param names (i.e. direction) in the URL
It looks like this
image
Notice how there's no direction or fromStop in the URL
Just the params

ckrajewski: But I think this code is less awkward than the current build path code
Do you think we can merge this, and then you can open up and take up the issues you were talking about? (i.e. null path issues)
That way, I won't block you guys.
Especially since I won't be there  this Wednesday (1/8)

hathix:
Terence talked to Jesse and Chris about this. Jesse's done a lot of work on this and implemented some of the backend stuff. We do want to keep Chris's styling though.

hathix:
I like Chris's styling. We need to fix the merge conflicts and other problems caused by the recent other updates (i.e. apply Chris's styling to tip of tree, more or less)

PR #348, opened 2019-10-22
Added error checking for invalid dates, per https://github.com/trynma…

### LINKED ISSUE #336: Picking a future date generates a non-user-friendly error
Screenshot 2019-10-16 20 12 02

Selecting a future date in the Isochrone generates a browser alert with the following message:
error loading wait times: Invalid JSON

SeannyPhoenix:
…ps/metrics-mvp/issues/336

Fixes #336.

### DIFF
--- a/frontend/public/isochrone-worker.js
+++ b/frontend/public/isochrone-worker.js
@@ -44,7 +44,7 @@ function loadJson(url)
             }
             catch (e)
             {
-                reject({message: "Invalid JSON", status: req.status});
+                return reject({message: "Invalid JSON", status: req.status});
             }
             resolve(res);
         });
--- a/frontend/public/isochrone-worker.js
+++ b/frontend/public/isochrone-worker.js
@@ -44,11 +44,11 @@ function loadJson(url)
             }
             catch (e)
             {
-                return reject("Invalid JSON");
+                reject({message: "Invalid JSON", status: req.status});
             }
             resolve(res);
         });
-        req.onerror = () => reject(req.statusText);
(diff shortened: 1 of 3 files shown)


frontend/src/components/DateTimePanel.jsx
```diff
@@ -68,6 +68,7 @@ function DateTimePanel(props) {
   const { graphParams } = props;
   const classes = useStyles();
   const [anchorEl, setAnchorEl] = React.useState(null);
+  const maxDate = initialState.graphParams.date;
```
exxonvaldez: The initial date is currently set to be yesterday's date.  However, it is possible to get data for the current day (today) since the job to generate data files runs hourly.  We start out with yesterday because it's assured to be a full day's worth of data.  You could use Moment here to generate today's date.

SeannyPhoenix: Now using a Moment of today's date.

exxonvaldez:
Sean, feel free to merge this in yourself if you want (and have permissions to), or you can wait for more feedback.

frontend/public/isochrone-worker.js
```diff
@@ -44,11 +44,11 @@ function loadJson(url)
             }
             catch (e)
             {
-                return reject("Invalid JSON");
+                reject({message: "Invalid JSON", status: req.status});
```
youngj: The `return` statement should be here so that it doesn't fall through to the `resolve` below, in which case both `reject` and `resolve` will be called.

SeannyPhoenix: Added it back.

PR #349, opened 2019-10-23
Remove Paper around route table and style links.

### LINKED ISSUE #339: Add padding to routes table
Currently the names of the routes are right up against the edge of the routes table. I think some better, more Material-y formatting would help to introduce padding and make them look less like plain ol' blue links.

image

exxonvaldez:
Fixes #339.

## Proposed changes

Remove `Paper` around route table, as this creates unnecessary borders and shadows around the table.  Style the route links using the theme's primary dark color and no underline.

## Screenshot

Screen Shot 2019-10-22 at 11 28 23 PM

## Link to demo, if any

n/a


hathix:
Could we keep the Paper component and try removing some of the other markup that might be interfering with the padding?

For instance, take this example: https://material-ui.com/components/tables/#SimpleTable.js. I think the padding would look nice, but somehow some CSS class is getting messed up when we do it -- perhaps by the table wrapper div?

PR #351, opened 2019-10-24
Button to apply current date and time as filter

### LINKED ISSUE #311: Show quick filter for current day/time
The scenario: a tourist is trying to figure out which bus to take home and needs to quickly know how good the various buses she could take are. She wants to quickly apply a filter for the current time of day and the current day of week to get an accurate view of the average quality of the buses nearby. She could do this manually in opentransit but it requires fiddling with settings, and she doesn't have much time.

The solution: a button in the date picker (or elsewhere in the header) that instantly applies a filter for the current time of day (e.g. 9-10pm) and day of week (e.g. Tuesday).

I literally had this exact problem in Pittsburgh just now- 5 buses around, Google maps had no idea which ones were nearby, so I had to guess which bus had the shortest wait.

wllgrnt:
Attempts to solve #311. Sets the current date to today, and the time to
the corresponding time range.

Fixes #311 

## Proposed changes

Adds a function setTimeAndDateToNow which sets the date, startTime and endTime props to the current local time/date.

Adds a Button which calls this function to the date/time picker.

...

## Screenshot

Button

...

## Link to demo, if any

...

### DIFF
--- a/frontend/src/components/DateTimePanel.jsx
+++ b/frontend/src/components/DateTimePanel.jsx
@@ -89,7 +89,7 @@ function DateTimePanel(props) {
   }
 
 
-  function setTimeToDateToNow() {
+  function setTimeAndDateToNow() {
     /** Sets the date as today
     * Gets the current time, sets start and end time as the
     * the corresponding members of TIME_RANGES
@@ -240,7 +240,7 @@ function DateTimePanel(props) {
         <br />
         <List style={{ color: 'black' }}>
           <ListItem>
-            <Button onClick={setTimeToDateToNow}>Set to current time</Button>
+            <Button onClick={setTimeAndDateToNow}>Set to current time</Button>
           </ListItem>
 
           <ListItem>
--- a/frontend/src/components/DateTimePanel.jsx
+++ b/frontend/src/components/DateTimePanel.jsx
@@ -14,6 +14,7 @@ import { List, ListItem } from '@material-ui/core';
 import TextField from '@material-ui/core/TextField';
 import IconButton from '@material-ui/core/IconButton';
 import CloseIcon from '@material-ui/icons/Close';
+import Moment from 'moment';
 import { TIME_RANGES, TIME_RANGE_ALL_DAY } from '../UIConstants';
 import { initialState } from '../reducers/routesReducer';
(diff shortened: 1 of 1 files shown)


wllgrnt:
Do people think the buttons need a colour so its more obvious they can be pressed?

exxonvaldez:
Agree that the button could use a color.  I think it's the `<List style={{ color: 'black' }}>` that is making everything black, otherwise the button would probably be blue by default.  Could try removing the 'black" and see what happens, else add a color to the new Button.

Also, the Reset button seems a little odd now, maybe it can be replaced by the new Button.  I think the new Button could just say 'Today'.  Reset goes to yesterday but I'm not sure how useful that is, especially since today is pretty close.

hathix:
I agree with the above sentiment. My summary:

* Get rid of the "reset" button
* Rename your new button to "Today"
* Move your new button to where the reset button is

I don't think you need color, but you can experiment with an "active" blue or pink color

wllgrnt:
Hi, changes now done. I went with the "secondary" colour - comparison screenshots attached

#### Secondary:
Secondary

#### Primary:
Primary

#### Default:
Default

exxonvaldez:
Having tested this branch out, I now realize that #311 is a little tricky to implement.  If you set the picker the current and date, what can happen is you get back little to no data (like it's 8pm now, and the 7pm-3am bucket for today has almost no vehicle runs in it).

From reading the issue more closely, I see that what would work better is using data that is **representative** of the current day and time.  That is, show data from previous Tuesdays for the last N months from 7pm-3am.  We don't have a UI for doing that kind of query yet (I am going to start on it) or the logic for building a list of days from the UI.

The quick and dirty solution would be to select 7 days ago and the appropriate time interval.  Not really sure how'd you label a button like that -- maybe just "Last Tues 7pm-3am", and later we can label it "Tuesdays 7pm-3am".

hathix:
Could we just have this send users to yesterday instead of today? Good point that there may not always be much data for today.

Or we could go back to the "reset" wording and have it send you back to last week, though it might confuse users.

hathix:
We're going to replace the reset & today buttons with one "Reset" button that shows the last 24 hours of data. #385 tracks that effort.

So don't merge this...

PR #355, opened 2019-10-24
Added Clear button and converted the toolbar to Material per #342

### LINKED ISSUE #342: Make the "clear" button on the isochrone more visible
The Isochrone's clear button is pretty easy to miss. Meanwhile, the dashboard has a nicely colored button in the bottom corner for clearing. We should use that kind of Material button on the isochrone for clearing. (And while you're at it, update the isochrone's left-side menu to be all Material-styled.)

SeannyPhoenix:
## Screenshot
Original:
#342 non-Material

After changes:
#342 Material

### DIFF
--- a/frontend/src/screens/Isochrone.jsx
+++ b/frontend/src/screens/Isochrone.jsx
@@ -642,7 +642,7 @@ class Isochrone extends React.Component {
               className="isochrone-controls"
               direction="column">
               <Grid item>
-                <Typography variant="subtitle1" color="primary">Statistic</Typography>
+                <Typography variant="subtitle1">Statistic</Typography>
               </Grid>
               <Grid item>
                 <Select
@@ -659,7 +659,7 @@ class Isochrone extends React.Component {
               className="isochrone-controls"
               direction="column">
               <Grid item>
-                <Typography variant="subtitle1" color="primary">Max Trip Time</Typography>
+                <Typography variant="subtitle1">Max Trip Time</Typography>
               </Grid>
               <Grid item>
                 <Select
@@ -676,10 +676,9 @@ class Isochrone extends React.Component {
             </Grid>
             <Grid container
               className="isochrone-controls"
-              direction="column"
-              alignItems="left">
+              direction="column">
               <Grid item>
-                <Typography variant="subtitle1" color="primary">Routes</Typography>
+                <Typography variant="subtitle1">Routes</Typography>
               </Grid>
               <Grid container item
                 direction="row"
--- a/frontend/src/screens/Isochrone.jsx
+++ b/frontend/src/screens/Isochrone.jsx
@@ -6,6 +6,12 @@ import L from 'leaflet';
 import Control from 'react-leaflet-control';
 import Grid from '@material-ui/core/Grid';
 import Typography from '@material-ui/core/Typography';
+import FormGroup from '@material-ui/core/FormGroup';
+import List from '@material-ui/core/List';
+import ListItem from '@material-ui/core/ListItem';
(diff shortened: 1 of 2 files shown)


SeannyPhoenix:
Updated.

image

frontend/src/screens/Isochrone.jsx
```diff
@@ -553,17 +564,17 @@ class Isochrone extends React.Component {
...
             onChange={this.handleToggleRoute}
             value={route.id}
-          />{' '}
-          {route.id}
-        </label>
-      </div>
+            color="primary"
+          />
+        </ListItemIcon>
+        <ListItemText id={route.id} primary={route.id}/>
```
exxonvaldez: Clicking on the checkbox labels doesn't toggle the checkbox.  There's a way to get a clickable Checkbox label, see https://material-ui.com/components/checkboxes/  for examples.

frontend/src/screens/Isochrone.jsx
```diff
@@ -625,49 +636,64 @@ class Isochrone extends React.Component {
...
+              </Grid>
+            </Grid>
+            <Grid container
+              className="isochrone-controls"
+              direction="column"
+              alignItems="left">
+              <Grid item>
+                <Typography variant="subtitle1" color="primary">Routes</Typography>
+              </Grid>
+              <Grid containter item
```
exxonvaldez: I'm seeing some warnings in the browser console about the attributes being passed into Grid -- container is misspelled here.

SeannyPhoenix:
I would like to break the scrolling problem into its own issue; this problem is actually inherited from before.

frontend/src/screens/Isochrone.jsx
```diff
@@ -625,49 +637,64 @@ class Isochrone extends React.Component {
...
-                  <span onClick={this.selectNoRoutesClicked}>none</span>
-                </div>
-                <div className="isochrone-routes">
+                </Select>
+              </Grid>
+            </Grid>
+            <Grid container
+              className="isochrone-controls"
+              direction="column"
+              alignItems="left">
```
exxonvaldez: I'm seeing a warning (always good to keep the developers tools open while developing) in the console:  index.js:1437 Warning: Failed prop type: Invalid prop `alignItems` of value `left` supplied to `ForwardRef(Grid)`, expected one of ["flex-start","center","flex-end","stretch","baseline"].

frontend/src/screens/Isochrone.jsx
```diff
@@ -625,49 +637,64 @@ class Isochrone extends React.Component {
...
-            <div className="isochrone-controls">
-              <div>
-                stat:
-                <select
+          <Control position="topleft" className="">
+            <Grid container
+              className="isochrone-controls"
+              direction="column">
+              <Grid item>
+                <Typography variant="subtitle1" color="primary">Statistic</Typography>
```
exxonvaldez: The primary color is meant for the primary action in a dialog or card.  So the primary color here throws me off, I keep expecting them to be clickable.  Can you keep these heading on the default color (no color attribute)?

exxonvaldez:
> I would like to break the scrolling problem into its own issue; this problem is actually inherited from before.

Sounds good.   Above the relevant pull request comment, you can click the `... -> reference in new issue` to create a new issue.  I'll do it this time.

exxonvaldez:
> Looks good to merge when you're ready! Thanks for making the fixes.

Oh, I see @akgupta89  asked for a multiple select.  Maybe we can create a new issue for that?  Feels like this PR should go in so it doesn't sit around too much longer.

PR #362, opened 2019-10-29
Hover Over Stops

### LINKED ISSUE #272: Highlight Stops on hovering over items on To/From Dropdown
It'd be nice if when you hover over entries on the To & From Dropdowns that the stop on the map highlights so you know exactly where the stop you're hovering on the dropdown is on the map.

zachzwy:
Fixes #272.

## Proposed changes
When you hover over entries in the To & From Dropdowns, the stop/route on the map will update correspondingly. So you know exactly where the stop you're hovering is on the map.

## Link to demo, if any
[https://open-transit.herokuapp.com/route/E/direction/E____O_F00/](https://open-transit.herokuapp.com/route/E/direction/E____O_F00/)

### DIFF
--- a/frontend/src/components/ControlPanel.jsx
+++ b/frontend/src/components/ControlPanel.jsx
@@ -1,5 +1,5 @@
 /* eslint-disable react/prop-types */
-import React from 'react';
+import React, {useState} from 'react';
 import { connect } from 'react-redux';
 // import PropTypes from 'prop-types';
 import { makeStyles } from '@material-ui/core/styles';
@@ -30,6 +30,7 @@ const useStyles = makeStyles(theme => ({
 function ControlPanel(props) {
   const { routes, graphParams } = props;
   let secondStopList = [];
+  const [allowHover, setAllowHover] = useState(false);
 
   /**
    * Sets the direction
@@ -101,13 +102,16 @@ function ControlPanel(props) {
   }
 
   function handleMouseOver(stopId) {
-    props.onGraphParams({
-      ...graphParams,
-      stopOnHoverId: stopId
-    });
+    if (allowHover) {
+      props.onGraphParams({
+        ...graphParams,
+        stopOnHoverId: stopId
+      });
+    }
   }
 
   function clearHover() {
+    setAllowHover(false);
     props.onGraphParams({
       ...graphParams,
       stopOnHoverId: null
@@ -208,6 +212,7 @@ function ControlPanel(props) {
                     value={graphParams.startStopId || 1}
                     onChange={onSelectFirstStop}
                     input={<Input name="stop" id="fromstop" />}
+                    onOpen={() => setAllowHover(true)}
                     onClose={() => clearHover()}
                   >
                     {(selectedDirection.stops || []).map(firstStopId => (
@@ -239,6 +244,7 @@ function ControlPanel(props) {
                     value={graphParams.endStopId || 1}
                     onChange={onSelectSecondStop}
                     input={<Input name="stop" id="tostop" />}
+                    onOpen={() => setAllowHover(true)}
                     onClose={() => clearHover()}
                   >
                     {(secondStopList || []).map(secondStopId => (
--- a/frontend/src/components/ControlPanel.jsx
+++ b/frontend/src/components/ControlPanel.jsx
@@ -195,7 +195,7 @@ function ControlPanel(props) {
                   value={graphParams.startStopId || 1}
                   onChange={e => {
                     onSelectFirstStop(e);
-                    setTimeout(() => clearStopOnHover(), 200);
+                    setTimeout(() => clearStopOnHover(), 400);
                   }}
                   input={<Input name="stop" id="fromstop" />}
                   onMouseOver={e => handleMouseOver(e.target.getAttribute("data-value"))}
@@ -224,7 +224,7 @@ function ControlPanel(props) {
                   value={graphParams.endStopId || 1}
                   onChange={e => {
                     onSelectSecondStop(e);
-                    setTimeout(() => clearStopOnHover(), 200);
+                    setTimeout(() => clearStopOnHover(), 400);
                   }}
                   input={<Input name="stop" id="tostop" />}
                   onMouseOver={e => handleMouseOver(e.target.getAttribute("data-value"))}
--- a/frontend/src/components/ControlPanel.jsx
+++ b/frontend/src/components/ControlPanel.jsx
@@ -68,7 +68,8 @@ function ControlPanel(props) {
     return secondStopInfo.stops.slice(secondStopListIndex + 1);
   }
 
-  function handleSelectFirstStop(stopId) {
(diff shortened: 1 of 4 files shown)


Dehorser:
Looks good. We could possibly add a debouncing feature if the user is simply scrolling through the dropdown, rather than deliberately hovering over it?

zachzwy:
Adding a debouncing feature is a great idea! I'll work on it.

zachzwy:
Addressed @exxonvaldez 's suggestions. I know that some pixels are off, but I want to hear more feedback.

hathix:
I like it! Terence?

exxonvaldez:
Looks good, but I see at least two issues:

1.  There is a regression where, after picking "from" and "to" stops, you then change the "from" stop using the dropdown to a stop beyond the "to" stop.  On the production site, this clears out the "to" stop (both on the map and in the control panel).  On this branch, the map continues to show the old "to" stop.  My hunch is that is caused by the use of setTimeout.

2.  In order to share state between components, we are using Redux (and Redux-Thunk).  That is, instead of using useState() and pushing state down to child components, we use action creators (in the actions directory) to set state and reducers (in the reducers directory) to apply the state changes.  It's important to stick with this pattern so that is is easy to find the shared state (as it is with graphParams).

frontend/src/components/MapStops.jsx
```diff
@@ -78,7 +78,34 @@ class MapStops extends Component {
...
+            Colors.INDIGO}" stroke-width="0.6" fill="none"/>` +
+          `</svg>` + 
+           // this is the stop title with a text shadow to outline it in white
+         
+         `<div style="position:relative; top:-29px; left:29px; font-weight:bold; color:` + Colors.INDIGO + `; ` +
+         `text-shadow: -1px 1px 0 #fff,` +
+         `1px 1px 0 #fff,` +
+         `1px -1px 0 #fff,` +
+         `-1px -1px 0 #fff;">${stop.title}</div>`,
+      });
```
exxonvaldez: Would still like to see more refactoring done here and less copy-pasting of existing code.

frontend/src/components/ControlPanel.jsx
```diff
@@ -190,13 +188,17 @@ function ControlPanel(props) {
...
-                <StartStopIcon fontSize="small" htmlColor={Colors.INDIGO}/>
+                <StartStopIcon fontSize="small" htmlColor={Colors.INDIGO} />
                 <FormControl className={classes.formControl}>
                   <InputLabel htmlFor="fromstop">From Stop</InputLabel>
                   <Select
                   value={graphParams.startStopId || 1}
-                  onChange={onSelectFirstStop}
+                  onChange={e => {
+                    onSelectFirstStop(e);
+                    setTimeout(() => clearStopOnHover(), 400);
```
exxonvaldez: Would be nice to avoid setTimeout and its sometimes complicated side effects if possible.  Is this mainly about the case where the dropdown gets closed?  If so, there should be an event for that (that can differentiated from a bona fide selection being made).

frontend/src/components/MapStops.jsx
zachzwy: Got it. Will try to refactor it.

frontend/src/components/ControlPanel.jsx
zachzwy: The setTimeout here is to handle the problem that onMouseover fires again after onChange is fired, which causes the problem that even after the user selects a stop, another stop will have the hover effect.

Is this the problem that you referred to when you said "the case where the dropdown gets closed"? setTimeout is definitely not the best way to handle it. What is the event that you have in mind?

exxonvaldez: > The setTimeout here is to handle the problem that onMouseover fires again after onChange is fired, which causes the problem that even after the user selects a stop, another stop will have the hover effect.
> 
> Is this the problem that you referred to when you said "the case where the dropdown gets closed"? setTimeout is definitely not the best way to handle it. What is the event that you have in mind?

Yeah, I think you understood my comment correctly... at the time I wrote it, it was just a gut feeling that setTimeout was for some edge case.

I have looked into more closely now, and it was actually pretty tough to figure out.  I was logging events on the `Select`.  Definitely better to log stop titles rather than stop IDs to understand what is going on.  The extra mouse over events you mentioned I'm guessing are from the `Select`'s closing animation, where the menu shrinks down.

So the cleaner fix I believe is to use `onClose` and `onOpen` handlers so that you can track when the `Select`s are open.  When a menu choice is made, the onClose will fire first, then the animation, and then the extra mouse over.  Probably can use local state in ControlPanel to keep track of this.

Oh, one more thing.  The mouse over events don't cover the keyboard scenarios where a user is using the keyboard to open the `Select` and arrow keys to highlight items (and then either making a choice or cancelling it).  I found that using an `onFocus` on the `MenuItem` was effective for tracking these highlight changes.  It starts getting complicated because the `Select` has both a keyboard focus and a separate mouse over highlight, but I think most users will only use of these devices at a time.

zachzwy:
1. Use the central store and reducer/action to control the hover effect;
2. Handle the keyboard select scenarios;
3. Refactor the code for drawing icon (still need more work).

frontend/src/components/MapStops.jsx
```diff
@@ -46,44 +46,52 @@ class MapStops extends Component {
   populateStop = (stop, IconType, currentPosition, rotation, onClickHandler, tooltip) => {
 
     let icon = null;
+    // this is the stop title with a text shadow to outline it in white
+    const stopTitle = (top, left) => `<div style="position:relative; top:${top}px; left:${left}px; font-weight:bold; color:` + Colors.INDIGO + `; ` +
+    `text-shadow: -1px 1px 0 #fff,` +
```
akgupta89: You don't need to break these up with pluses on a new line when using the template literal syntax

frontend/src/screens/RouteScreen.jsx
```diff
@@ -70,11 +70,11 @@ function RouteScreen(props) {
 
       <Grid container spacing={0}>
         <Grid item xs={12} sm={6}>
-          <MapStops routes={routes} />
```
akgupta89: More of a nitpick but I believe the space here is more proper, but I'm wondering why prettier didn't auto format it back in.

zachzwy:
Reviewed with @exxonvaldez tonight and want to take some notes:

**Issue:**
The current scene works perfectly with routes with few stops (eg. E-Embarcadero with 15 stops). However, the hover effect gets delayed when there are plenty of stops (eg. N-Judah with 34 stops). This is probably because React needs to re-render the route and all the stops when hovering over Select component.

**Propose change**
Try to locally update the icon without going through the full React rendering cycle.
- Give icon unique className when rendering, then select the "hover" DOM node and manipulate the CSS.
- Or have a special class controlling the hovering effect style, then apply the className to the "hover"DOM node.

PR #376, opened 2019-11-02
update backend to use new tryn-api

youngj:
Updates metrics-mvp to support tryn-api changes in https://github.com/trynmaps/tryn-api/pull/46 with timestamps in seconds, exclusive endTime parameter, and some different field names. Will need to also update the default tryn-api URL before merging.

### DIFF
--- a/backend/models/eclipses.py
+++ b/backend/models/eclipses.py
@@ -18,7 +18,6 @@ def produce_buses(route_state: dict) -> pd.DataFrame:
             .reindex(['RAW_TIME', 'VID', 'LAT', 'LON', 'DID', 'AGE'], axis='columns')
 
     # adjust each observation time for the number of seconds old the GPS location was when the observation was recorded
-    # and convert time from milliseconds since Unix epoch to seconds since Unix epoch
     buses['TIME'] = (buses['RAW_TIME'] - buses['AGE'].fillna(0)) #.astype(np.int64)
 
     buses = buses.drop(['RAW_TIME','AGE'], axis=1)
--- a/backend/models/eclipses.py
+++ b/backend/models/eclipses.py
@@ -19,7 +19,7 @@ def produce_buses(route_state: dict) -> pd.DataFrame:
 
     # adjust each observation time for the number of seconds old the GPS location was when the observation was recorded
     # and convert time from milliseconds since Unix epoch to seconds since Unix epoch
-    buses['TIME'] = (buses['RAW_TIME'].astype(np.int64) - buses['AGE'].fillna(0)) #.astype(np.int64)
+    buses['TIME'] = (buses['RAW_TIME'] - buses['AGE'].fillna(0)) #.astype(np.int64)
 
     buses = buses.drop(['RAW_TIME','AGE'], axis=1)
     buses = buses.sort_values('TIME', axis=0)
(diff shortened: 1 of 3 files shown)


backend/models/eclipses.py
```diff
@@ -7,21 +7,21 @@
...
                              'secsSinceReport': 'AGE',
-                             'routeStates.vtime': 'RAW_TIME_MS'}) \
-            .reindex(['RAW_TIME_MS', 'VID', 'LAT', 'LON', 'DID', 'AGE'], axis='columns')
+                             'states.timestamp': 'RAW_TIME'}) \
+            .reindex(['RAW_TIME', 'VID', 'LAT', 'LON', 'DID', 'AGE'], axis='columns')
 
     # adjust each observation time for the number of seconds old the GPS location was when the observation was recorded
     # and convert time from milliseconds since Unix epoch to seconds since Unix epoch
-    buses['TIME'] = (np.round(buses['RAW_TIME_MS'].astype(np.int64)/1000) - buses['AGE'].fillna(0)) #.astype(np.int64)
+    buses['TIME'] = (buses['RAW_TIME'].astype(np.int64) - buses['AGE'].fillna(0)) #.astype(np.int64)
```
EddyIonescu: With the API returning Int I'm guessing we wouldn't need this cast and the above comment (line 21)?

backend/models/trynapi.py
```diff
@@ -132,19 +127,19 @@ def get_cache_path(agency_id: str, d: date, start_time, end_time, route_id) -> s
     source_dir = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
     return os.path.join(source_dir, 'data', f"state_{agency_id}/{str(d)}/state_{agency_id}_{route_id}_{int(start_time)}_{int(end_time)}.json")
 
-def get_state_raw(agency, start_time_ms, end_time_ms, route_ids):
+def get_state_raw(agency, start_time, end_time, route_ids):
     tryn_agency = 'muni' if agency == 'sf-muni' else agency
 
-    params = f'trynState(agency: {json.dumps(tryn_agency)}, startTime: {json.dumps(str(int(start_time_ms)))}, endTime: {json.dumps(str(int(end_time_ms)))}, routes: {json.dumps(route_ids)})'
+    params = f'state(agency: {json.dumps(tryn_agency)}, startTime: {json.dumps(str(int(start_time)))}, endTime: {json.dumps(str(int(end_time)))}, routes: {json.dumps(route_ids)})'
```
EddyIonescu: Should be Int as well if that's changed in tryn-api.

backend/models/eclipses.py
youngj: That's correct, removed.

youngj:
Updated endpoints to use http://tryn-api by default (which works in production), and set `TRYNAPI_URL: http://34.67.128.233` in docker-compose.yml for development.

@EddyIonescu  could you configure a subdomain in DNS pointing to 34.67.128.233 ? (e.g. tryn-api.opentransit.city or whatever you think is best)

Alternatively, we could use that the subdomain as the default value if the TRYNAPI_URL environment variable is not set, and override it with http://tryn-api in production.

youngj:
Will merge this shortly so that I can test portland-sc (which requires the new tryn-api because it has a dash in the agency name). Could change the IP address to a DNS name in a subsequent PR after the DNS record is set up.

EddyIonescu:
@youngj added the subdomain for tryn-api. Let me know if there are any issues.

youngj:
@EddyIonescu thanks!

PR #409, opened 2019-11-11
Refactor agency-specific configuration; base route config on GTFS instead of Nextbus

youngj:
### Agency specific configuration

This PR allows supporting new transit agencies by adding a small yaml configuration file at `/backend/agencies/<agency-id>.yaml` , and allows changing the transit agency in the frontend by setting `OPENTRANSIT_AGENCY_IDS` environment variable to a comma delimited string of agency IDs. 

In development, this can be done via a docker-compose.override.yml file like this:

```
version: "3.7"
services:
  flask-dev:
    environment:
      OPENTRANSIT_AGENCY_IDS: portland-sc,trimet
```

All SF Muni-specific configuration for both frontend and backend is moved to /backend/agencies/muni.yaml . 

The most important configuration settings for each agency are:

`id` - must match the agency ID used by orion and tryn-api (e.g. "muni", not "sf-muni")
`provider` - can be set to `nextbus` for Nextbus-based agencies to augment the static GTFS data with data from the Nextbus API. 
`nextbus_agency_id` - for Nextbus agencies, the ID of the agency in Nextbus API (e.g. "sf-muni")
`timezone_id` - local time zone of agency, e.g. "America/Los_Angeles"
`gtfs_url` - public url where static GTFS ZIP file can be downloaded
`route_id_gtfs_field` - the column name in routes.txt in the static GTFS file where the route ID that is used by OpenTransit is stored. By default this is the column named `route_id`. However, some providers store an opaque route ID in this column that would not have meaning to riders and that is different from the route ID stored by orion and returned by tryn-api. These agencies generally return the route ID in the column named `route_short_name`. The route ID used by metrics-mvp must be the same as the same as the route ID used by orion and tryn-api (for Nextbus agencies, this is the same as the route ID Nextbus). There is an added edge case for Muni where some routes have a `route_short_name` containing a dash in the GTFS file (e.g. T-OWL), but Nextbus returns a route ID containing an underscore (e.g. T_OWL). For now, the code replace dashes with underscores in the route IDs for all agencies using the nextbus provider, although it may be possible that there are other Nextbus agencies that actually do have dashes in the Nextbus route IDs.

### Frontend configuration

Agency-specific configuration is not compiled directly into the React JavaScript code. Instead, the HTML page hosting the React code first loads a script tag from the backend at the URL /api/js_config , which sets the global variable `window.OpentransitConfig`. The `config` React module reads this variable and exposes the constants `Agencies` (array of agency objects) and `S3Bucket` (name of the S3 bucket where route configuration, arrival times, cached wait time stats, and cached trip time stats are stored). 

Currently the frontend only displays routes from the first agency in the Agencies array (for the first agency listed in `OPENTRANSIT_AGENCY_IDS`). However, a future version of the frontend could use this configuration to display statistics from multiple transit agencies in the same region.

To make it easy to pass arbitrary agency-specific data to javascript, the `js_properties` setting in the agency configuration file can be an object with arbitrary key/value pairs. Currently the frontend uses the keys `routeHeuristics`, `serviceArea`, `defaultDisabledRoutes`, `title`, `initialMapCenter`, and `initialMapZoom`. The `timezone_id` setting in the agency configuration file is also passed to the backend as `timezoneId`.

### Data storage

All data stored by the backend is now stored in a single S3 bucket, which can be configured via the environment variable `OPENTRANSIT_S3_BUCKET`. (Previously there were different hardcoded S3 buckets for arrival times, cached stats, and timetables.) Having a single S3 bucket makes it easy for developers to set up their own S3 bucket for storing data, instead of needing write permission on a centralized S3 bucket. (This is different from the S3 bucket used by orion and tryn-api for the raw vehicle location data.) This also would make it easier for other groups to set up their own independent instances of orion, tryn-api, and metrics-mvp using their own S3 buckets and Kubernetes clusters.

### API Changes

The backend GraphQL API fields on the root Query object now have a required `agencyId` parameter.

### Changes to directions and stops

Direction IDs and stop IDs are now based on the IDs in the static GTFS files. Previously, we used the direction ID returned from the Nextbus API, like `12___I_F00`. Now, the direction IDs are typically either the string '0' or the string '1'. (For Muni, '0' is always used as Outbound and '1' as Inbound, but other agencies do not necessarily categorize directions as Outbound or Inbound.) 

The stop IDs from GTFS files are often the same as the stop IDs from the Nextbus API, but not always. In particular, the same stop ID may now appear in both the outbound and inbound directions (typically only at the terminals).

Static GTFS files do not directly show the list of stops that are associated with a route/direction. Instead, the same route/direction combination can be associated with multiple scheduled trips at various times of day on various days of the week. Each trip may potentially have different stops associated with it, and may be associated with a different "shape".  In order to determine the list of stops that we use, we find the shape ID that appears the most often for that route/direction, and get the list of stops from one trip associated with that shape ID. 

For Muni this means that there are no longer any routes with more than 2 directions. However, some vehicles may not have been scheduled to travel the same sequence of stops that we display.

Due to the changes in direction IDs and stop IDs, as well as the change in S3 storage location, old arrival times and statistics would need to be recomputed.

### Timetables (not done yet)

The code for parsing GTFS timetables has not yet been refactored to support non-Nextbus agencies. However, the app currently does not use GTFS timetables so this should not break anything unless people manually run the command line scripts related to timetables. It will probably be easier to update that code in a separate PR while also changing the output from CSV to a JSON format that is easier to work with.


exxonvaldez:
> Wow this is a huge PR.

Yeah, it's huge.  Looks good generally for the front end changes.

>     * At the same time, I'm wondering how much more complicated would it be to show all branches rather than just picking the most common shape_id? Especially since some transit routes (especially TTC routes like the 52) have lots of branches. I'd be fine with just showing all different stop sequences, even if it does show a bunch of trippers/partial runs (I guess they could be sorted by most common by shape_id?)

Going by memory, for Muni there were a couple of routes with interesting branches that are handled in different ways in Nextbus.  The 38 has 2 inbound and 2 outbound directions.  It's nice to see each direction plotted on the map, and route summaries and stop to stop stats still work well.  And there was a light rail line (maybe the J) that had an oddball direction going to the yard at 23rd and 3rd.  Then there is the 9, which has multiple terminals but only 1 inbound and 1 outbound direction.  The stop list is union of all the stops, which makes plotting travel time difficult.  The solution here was to only look at the "common" stops along the line.

So I think showing all branches as directions gives the user more confidence since they can see the full route instead of a subset.  Using only one shape drops stops -- seems like it would be good to have the all the stops, as we on the 9, I guess Nextbus must merge the shapes into one franken-shape.

About stop ids being the same at terminals -- this could cause minor rendering problems for the UI on the maps, since we've been able to assume stop ids are unique in each direction.  Can you file an issue for this?

youngj:
Thanks for the feedback. I agree that it's important to be able to handle multiple branches.

Depending on the transit agency, each route/direction in the GTFS feed may be associated with a large number of different GTFS shapes. For example, TriMet's MAX Blue Line is associated with 60 shapes in one direction and 57 shapes in the other direction. Displaying all of the GTFS shapes as separate directions in the UI would be unwieldy, in addition to greatly increasing the amount of computation and data storage needed. Also, some shapes have the same start/end stops but have different stops in the middle, so showing the start and/or end stop as the direction title in the UI wouldn't allow users to distinguish between different directions.

For some agencies, there are many cases where one shape is associated with a list of stops that is a subsequence of the list of stops associated with another shape (and in some cases, two shapes may have exactly the same list of stops). We can reduce the number of directions significantly by ignoring shapes that are contained within other shapes. This mostly matches the previous behavior when directions where derived from the Nextbus API. In the case of the MAX Blue Line, this algorithm combines all 60 shapes in 1 direction and 57 shapes in another direction as expected. 

However, even merging common subsequences leaves some cases where multiple shapes are very similar but not exactly the same. This can happen when a vehicle starts/ends service in the middle of the normal route and has one stop at the beginning/end which is not contained in the normal route (maybe immediately around the corner from the normal stop). I don't think it would be meaningful to the user to show all of these cases as separate directions.

Another issue to consider is what to use for OpenTransit's "direction ID" for routes with multiple branches, considering how GTFS feeds change over time. It is necessary for OpenTransit's direction ID to be consistent over time, since it is stored within the computed JSON files in S3, in order for OpenTransit to interpret historical data associated with different versions of a GTFS feed. In different versions of Muni's GTFS feed, for example, the shape IDs are not consistent, so using the GTFS shape ID as part of OpenTransit's direction ID would cause the direction ID to change over time. Similarly, the list of stops may change slightly in different versions of the GTFS feed, so having a direction ID that is derived from the list of stops would also be prone to changing over time.  

Due to the above issues, it doesn't seem practical to automatically extract branches from GTFS feeds. However, it does seem practical for OpenTransit developers to manually define branches in the YAML configuration for each agency, for the small fraction of routes where we don't want to just use the most common shape ID.

As an example of how developers could manually define branches, I added a custom_directions property to muni.yaml and started defining a few branches, copied below:

```
custom_directions:
  '38':
    - id: "1-48th"
      title: "Inbound from 48th Ave & Point Lobos Ave to Transit Center"
      gtfs_direction_id: "1"
      included_stop_ids: ["13608"]
    - id: "1-VA"
      title: "Inbound from V.A. Hospital to Transit Center"
      gtfs_direction_id: "1"
      included_stop_ids: ["15511"]
    - id: "0-32nd"
      gtfs_direction_id: "0"
      included_stop_ids: ["14275"]
    - id: "0-48th"
      gtfs_direction_id: "0"
      included_stop_ids: ["13608"]
      excluded_stop_ids: ["15511"]
    - id: "0-VA"
      title: "Outbound to 48th Ave & Point Lobos Ave via V.A. Hospital"
      gtfs_direction_id: "0"
      included_stop_ids: ["15511","13608"]
  '47':
    - id: "1"
      gtfs_direction_id: "1"
      excluded_stop_ids: ["14967"]
    - id: "1-3rd"
      title: "Inbound to Powell St & Beach St via 3rd St"
      gtfs_direction_id: "1"
      included_stop_ids: ["14967"]
    - id: "0"
      gtfs_direction_id: "0"
      included_stop_ids: ["13723"]
  '81X':
    - id: "1"
      gtfs_direction_id: "1"
      included_stop_ids: ["16694"]
  '83X':
    - id: "0"
      gtfs_direction_id: "0"
      included_stop_ids: ["17776"]
    - id: "1-18010"
      title: "Inbound to 9th St & Market St via Brannan St & 8th St"
      gtfs_direction_id: "1"
      included_stop_ids: ["18010"]
    - id: "1-17683"
      title: "Inbound to 9th St & Market St via Townsend St & 8th St"
      gtfs_direction_id: "1"
      included_stop_ids: ["17683"]
```

Since stop IDs are generally consistent across different versions of the GTFS file, it seems practical to define directions based on which stop IDs do or don't appear within that direction. However, if one of the included_stop_ids stops appearing in the route, or if the included_stop_ids/excluded_stop_ids no longer match exactly one shape (after merging shapes with common subsequences), there would be an error when running save_routes.py.

This approach also allows defining custom titles to allow users to distinguish directions in cases where multiple branches have the same end stop.

## Normalizing Muni stop IDs

As part of this change, I also updated Muni to use the GTFS "stop_code" as OpenTransit's stop ID instead of "stop_id". The stop_code is the same as what appears on the shelters near bus stations and  what is listed as the "stop ID" on Muni's website (e.g. https://www.sfmta.com/routes/n-judah). In Muni's GTFS feed, there are several cases where multiple stops have different stop_id values (and slightly different coordinates) but the same stop_code, e.g:

```
stop_lat,stop_code,stop_lon,stop_url,stop_id,stop_desc,stop_name,location_type,zone_id
37.771491,17684,-122.405042,,7684_merged_9175738,,Brannan St & 8th ST,0,
37.771322,17684,-122.405237,,7684_merged_9175728,,Brannan St & 8th ST,0,
```

Instead of having stop IDs 7684_merged_9175738 and 7684_merged_9175728, there is just one stop ID 17684.

Using the same stop ID for these stops is helpful for merging subsequences. Without this change, there were some cases where the same Muni route had two shapes with sequences of stops that were identical except with the _merged_xxxxxx suffix of the GTFS stop_id.

Currently it isn't important that OpenTransit's stop ID matches Nextbus stop ID (although we can still get this new stop_code from Nextbus API if needed). The GTFS stop_code is the same as the "stopId" returned by Nextbus's routeConfig API. The GTFS stop_id is usually (but not always) the same as the Nextbus stop "tag" (yes, it's confusing). It's more important that the stop IDs don't change over time, since they are stored in all the computed JSON files in S3, and changing stop IDs requires recomputing all historical data (which we will need to do anyway once this PR is merged).

However, stop_code is an optional GTFS field, so we can't rely on using it for all agencies. To handle this for other agencies, save_routes.py will use the GTFS stop_id by default unless overridden by the stop_id_gtfs_field property in the agency config file.

backend/get_state.py
```diff
@@ -9,23 +9,23 @@
...
+        raise Exception("Must specify --agency with --route")
 
     date_str = args.date
 
     d = util.parse_date(date_str)
 
+    # todo use default_day_start_hour
     start_time_str = args.start_time
     if start_time_str is None:
         start_time_str = '03:00'
```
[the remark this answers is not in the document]
youngj: If a trip crosses the 3am boundary, only the vehicle positions before 3am will be used to compute trips for the day when it started. The vehicle positions after 3am will be associated with separate trips on the next date. It could be something to work on once we support agencies with 24 hour routes.

youngj:
> in the GTFS the branch is part of the trip_headsign in `trips.txt', which is pretty much the destination sign.

Ignoring trip_headsign now since the format varies by transit agency, and instead doing "To <last_stop_title>" or "Inbound/Outbound to <last_stop_title>" (for Muni) by default unless overridden via custom_directions.

> Just to make sure I didn't miss it looking through the code - which files/modules make use of GTFS data?

save_routes.py is the only file that uses data from the GTFS feed directly. (Ignore gtfs_scraper.py, models/gtfs.py, and models/timetable.py for now - the logic in those files for extracting timetable data from GTFS feeds will need to be updated in a subsequent PR.)

Many different files on both the backend and frontend use the data derived from GTFS that was saved by save_routes.py. On the backend this typically looks like calls to `agency.get_route_list()` and `agency.get_route_config(route_id)`.

youngj:
Does someone have time to review the updated PR today? I'm hoping to be able to merge this in before Code for PDX tomorrow evening so that people could start working on projects that depend on this.

docs/ADVANCED_DEV.md
```diff
@@ -13,7 +13,7 @@ If the arrival times for a particular route/day haven't been computed yet, you'l
 To get arrival times for one or more routes/days that haven't been precomputed yet, run `compute_arrivals.py`
```
exxonvaldez: I think there references to other S3 bucket names in this doc, can you update them?  It's hard for me to tell if everything is one S3 bucket now or if there are any exceptions.  And a reference to the S3 bucket env variable in config.py would be helpful.

frontend/src/routeUtil.js
```diff
@@ -2,9 +2,9 @@ import { push } from 'redux-first-router';
 
 // path constants
 export const ROUTE = 'route';
-export const DIRECTION = 'direction';
-export const FROM_STOP = 'fromStop';
-export const TO_STOP = 'toStop';
+export const DIRECTION = 'dir';
```
exxonvaldez: I know we'll probably not need these constants once we address the issues with #347, just curious as to what the motivation was for shortening these?  Once the Path class is rewritten, we'll either have no text in the URLs or the text will match the graphParams fields, so seems like it is best to hold off changing this?  Not a big deal.

EddyIonescu:
Thanks for answering my questions! 

Would custom_directions also allow specifying the shape_id(s)? 

As for changing stop IDs, this could happen either through renumberings or the stop being moved (in which case it'd usually get a new ID); do you think versioned GTFS's will be needed? As well, I think versioned ```custom_directions``` will also be needed for cases where branches get changed/added/removed (TTC does this all the time) so that viewing older data is still accurate - but that's for a future PR.

And thanks for allowing multiple stop_ids to be entered in order for distinguishing branches rather than just a unique one, which would've caused issues for TTC. Also yes ```trip_headsign``` isn't good since different branches could use the same one (such as the 38 heading inbound). I agree that defining them manually works best.

For large agencies I guess the ```custom_directions``` could be generated through an agency-specific script by parsing the headsigns as TTC has somewhere around 100 routes that have multiple branches - obviously I'll be the one to do that.

Anyways the code itself looks good and I assume you've tested it. Feel free to merge!

youngj:
@EddyIonescu I didn't implement defining custom_directions via shape_id since Muni's GTFS feed doesn't keep the same shape IDs across different versions. There might be other agencies that do have stable GTFS shape_ids, but I haven't checked.

I think that custom_directions will occasionally need to be changed if the GTFS file changes in ways that require changing included_stop_ids/excluded_stop_ids. We could probably just change custom_directions without keeping the old definitions in the YAML file as long as we don't need to recompute arrival times on old vehicle data. But it may be necessary to keep multiple versions of custom_directions in the YAML file to allow rerunning compute_arrivals.py on old data again. I guess we can deal with that if it happens. In order to support that we would also need a way to fetch old versions of the GTFS feed.

Could you create an opentransit-data S3 bucket in AWS and enable public read access?

EddyIonescu:
@youngj added S3 bucket ```opentransit-data``` with same permissions and IAM policies as ```opentransit-precomputed-stats```.

I see, so how would one define a shape for a branch? Or would it find one based on the included stops?

Also, versioned GTFS's can be fetched from transit.land (http://transit.land/feed-registry/operators/o-9q8y-sfmta), which has a lot of agencies. It was part of Mapzen and is now maintained by Interline and supported by AWS research grants so it should be dependable for a while (https://transit.land/news/2019/01/08/transitland-continues.html).

I see, ```custom_directions``` is only needed when pre-computing the data; different versions would be needed for computing older Orion data but I don't think we really need that anyways.

exxonvaldez:
Being able to recompute would be great.  Today the app runs out of data at the date of the last version change, currently Oct 1, I think.  The UI isn't able provide any guidance about what the earliest allowed date is.

It'd be nice if the custom directions could have dates associated with them, where the date indicates when this entry supercedes prior versions of that direction.

Also there is an extensive gtfs archive here:  https://transitfeeds.com/p/sfmta/60

docs/ADVANCED_DEV.md
youngj: Done

frontend/src/routeUtil.js
youngj: Reverted this change -- briefly tried to add the agency ID to the URL to allow frontend to show routes from multiple agencies, and was trying to modify the URL structure before giving up.

Issue #387, opened 2019-11-07
Loading indicator while the app pulls data


akgupta89:
When selecting a route and selecting tops X to Y there is a significant delay while the app pulls data from the bucket and that varies based on a user's bandwidth. For my laptop it took between 3-4 seconds without any indication that it was performing this process which isn't great for a UX perspective.

exxonvaldez:
Also would be worthwhile to do some profiling.  I suspect that moving the Marey chart to its own tab and also deferring the load and rendering of its data will help with performance. #375

exxonvaldez:
Okay, I think this one can be closed now.  #432 covers profiling.  The Marey Chart is now on its own tab and loaded lazily.  And there is an indeterminate circular loading indicator to the left of the date picker via #411 .

Issue #380, opened 2019-11-06
Standarize margin/padding for control panel / graphs


akgupta89:
The margin/padding between the control panel and graphs do not look standardized, this is especially noticeable in mobile view. Ideally they should all be equal distant from the edge.

![image](https://user-images.githubusercontent.com/40367955/68267125-36f43800-0006-11ea-8e24-f854a8c84002.png)

sourcediz:
Hey @akgupta89 I have fixed this issue through implementation of changing styles so I was wondering is there is a file architecture you have in mind for style sheets?

For now I can simply add the css to the current App.css under an inline comment then a seperate issue can be put forth later to refactor styles into your chosen choice of file structure?

Stumbled upon this repo/app looks really interesting, always love my data & graphs and can't wait to contribute more to this project.
Maybe Ive missed it but is there any specific documentation relating to contributing to the frontend?

Cheers

akgupta89:
> Hey @akgupta89 I have fixed this issue through implementation of changing styles so I was wondering is there is a file architecture you have in mind for style sheets?
> 
> For now I can simply add the css to the current App.css under an inline comment then a seperate issue can be put forth later to refactor styles into your chosen choice of file structure?
> 
> Stumbled upon this repo/app looks really interesting, always love my data & graphs and can't wait to contribute more to this project.
> Maybe Ive missed it but is there any specific documentation relating to contributing to the frontend?
> 
> Cheers

Hi Dali,

Currently it'll be fine to do it inline (preferably inside the component file itself), we do plan to split out the css into separate files. We currently do not have any advanced frontend documentation, but it's something we might want to look into as the frontend codebase expands @hathix

exxonvaldez:
Just a heads up that in #375 I'm going to remove the Paper component wrapper around the "travel time" chart and Marey chart, and put them behind tabs.  See issue for details.

sourcediz:
@exxonvaldez sweet would you be able to tag me or comment here when you submit the PR or the branch is ready to clone? Cheers

exxonvaldez:
@sourcediz Took a while, but the PR is now merged into master.

PR #411, opened 2019-11-11
Route Summary UI fixes

### LINKED ISSUE #375: Better support for mobile on Route Summary page
@hathix @akgupta89 . I noticed that on a small display (iPad), you cannot see the entire Marey chart or scroll it.  On a phone, you cannot see the entire travel time chart.  This is because both the Marey chart and travel time chart are within Cards (similar to Paper) that are sized to the window and clip their contents.

I propose the following fixes:

1.  Remove the Cards and enclosing Grids so that the charts can push out the width of the page and scroll (this is what we do on the stop-to-stop page, so the bar charts are scrollable).
2.  Add Tabs to the Route Summary (like we have on the Info component) for the Summary cards, the travel time chart, and the Marey chart.  This should make it easier to browse/discover the Route Summary content.

It should be really quick to implement this, just wanted to file an issue first to give you guys a heads up.

### LINKED ISSUE #400: Fix inconsistency between route stats & marey chart when no data
Screen Shot 2019-11-06 at 7 05 09 PM

Go to a trip with no data logged, and you'll see the expected gray cards with no stats (the NaN is weird, btw), but you'll also see a Marey chart that shows a bunch of runs. I'm not sure whether the Marey chart or the stats are wrong, but there's an inconsistency here... let's figure out what's the matter.

exxonvaldez:
Fixes #375. Fixes #400. Partial fix for #387.  Mitigations for #380.

## Proposed changes

- actions/index.js: New action creator to reset arrival data (analogous to reset graph data action) which also doubles as an arrivals error action, depending on the payload (null means no error, non-null means error).
- MareyChart: update `useEffect` hooks to clear stale data/chart renderings (#400) and load arrivals data lazily (partial fix for #387).  Improve mobile experience by removing enclosing `Grid`/`Card` elements that clip (#375).  When there is no data, show a simple text loading indicator (assumes we must be loading unless there is an arrivals error, in which case we show the arrivals error).
- RouteSummary:  reduce scrolling and improve mobile experience by adding AppBar (tabs bar) like the existing ones on the `Info` screen with three tabs (summary, travel time, marey chart) (#375).
- TravelTimeChart:  Remove enclosing `Grid`/`Card` elements to prevent clipping on mobile.  Also add a simple text prompt that appears when no direction is selected, so that the tab is never completely blank.
- routesReducer: add UI's requested date to arrivals for comparison with currently selected date.  Handle request for reset of arrivals data (and setting of arrivals error payload, if any).

Note that testing the fixes for #400 and #387 has a number of test cases.  I used a (now commented out) console.log statement to verify that arrivals data was being purged as soon as the route or date was changed.  I used the developer tools to verify that `arrivals_v4a...json.gz` was not getting loaded unless the Marey Chart tab was selected, and that it was getting loaded only once.  And then with the summary tab showing, I verified that page response time was improved when changing routes or dates and then trying to pick a from stop.  I call this a partial fix because I didn't do profiling or slow bandwidth simulation and only a text loading indicator for just the Marey Chart.  I figured this was an obvious heavy network and CPU load that should at least be made lazy so that it doesn't always fire when changing the route selection.

## Screenshot

New route summary tabs:

Screen Shot 2019-11-11 at 3 23 56 AM

Able to scroll Marey chart on small screens (same goes for travel time along route chart), not clilpped at window width:

Screen Shot 2019-11-11 at 3 44 47 AM

New arrivals error display (this is where 'Loading...' is shown if no error):

Screen Shot 2019-11-11 at 4 30 22 AM

## Link to demo, if any

http://ot-terence.herokuapp.com


hathix:
In #427, Jesse mentions that not all transit agencies use Inbound and Outbound, so the Marey chart treatment is a bit flaky. Plus, in this case, having 2 checkboxes would be cleaner UI. But I won't let that one block merging this -- we can track it in that issue.

Screen Shot 2019-11-20 at 9 02 44 PM

hathix:
Everything looks good besides the Marey chart thing -- when it scrolls off the side of the page, it looks weird, especially when the header bar stops spanning the whole page.

Screen Shot 2019-11-20 at 9 10 00 PM

Could we put this chart inside its own scrolling pane (so you can't "over-scroll"), or let the user pop the Marey chart into a lightbox?

exxonvaldez:
@hathix Demo at https://ot-terence-route-summary.herokuapp.com/

Marey chart and travel time chart now scroll horizontally as needed within the boundaries of the screen.  This works better on mobile, where you can scroll by swiping anywhere on the charts.  On desktop, you have to go the bottom of chart to get to the scrollbar, but it less likely that horizontal scrolling is needed.  There is not a lot we can do here because the react-vis charts are currently fixed width.  They can also be made responsive, but it's a lot more work to make a properly resizable chart.

Marey Chart tab now has a scroll bar:

Screen Shot 2019-11-24 at 11 52 27 PM

Travel time chart tab now has a scroll bar:

Screen Shot 2019-11-24 at 11 52 41 PM

exxonvaldez:
Adding in some tiny doc updates for testing with other devices against a dev machine, to reflect the effect of #409 .

PR #412, opened 2019-11-14
Attempt at clipping isochrone overlap to service area

### LINKED ISSUE #338: Disallow walking on water on the isochrone
We don't let users place the "start" of their trip on water, but the map drawing still thinks you can walk on water. (Try placing your pin on Treasure Island and see how it thinks you can walk on the Bay.) We should get our routing algorithm to ignore watery areas.

wllgrnt:
Doesn't fix #338.

We use turf's geoJson processing tools to clip all the geoJson as it is added to the layer. This is slow, and buggy when malformed/self-intersecting polygons are passed (which they are).

This commit also stops you from clicking a second point to set an endLatLng that's outside the service area (i.e. the sea).

## Proposed changes

- Adds an isInServiceArea check to endLatLng at the start of the showTripInfo function in Isochrone.jsx.
- Adds a clipGeoJsonToServiceArea to the end of mapGeometry.js which clips any geoJson given, returning unclipped geoJson if  a TopologyError is thrown (for now) to check why that's happening

Heroku demo to follow

### DIFF
--- a/frontend/src/helpers/mapGeometry.js
+++ b/frontend/src/helpers/mapGeometry.js
@@ -98,3 +98,38 @@ export function isInServiceArea(latLng) {
     return turf.booleanWithin(point, feature);
   });
 }
+
+// Clips any geoJson to the service area.
+// For all geoJson Features, check each Feature in the ServiceArea.
+// If the geoJson feature is completely enclosed in the serviceArea, retain the feature
+// If not, clip it.
+export function clipGeoJsonToServiceArea(geoJson) {
+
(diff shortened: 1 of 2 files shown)


wllgrnt:
This is a draft PR since I can't see a way around the clipping problem using turf - our polygons are causing errors when we try to work out the intersection, since they're complicated, and trying to use turf.simplify() or turf.truncate() throws a separate set of errors.

However, the isInServiceArea check is an easy improvement on the current code.

wllgrnt:
See https://opentransit-cliptoservicearea.herokuapp.com/isochrone

Issue #78, opened 2019-04-25
Steal Uber Movement's UI for date/time picking


hathix:
![Screenshot 2019-04-24 at 8 04 34 PM](https://user-images.githubusercontent.com/728084/56707128-84ddca00-66cc-11e9-8549-c5577bf16757.png)

hathix:
They do a great job adding power and flexibility to this. This way, you can ask questions like "what's the average travel time on weekdays last march?"

exxonvaldez:
There's a bit of conumdrum when it comes to selecting a date picker UI.  You would think Material UI core would have at least a basic one (and it used to, apparently), but it doesn't.  It relies on the browser's native date picker, if any.  This isn't great -- it's inconsistent among browsers (some implement a picker, some don't), and doesn't look like the rest of the UI.

There is an active, non-core Material UI date picker called "Material UI Pickers" at https://material-ui-pickers.dev/demo/datepicker .  It doesn't support range selection, but is built to the Material UI design spec and looks good.   Range seleciton has been a frequently requested feature since 2018, but there's still no sign of that particular feature.

Uber uses its own React component library (a competitor to Material UI) called Base Web, and they use it for Uber Movement.  Their picker supports date ranges.

There's a React picker with range selection that hasn't been updated since June 2018: https://github.com/onefinestay/react-daterange-picker .  Stylistically it doesn't match Material UI.

I'm thinking that most of the time, predefined ranges should meet people's needs (last 7, 30, 90 days relative to the main date picker's date).  Those choices could be made via buttons or radio button group.  Then we can also offer a custom range UI where they use a secondary date picker to pick the starting date.

So, basically, whether we continue to use the native picker or upgrade to the Material UI Pickers one, we are forced to do a date picker UI that is not as optimal as the Uber Movement one.

hathix:
We've stolen A LOT of it, and culturally we understand to look to Uber movement for guidance

Issue #379, opened 2019-11-06
Fix Header Layout in Mobile


akgupta89:
The header should have full width columns in mobile view, such that the breadcrumbs stack above the date range picker

![image](https://user-images.githubusercontent.com/40367955/68266773-f6e08580-0004-11ea-80b3-bfeebfc2edc6.png)

exxonvaldez:
Fixed this in #415 by moving most of the bread crumb below the blue app bar into its own Paper.  This will also make room in the app bar for a second date picker for date range comparisons (#321).  Still waiting to see if Chris makes updates on #347 , it'll need to be updated to reflect the new layout of bread crumbs.

![Screen Shot 2019-11-26 at 12 06 29 AM](https://user-images.githubusercontent.com/44861283/69611299-55849800-0fe2-11ea-9b8d-6fcf33ccc5a1.png)

Issue #321, opened 2019-09-26
add support for a/b testing


hathix:
Let the user compare two date ranges to see if a fix, crisis, new policy, etc had an impact on various metrics. This was a big ask from tal of sfmta - their engineers would love it.

exxonvaldez:
This isn't blocked for stop to stop statistics.  Just depends on #415 .

hathix:
We can now do date range comparison for stop-to-stop routes in particular routes (e.g. route 14 from stop A to stop B)

PR #415, opened 2019-11-15
Date Range UI

exxonvaldez:
## Proposed changes

As discussed in #78, there isn't a good date picker widget with range selection available.  So for now, adding a UI that uses radio buttons to choose a date range, with a custom date range option that reveals a second native date picker.  I think the predefined date ranges options are more likely to be used than the custom date.

Couple of issues that I noticed that we can either address in this PR or create new issues for:

1. This gets really slow for querying for over 30 days of data, plus we usually don't have that much data available due to missing arrivals files (due to version changes).  30 may be a more practical limit.

2. Since date range only affects "stop to stop", I'm thinking I should add something to show the current date range on the Info page only.   Right now there's no indication of the current date range -- if date ranges worked everywhere, I would add it to the top right where the end date is.
 
Summary of changes:

- actions/index.js: Add logic to compute a list of selected days based on graphParam state for the graphQL API.
- DateTimePanel: Add radio button selection of "days back" for the stop to stop info, including a custom range with a second native date picker.  Add checkboxes for enabling/disabling days of the week as in Uber Movement.
- UIConstants: Add constants for date and day of week selection.

## Screenshot

This is the initial state of the date range UI:

Screen Shot 2019-11-15 at 9 50 23 AM

Showing when "custom start date" is chosen:

Screen Shot 2019-11-15 at 9 50 41 AM

Proof that there's more than one day's data (2938 arrivals):

Screen Shot 2019-11-15 at 9 51 12 AM

## Link to demo, if any


youngj:
Days of the week UI looks good to me.

Having radio buttons that set the start date "back" relative to a set end date seems somewhat unintuitive, as well as displaying the end date visually above the custom start date. I don't remember seeing other date range UIs like that. I think it would be simpler if there were buttons like "last day" "last week" "last month" which set both the start date and end date, and then an ability for the user to set both a custom start date and end date below that.

"Stop To Stop Info" is more like internal jargon than something that users would understand. I think users would almost universally assume if they set a date range that all displayed stats were relative to the entire date range, regardless of the disclaimer that the date range applies to "Stop To Stop Info".

If the user sets a date range and we don't have the ability to compute stats for that date range yet, I think we should just hide that stat from the UI and perhaps replace it with a message saying that it can't be shown for that date range.

exxonvaldez:
Good feedback.  The "days back" approach is part of making the range look secondary, as only the end date is used by these screens:

- isochrone
- route table
- route summary (including route summary stats, travel time chart, and marey chart)
- data diagnostic page

  I see a couple of short term options.

1.  We could show the date or date range on each screen that is actually in effect, in the blue app bar.  There would be an info icon acknowledging that a date range was set, but date ranges are not supported for the current screen.

2.  We can move setting of the date range to a separate UI that lives only on the stop to stop screen.  It'll be a little tricky to make it mesh well with the app bar's date picker (if you thought the current UI was not intuitive, this will probably be worse).

I'm currently favoring the first option, and I can redo the date range selection UI as you suggested.  So two date textfields, start date and then end date, followed by buttons for preset ranges.

exxonvaldez:
@youngj  , review feedback has been checked in.

Revised date range picker and complete date range in app bar:

Screen Shot 2019-11-19 at 10 10 06 PM

App bar showing info icon and not full date range because isochrone doesn't support it.  Info icon won't appear if date range is not set:
 
Screen Shot 2019-11-19 at 10 08 43 PM

Info icon message:

Screen Shot 2019-11-19 at 10 09 04 PM

exxonvaldez:
> * Can you make the Weekday/Weekend boxes look more like headers? Perhaps by bolding or adding an underline.

Added a Divider (hr), see screenshot below.

>  * Can you make it impossible to choose an end date before the start date, perhaps by graying out end dates that are earlier than the start date? i.e. I shouldn't be able to do this:

The start date is already restricted to 0-90 days before the end date.  So restricting end date to after start date would make it hard to query dates in 2018 (if we had the data for it).  I did a fix where, upon changing the end date, the start date is adjusted to stay in the 0-90 day window.  Seems to work well.

Screen Shot 2019-11-21 at 7 29 03 PM

exxonvaldez:
Need a re-review for these changes:

- Moved most of the bread crumb (except agency name) out of the app bar.  This is to make room for a future second date range widget for date range comparisons (#321).  This is also to solve bread crumb wrapping issues on small screens (#379 and #431).  Now, when the bread crumb wraps, it doesn't squash neighboring elements and cause layout issues.

- Added an interdeterminate progress loading indicator to the left the date-time dropdown (and left of the date range info icon) (#387).  This indicator will appear whenever any of the five axios requests fire.  Added a new "loading" Redux reducer to track loading state based on the names of action types.  When @wllgrnt tested this PR, he mentioned the need for loading indicator because when changing the date range, the UI sometimes will not respond for a while.

New route screen layout:

Screen Shot 2019-11-22 at 3 29 33 AM

Loading indicator (note: I plan to make the date picker have an "apply" button in #321, so that clicking widgets in the date panel doesn't trigger constant queries and loads) :

Screen Shot 2019-11-22 at 3 28 56 AM

akgupta89:
The one callout, I'd like to make is that having a bunch of checkboxes probably isn't the most mobile friendly option, maybe consider a multi-select dropdown of some sort.

exxonvaldez:
> The one callout, I'd like to make is that having a bunch of checkboxes probably isn't the most mobile friendly option, maybe consider a multi-select dropdown of some sort.

It’s a good thought.  I don’t know if you saw the indeterminate checkbox state for Weekend and Weekday, when there’s a mix of checked and unchecked boxes below those.  Live demo at https://ot-terence.herokuapp.com/ . That’s copied from uber movement.  There is something nice about how they are “parent” checkboxes that double as indicators of whether none/some/all are selected, even if there’s no indenting of the children below them.

When you say not mobile friendly, is that referring the vertical space taken up by the checkbox group or something else?  Since the user should be able to see the all of the current day of week selections even when a multiple-select is closed, I was thinking we’d probably wind up with the “chip” version which nearly as tall.

hathix:
> Moved most of the bread crumb (except agency name) out of the app bar. This is to make room for a future second date range widget for date range comparisons (#321). This is also to solve bread crumb wrapping issues on small screens (#379 and #431). Now, when the bread crumb wraps, it doesn't squash neighboring elements and cause layout issues.

I like your solution -- it's good UX, especially on mobile. But the breadcrumb links don't work, and the breadcrumb pane doesn't stand out from the rest of the page (try bolding it?) Perhaps you can also turn it into a "real" Material Breadcrumb component. Or just file a new issue to enable this work.

Screen Shot 2019-11-24 at 6 24 59 PM


> Added an interdeterminate progress loading indicator to the left the date-time dropdown (and left of the date range info icon) (#387). This indicator will appear whenever any of the five axios requests fire. Added a new "loading" Redux reducer to track loading state based on the names of action types. When @wllgrnt tested this PR, he mentioned the need for loading indicator because when changing the date range, the UI sometimes will not respond for a while.

I like it.

> The one callout, I'd like to make is that having a bunch of checkboxes probably isn't the most mobile friendly option, maybe consider a multi-select dropdown of some sort.

Like Terence, I don't think it's a huge problem. It works pretty well on my phone.

Screenshot_20191124-182832

exxonvaldez:
> I like it. Just make sure you either fix the problems with the new breadcrumb solution, or raise a new issue.

The breadcrumb you're looking at is essentially the same as the current production breadcrumb, except for the black on white colors instead of white on blue.  There's no need for a new issue since @ckrajewski has linked breadcrumbs in progress ( https://github.com/trynmaps/metrics-mvp/pull/347 ), which would have be updated to reflect the slightly different layout. It's been 14 days since we requested significant changes and no sign of Chris since, not sure whether we should reassign to someone else to finish up -- mostly just needs internal reworking, basically that PR doesn't use any new material UI classes, just some styling adjustments on the links.

I'll test my PR against master since some stuff was merged in, and then merge.

PR #417, opened 2019-11-19
Wording fixes.

### LINKED ISSUE #382: Fix confusing "90% of trips" card
When you choose a trip with a to & from stop selected ([example](https://muni.opentransit.city/route/L/direction/L____I_F00/fromStop/5726/toStop/5417)), you get this weird "90% of trips" card. It's confusing.

Screen Shot 2019-11-06 at 6 07 41 PM

Instead, we should:

* Put the average trip card first, since that's what people are interested in knowing
* Call "90% of trips" the "worst case" and explain somewhere (in a tooltip?) that it represents the 90th percentile of trips

### LINKED ISSUE #392: Provide descriptions of what each measurement means
In the Routes table we have:

- Name
- Score
- Wait (min)
- 20 min wait %
- Speed (mph)
- Extra Travel (min)

I can kind of guess what some of these mean, but "Extra Travel" for instance, is baffling. We should provide descriptions of what each column means and make it available through something like a tooltip.

exxonvaldez:
## Proposed changes

On the stop to stop page:
- Rename "90% of trips" to "most trips (less than)"
- Add an info icon for "most trips" to explain that "most" is 90% of trips.
- Add an expanded definition of travel variance to the popover.

On the route summary:
- Add an expanded definition of travel variance to the popover.

On the route table:
- add a popover explaining the column titles.

Also on these three pages:
- Rename "extra travel" to "travel variance"
- Limit popover widths to 500px max.

## Screenshot

Stop to stop with "most trips (less than)" and "travel variance":

Screen Shot 2019-11-18 at 7 34 39 PM

Stop to stop with travel variance popover:

Screen Shot 2019-11-18 at 7 35 19 PM

Route summary with travel variance popover:

Screen Shot 2019-11-18 at 7 36 03 PM

Route table, new column headings and info icon:

Screen Shot 2019-11-18 at 7 36 50 PM

Route table with popover:

Screen Shot 2019-11-18 at 7 37 07 PM

## Link to demo, if any

n/a

### DIFF
--- a/frontend/src/UIConstants.js
+++ b/frontend/src/UIConstants.js
@@ -25,6 +25,7 @@ export const CHART_COLORS = [Colors.GRAY, Colors.PURPLE];
 // long the trip will "usually" take.  So for 90th percentile
 // this would be the historic performance 9 times out of 10.
 export const PLANNING_PERCENTILE = 90;
+export const TENTH_PERCENTILE = 10;
 
 // a commonly used style option for react-vis
 // Crosshairs (hovers) where we don't want a crosshair line
--- a/frontend/src/components/InfoScoreCard.jsx
+++ b/frontend/src/components/InfoScoreCard.jsx
@@ -15,7 +15,7 @@ import Rating from '@material-ui/lab/Rating';
 import Box from '@material-ui/core/Box';
 import {
   quartileBackgroundColor,
-  quartileForegroundColor,
+  quartileContrastColor,
 } from '../helpers/routeCalculations';
 
 /**
@@ -60,7 +60,7 @@ export default function InfoScoreCard(props) {
         ? quartileBackgroundColor(myGrades[myGradeName] / 100.0)
         : 'gray',
       color: myGrades
-        ? quartileForegroundColor(myGrades[myGradeName] / 100.0)
+        ? quartileContrastColor(myGrades[myGradeName] / 100.0)
         : 'black',
       margin: 4,
     };
--- a/frontend/src/components/InfoScoreCard.jsx
+++ b/frontend/src/components/InfoScoreCard.jsx
@@ -38,6 +38,7 @@ export default function InfoScoreCard(props) {
   const useStyles = makeStyles(theme => ({
     popover: {
       padding: theme.spacing(2),
+      maxWidth: 500,
     },
   }));
 
--- a/frontend/src/components/InfoScoreLegend.jsx
+++ b/frontend/src/components/InfoScoreLegend.jsx
@@ -8,7 +8,7 @@ import { Table, TableBody, TableCell, TableRow } from '@material-ui/core';
 
 import {
   quartileBackgroundColor,
-  quartileForegroundColor,
+  quartileContrastColor,
 } from '../helpers/routeCalculations';
 
 export default function InfoScoreLegend(props) {
@@ -24,7 +24,7 @@ export default function InfoScoreLegend(props) {
               <TableCell
                 align="right"
                 style={{
-                  color: quartileForegroundColor(row.value / 100),
+                  color: quartileContrastColor(row.value / 100),
                   backgroundColor: quartileBackgroundColor(row.value / 100),
                 }}
               >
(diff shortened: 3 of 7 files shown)


youngj:
now I'm thinking that "Typical Travel Time" would be a better phrase than "Travel Time Range" (to replace "Most Trips"), since it would show the interdecile range, not the full range (min/max trip times). Then the value could show both the 10th percentile and 90th percentile, e.g. "18 to 23 min"

exxonvaldez:
> Also, thanks for sharing what you are thinking about the use case for showing "planning" wait+travel time as a way for users to plan when to start checking predictions in order to be sure to make it to their destination on time.
> 
> If someone is asking "how long should I plan for the whole thing to take?", I still think that adding in 90th percentile wait times if the rider ignores schedules and predictions (which is not an accurate model of most transit riders' behavior) will mean that the big Most Trips number is less relevant to most users, and in particular it will make public transit appear significantly worse than it actually is.

Hmm, you lost me a bit here.  Let's make this more concrete.  Let's say I want to get to a meeting at 6pm and don't want to be late.  Call this model is "get to a meeting" model.  Suppose for my route the 90th percentile travel time is 20min, the 90th percentile wait is 25min, and the walk time to my starting stop is 5min.   Assume the destination stop is next to my meeting.  Isn't the correct rider behavior model to start checking predictions at 20 + 25 + 5 = 50 min, or 5:10pm, if not earlier?  It's possible that I might have to leave before 5:15, if the predictions are for my starting stop are 5:20 and 5:45.  Or the predictions might say there's one at 5:40, and I can wait where I am until 5:35.

In this model, I would be ignoring schedules but not ignoring predictions.  Correct me if you think otherwise.  90th wait plus 90th travel is a valuable number in this model.  What if we label this number something like nyc busstat's "planning time" or "check predictions time" or "buffered time" and add an explanation of the "arrive at or before a certain time" model in a popover?

Let's call the other rider model I can think of the "ad hoc" model.  I'm ready to go at some random time and don't care when I arrive.  So I'd check predictions and head out to meet the next vehicle that's more than 5min out.  In this case, I'd compare the 50th percentile wait with the predictions, to make sure that the route is operating normally.  If the vehicles are bunched up 35 min away, they may be blocked or there was a gap and there will be crowded and empty vehicles moving together in a slow pack.  And then use 50th percentile travel time as a rough expectation of when I'd arrive.

I use both models.  I see them as complementary.

exxonvaldez:
Okay, here's a quick tour of the latest iteration.  See also commit description.

First, a look at the stop to stop screen, with updated "typical journey" and "journey planning" cards.  "Typical" is now first:

Screen Shot 2019-11-20 at 2 03 22 PM

.
. (some blank space)
.
.
This is the info popover for typical journey:

Screen Shot 2019-11-20 at 2 04 03 PM

.
.
.
.
This is the info popover for journey planning:

Screen Shot 2019-11-20 at 2 04 10 PM

.
.
.
.
A look at the travel variability popover:

Screen Shot 2019-11-20 at 2 04 40 PM

.
.
.
.
A quick look at the route summary (popovers are similar to stop to stop, with wording adjustments to reflect route level differences, such as average speed):

Screen Shot 2019-11-20 at 2 06 02 PM

.
.
.
.
Updated route table with new headers and less padding.  I also used lighter text for the four metrics where the subscore was above 50, to emphasize which parts of routes have issues:

Screen Shot 2019-11-20 at 2 07 16 PM

.
.
.
.
And lastly, revised info popover for the route table:

Screen Shot 2019-11-20 at 2 07 31 PM

exxonvaldez:
Demo at https://ot-terence-wording.herokuapp.com/

exxonvaldez:
Thanks for the additional feedback.  I'll drop the "<" symbol.

For your other proposals, I'd like to get some feedback on this iteration first (from the OpenTransit team or user research) as well as the proposals.  I can create an umbrella issue to track the proposals, and ideally we'd be able to do some A/B testing.  Maybe just surveys on #opentransit slack if nothing else, using screenshots or rapid prototypes. For example, I was able prototype your table changes quickly, and I'd want to know how people feel about them:

Screen Shot 2019-11-20 at 3 46 24 PM
 
vs.

Screen Shot 2019-11-20 at 2 07 16 PM

youngj:
Sounds good, thanks for the prototype!

PR #418, opened 2019-11-20
replace %REACT_APP_METRICS_BASE_URL% with empty string in production index.html file

youngj:
Apparently %REACT_APP_METRICS_BASE_URL% is replaced in index.html when running the development server, but is not replaced when running `npm run build`. This caused https://muni.opentransit.city/ to stop working. To fix this, replace %REACT_APP_METRICS_BASE_URL% when building production assets in the Dockerfile.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,6 @@
 node_modules
 frontend/node_modules
+backend/data
 data
 venv
 .git
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,7 @@ ENV PORT 5000
 CMD gunicorn --bind 0.0.0.0:$PORT metrics-api:app
 
 FROM react-dev as react-build
-RUN cd /app/frontend && npm run build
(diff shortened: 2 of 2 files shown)


exxonvaldez:
Interesting.  Seems like the original code should work, I don’t see any caveats at https://create-react-app.dev/docs/adding-custom-environment-variables

exxonvaldez:
Or is the problem that docker-compose.prod.yml doesn't define the variable so no injection occurs?

exxonvaldez:
(in which case do we need the same fix for the heroku deployment also?)

youngj:
This change fixed the issue for GKE and should also fix it for Heroku. It doesn't need to be set in docker-compose.prod.yml because nginx only serves static files. Variables in index.html (like %PUBLIC_URL%) are replaced before then when the image was built from the Dockerfile. 

I'm guessing it's because REACT_APP_METRICS_BASE_URL is undefined in the Dockerfile rather than the empty string (if React loops over all environment variables starting with REACT_APP then it wouldn't include REACT_APP_METRICS_BASE_URL).

Issue #429, opened 2019-11-21
Get feedback on #417 UI changes and followup proposals for route table, stop to stop


exxonvaldez:
#417 was just merged in, and it would be good to get feedback on the changes.  Also @youngj proposed additional changes in https://github.com/trynmaps/metrics-mvp/pull/417#pullrequestreview-320462483

1.  On stop to stop:  Remove median wait time from "typical journey", and make median travel time a large number.  May mean retitling the card to "median travel time".  Depends on the prevalent user model for a typical trip -- should it include wait time (either at the stop or checking predictions elsewhere) or not?

2.  Or instead of number 1 above, replace the typical journey and journey planning cards with an nyc busstats style bar chart.  This takes away emphasis from the sums of wait and travel, showing 4 numbers for the user to interpret (median wait, median travel, 90th %ile wait, 90th %ile travel).

3.  For the route table, add quartile background colors and units to the cells.  This is prototyped in https://github.com/trynmaps/metrics-mvp/tree/tc-route-table-colors .

Proposed:

![Screen Shot 2019-11-20 at 3 46 24 PM](https://user-images.githubusercontent.com/44861283/69289791-844dd900-0bb2-11ea-9136-06f8d739cf35.png)

vs latest:

![Screen Shot 2019-11-20 at 2 07 16 PM](https://user-images.githubusercontent.com/44861283/69289794-8748c980-0bb2-11ea-97b6-9cf08d51b3fb.png)

For all these changes, it would be good to an A/B survey (perhaps on opentransit slack, or via user research), maybe with quick prototypes or mockups.

hathix:
Terence, can you deploy your route table colors to Heroku so we can share it with people? Then we can start the poll

PR #437, opened 2019-11-21
RouteTable:  use filled Chips with light background colors and add units

exxonvaldez:
Related to #429.

## Proposed changes

Color table cells with metric values (wait, speed, etc.) by quartile.  Add units (min, mph) to each cell's value.

## Screenshot

Screen Shot 2019-11-20 at 3 46 24 PM

## Link to demo, if any

http://wpg-opentransit.herokuapp.com

### DIFF
--- a/frontend/src/components/RouteTable.jsx
+++ b/frontend/src/components/RouteTable.jsx
@@ -25,6 +25,7 @@ import {
   getAllSpeeds,
   getAllScores,
   quartileBackgroundColor,
+  quartileContrastColor,
 } from '../helpers/routeCalculations';
 
 import { handleGraphParams, fetchPrecomputedWaitAndTripData } from '../actions';
@@ -386,6 +387,9 @@ function RouteTable(props) {
                     >
                     <Chip
                       style={{
+                        color: quartileContrastColor(
+                          row.totalScore / 100,
+                        ),
                         backgroundColor: quartileBackgroundColor(
                           row.totalScore / 100,
                         ),
@@ -401,6 +405,9 @@ function RouteTable(props) {
                     >
                     <Chip
                       style={{
+                        color: quartileContrastColor(
+                          row.medianWaitScore / 100
+                        ),
                         backgroundColor: quartileBackgroundColor(
                           row.medianWaitScore / 100
                         ),
@@ -417,6 +424,9 @@ function RouteTable(props) {
                     >
                     <Chip
                       style={{
+                        color: quartileContrastColor(
+                          row.longWaitScore / 100,
+                        ),
                         backgroundColor: quartileBackgroundColor(
                           row.longWaitScore / 100,
                         ),
@@ -439,6 +449,9 @@ function RouteTable(props) {
                     >
                     <Chip
                       style={{
+                        color: quartileContrastColor(
+                          row.speedScore / 100,
+                        ),
                         backgroundColor: quartileBackgroundColor(
                           row.speedScore / 100,
                         ),
@@ -455,6 +468,9 @@ function RouteTable(props) {
                     >
                     <Chip
                       style={{
+                        color: quartileContrastColor(
+                          row.travelVarianceScore / 100,
+                        ),
                         backgroundColor: quartileBackgroundColor(
                           row.travelVarianceScore / 100,
                         ),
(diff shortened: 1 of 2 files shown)


hathix:
What are our thoughts on the "pill" solution that Terence mentioned on Slack? Looks better than the heavy colors this currently uses.

Screen Shot 2019-11-23 at 11 46 24 PM

exxonvaldez:
> What are our thoughts on the "pill" solution that we mentioned on Slack? Looks better than the heavy colors this currently uses.

- Pills seem to be much better received than cell backgrounds colors and text colors, so we can set aside the other options, and I can update this PR to be pills for people to try out.
- The hues still need some attention on differentiation.  There is a lime and an amber (for 2nd best and 3rd best quartiles) that look a bit too similar to me.
- If we go with these new hues, then should trying carrying them over to the cards on the route summary and trip summary.
- I think it's better with units, can revisit dropping units after we try a fixed table header (still might want units even then).

exxonvaldez:
I forgot that the colors automatically carry over to the other cards because they use the same function for colors.

Updated this PR as follows:

Switched from background cell colors to filled Chips.  Removed table row borders, added custom padding.

I made the lime and green backgrounds slightly darker than in the first Chip screenshot.

I also made the text color for the route and trip summary match the Material UI default (black with alpha of 0.87, as opposed to just solid black).  This change is imperceptible unless you know to look for it.

Updated route table:

Screen Shot 2019-11-25 at 1 10 30 PM

Route summary:

Screen Shot 2019-11-25 at 1 12 01 PM


Trip summary:

Screen Shot 2019-11-25 at 1 12 19 PM

exxonvaldez:
Removed table header units (except "Long Wait %", as that one didn't seem to work without the percent sign):

Screen Shot 2019-11-25 at 11 55 03 PM

exxonvaldez:
With headers scooted over to align with columns, which have extra padding due to chips.  Need some feedback on this and the potential issues I mentioned earlier.

Screen Shot 2019-11-26 at 12 01 43 AM

youngj:
What do you think about making some of the background colors a bit darker? I think it might not be as clear how the very light background colors rank from good to bad.

Using `range([red[200], amber[200], lime[200], green[100]])`:

image

exxonvaldez:
> What do you think about making some of the background colors a bit darker? I think it might not be as clear how the very light background colors rank from good to bad.
> 
> Using `range([red[200], amber[200], lime[200], green[100]])`


Yeah, that's a good point.  Going darker makes the colors punchier and easier to distinguish.  I tried a few iterations, and realized that going down to three colors made differentiation a lot easier and allowed for less punchy colors.  I think this why traffic maps from Google and Waze use green/yellow/red/dark red, and why I copied that for our speed colors, because any four color scheme was too complicated.

How about this one?  [red[200], red[50], yellow[100], green[50]]

Screen Shot 2019-11-27 at 12 54 50 AM

(The built in colors don't include a rust or dark red.  There is brown that wasn't close enough to red).

One thing I should mention is that using color alone to convey information is an accessibility no-no.  We'll need affordances for color blind users, like some other layer that conveys the same information (perhaps toggleable by user preference), such as a background texture, font weight or decoration, or a border pattern (like the partial circles), or small horizontal bars.  Ideally it'd be something that can be left on and help all users differentiate.

exxonvaldez:
Here's another three hue variant with two shades of green:  [red[100], yellow[100], green[100], green[200]]

Screen Shot 2019-11-27 at 10 24 34 AM

hathix:
You can merge this if you like -- just file an issue for making a color blind-friendly adjustment to the chips. Trello does it well, I think:

Screen Shot 2019-12-04 at 4 51 10 PM

hathix:
Can we use this color scheme with chips?

image

hathix:
Other ways to add patterns: show multiple dots (1-4 based on how good/bad), vary the outline width, etc. 

But I think making this color blind friendly is its own issue...

hathix:
We could also use different color boldnesses to make sure the colors stand out. Ideally you would want the chips to look clearly distinct even if viewed in grayscale.

exxonvaldez:
Okay, the SF team was good with going full circle back to the current colors with chips:  [red[300], yellow[300], lightGreen[700], green[900]].  The lightGreen was lightened from 800 to 700 to make things a little more distinguishable on the monochromatic scale.  Seanny has red-green color blindness of some kind, and the more intense colors are easier for him to distinguish.

I'm sure we'll revisit this again soon.

Screen Shot 2019-12-05 at 12 05 47 AM

PR #443, opened 2019-11-21
Tc route table colors lighter

exxonvaldez:
## Proposed changes

testing the auto labeler.
...

## Screenshot

...

## Link to demo, if any

...


hathix:
I don't think the labeler is working. It's a low priority thing so don't worry about it.

PR #446, opened 2019-11-21
Beautified the Sidebar with Material-UI

### LINKED ISSUE #381: Update Sidebar Navigation Styling
We recently implemented a footer to our sidebar with a very material-y design ( courtesy of @zachzwy ) and look to it which makes our current sidebar navigation look outdated, it would be great to update the app navigation links to match the look of the sidebar's footer

image

The bottom of the sidebar is the new design, and the top of the sidebar is the old design. Ideally the active app should stay highlighted when navigating between the dashboard and isochrone. Perhaps rename the "." to "debug".

### LINKED ISSUE #390: Rename the "." in the sidebar
We have a developer log page that's hidden in the sidebar as a ".". Just make the name clear -- "Developer Tools" or something. The "." is confusing and looks like a bug or speck of dust on the screen.

### LINKED ISSUE #402: Make menu and date popup more interactive (fly-out?)
A fly-out menu would be easier to use. Likewise, when clicking on the map or other parts of the page, the menu should flay away again.

### LINKED ISSUE #472: Check which page we're on and integrate into the navigation bar hover and click ripples.
If we set the `button` property of `ListItem`, we get a lovely hover effect. We can set `selected` to indicate a current selection, but getting that in code from `NavLink` is difficult.

SeannyPhoenix:
Fixes #402. 
Menu now disappears when clicking anywhere else on the page.

Fixes #390.
Dot is replaced with a meaningful label. Added icons.

Fixes #381.
Brings the rest of the menu in line with the Material-UI of the 'About' link.

Fixes #472.
Hover and active page works now.

## Screenshot

image

### DIFF
--- a/frontend/src/components/SidebarButton.jsx
+++ b/frontend/src/components/SidebarButton.jsx
@@ -1,5 +1,7 @@
 import React from 'react';
 import { NavLink } from 'redux-first-router-link';
+import { connect } from 'react-redux';
+import { components } from '../reducers/page.js';
 import Drawer from '@material-ui/core/Drawer';
 import IconButton from '@material-ui/core/IconButton';
 import List from '@material-ui/core/List';
@@ -14,7 +16,8 @@ import CodeRoundedIcon from '@material-ui/icons/CodeRounded';
 import InfoRoundedIcon from '@material-ui/icons/InfoRounded';
 import Divider from '@material-ui/core/Divider';
 
-function SidebarButton() {
+function SidebarButton(props) {
+  const currentPage = props.currentPage;
   const [drawerOpen, setDrawer] = React.useState(false);
 
   function toggleDrawer() {
@@ -62,6 +65,8 @@ function SidebarButton() {
               activeStyle={activeStyle}
               exact
               style={inactiveStyle}
+              button
+              selected={currentPage==components.DASHBOARD}
             >
               <ListItemIcon>
                 <MapRoundedIcon color="primary"/>
@@ -74,6 +79,8 @@ function SidebarButton() {
               activeStyle={activeStyle}
               exact
               style={inactiveStyle}
+              button
+              selected={currentPage==components.ISOCHRONE}
             >
               <ListItemIcon>
                 <TimelineRoundedIcon color="primary"/>
@@ -86,6 +93,8 @@ function SidebarButton() {
               activeStyle={activeStyle}
               exact
               style={inactiveStyle}
+              button
+              selected={currentPage==components.DATADIAGNOSTIC}
             >
               <ListItemIcon>
                 <CodeRoundedIcon color="primary"/>
@@ -98,6 +107,7 @@ function SidebarButton() {
               href="https://sites.google.com/view/opentransit"
               target="_blank"
               onClick={toggleDrawer}
+              button
             >
               <ListItemIcon>
                 <InfoRoundedIcon color="primary"/>
@@ -111,4 +121,10 @@ function SidebarButton() {
   );
 }
 
-export default SidebarButton;
+const mapStateToProps = state => ({
+  currentPage: state.page,
+});
+
+export default connect(
+  mapStateToProps,
+)(SidebarButton);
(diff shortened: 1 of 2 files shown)


frontend/src/components/SidebarButton.jsx
```diff
@@ -8,7 +8,13 @@ import ChevronLeftIcon from '@material-ui/icons/ChevronLeft';
 import MenuIcon from '@material-ui/icons/Menu';
 import { NavLink } from 'redux-first-router-link';
 import ListItemIcon from '@material-ui/core/ListItemIcon';
-import HomeIcon from '@material-ui/icons/Home';
+import TimelineRoundedIcon from '@material-ui/icons/TimelineRounded';
+import MapRoundedIcon from '@material-ui/icons/MapRounded';
+import CodeRoundedIcon from '@material-ui/icons/CodeRounded';
+import InfoRoundedIcon from '@material-ui/icons/InfoRounded';
+import Typography from '@material-ui/core/Typography'
```
exxonvaldez: Getting a warning about Typography defined but not used.

frontend/src/components/SidebarButton.jsx
```diff
@@ -82,11 +107,17 @@ function SidebarButton() {
...
+              href="https://sites.google.com/view/opentransit"
+              target="_blank"
+              onClick={toggleDrawer}
+            >
               <ListItemIcon>
-                <HomeIcon />
+                <InfoRoundedIcon color="primary"/>
               </ListItemIcon>
-              <ListItemText primary="About" />
+              <ListItemText primary="About" color="primary"/>
```
exxonvaldez: The About link is black, I guess the color attribute doesn't work here.  I'm not sure we want to go all blue for the drawer... the examples here are black:  https://material-ui.com/components/drawers/

Also check the Material UI style guide for drawers here:  https://material.io/components/navigation-drawer/#navigation-drawer .  I think the idea is to color the active link and icon, but not the inactive ones.

The About link looks kind of odd at the bottom, I think it should be with the other links.

akgupta89: The reason I originally asked for the About link to be at the bottom was because the rest are internal links that are maintained in the SPA, and the about link is an external link that opens a new tab.

exxonvaldez: I was thinking the same thing.  Divider: https://material-ui.com/components/dividers/

akgupta89:
This PR also addresses this issue: https://github.com/trynmaps/metrics-mvp/issues/381

akgupta89:
I think the design looks fantastic, but the icons don't look vertically centered with the text, and also maybe we should consider switching the dashboard and isochrone icons because the isochrone icon looks "very mappy" and like it should be the primary icon for this application.

frontend/src/components/SidebarButton.jsx
```diff
@@ -34,59 +45,64 @@ function SidebarButton() {
       >
         <MenuIcon />
       </IconButton>
-      <Drawer variant="persistent" anchor="left" open={drawerOpen}>
+      <Drawer anchor="left" open={drawerOpen} onClose={toggleDrawer}>
```
exxonvaldez: I noticed that on the data diagnostic page, the non-peristent drawer takes a very long time to appear (15 seconds or more, if at all), and pegs the CPU.  There's probably too much stuff on that page to work with the modal shading that transitions in.  I'll create a new issue for that.

frontend/src/components/SidebarButton.jsx
```diff
@@ -34,59 +45,64 @@ function SidebarButton() {
...
-            <ListItem>
-              <NavLink
-                to={{ type: 'DASHBOARD' }}
-                activeStyle={activeStyle}
-                exact
-                strict
-              >
-                Dashboard
-              </NavLink>
+            <ListItem
```
exxonvaldez: https://material-ui.com/components/lists/#selected-listitem I noticed that the demo drawers/list items have a nice hover effect and click effect (touch ripple).  You just have to add the attribute `button` to the `ListItem`s.

Also, it'd be great if you can figure out how to copy the styling for the active link (shaded background, no text color change) that appears in these list item demos.  It's an easier way to see what is active compared to changing the link color, and follows the MUI design guidelines.

hathix: Seanny: the problem with these links is that they don't auto-apply an "active" state to the page you're currently on, so adding Terence's requested feature is very hard.

PR #447, opened 2019-11-22
Map and Info scroll separate

### LINKED ISSUE #394: Map fixed when scrolling down the routes list
Please see issue #406 for more information.

Currently on the dashboard page, when you scroll down the routes list, the map will be "scrolled up" and you will see blank on the left. It would be great if the map is sticky there.

Screen Shot 2019-11-06 at 7 01 20 PM

### LINKED ISSUE #406: Prevent scrolling of page to keep map visible at all times
As the map is fixed to the height of the window, so should the right info panel, and instead should have a nested scrollbar to allow information to be seen while still keeping the map visible at all times.

It would be great to make the navigation component sticky while the page scrolls down so it's always visible, but this might not be ideal for mobile pages.

SeannyPhoenix:
Fixes #394.
Fixes #406.
On each of these screens, I implemented a nested scroll bars, as suggested by @zachzwy (https://github.com/trynmaps/metrics-mvp/issues/406#issuecomment-552127409).

## Screenshot

image

image


exxonvaldez (commented):
Thanks for taking this on, we were just talking about this yesterday about how we'd like our scrolling behaviors to be more sophisticated.

I'm not sure we want nested scrollbars, though?  That always looks a little odd.  Do you know why @zachzwy changed his mind about fixing the map and preferring nested scrollbars?  Is it easy to prototype having the app bar and map locked in place and have the rest of the page scroll below them?

akgupta89:
> Thanks for taking this on, we were just talking about this yesterday about how we'd like our scrolling behaviors to be more sophisticated.
> 
> I'm not sure we want nested scrollbars, though? That always looks a little odd. Do you know why @zachzwy changed his mind about fixing the map and preferring nested scrollbars? Is it easy to prototype having the app bar and map locked in place and have the rest of the page scroll below them?

I believe it looks odd currently because we see two scroll bars next to each other, which I'm not sure why is happening as we want to lock the normal scroll bar and make this a full page application. The reason I suggested nested scrollbars is because we want the map, the top bar, and maybe the route selector to be always visible ideally, I'm not sure how this would play out in mobile view however.

zachzwy:
> Do you know why @zachzwy changed his mind about fixing the map and preferring nested scrollbars? Is it easy to prototype having the app bar and map locked in place and have the rest of the page scroll below them?

In my mind, either "having the app bar and map locked in place and have the rest of the page scroll below them" or "having nested scroll bar on the RouteTable component" should achieve a pretty similar effect.

hathix (changes requested):
Just flagging this as "changes requested" given the feedback received so far.

hathix:
When I ran the page and nuked the header, the problem (of having the content be like 50px taller than the page) went away. So I think your problem has to do with the fixed header.

frontend/src/App.css
```diff
@@ -53,6 +53,18 @@
...
+.grid-container {
+  height: 100%;
+  overflow-y: hidden;
+}
+
+.grid-info-box
+{
+  height: 100%;
+  overflow-y: auto;
+  padding-bottom: 50px;
```
exxonvaldez: What is the purpose of the padding-bottom?  I tried removing it, and things seem to make more sense without the extra space.

frontend/src/screens/Dashboard.jsx
```diff
@@ -36,13 +36,13 @@ function Dashboard(props) {
...
       </AppBar>
-      <Grid container spacing={0}>
+      <Grid container spacing={0} className="grid-container">
         {/* Using spacing causes horizontal scrolling, see https://material-ui.com/components/grid/#negative-margin */}
         <Grid item xs={12} sm={6}>
           {/* map and table are both full width for 640px windows or smaller, else half width */}
           <MapSpider />
         </Grid>
-        <Grid item xs={12} sm={6} style={{ padding: 12 }}>
+        <Grid item xs={12} sm={6} style={{ padding: 12 }} className="grid-info-box">
```
exxonvaldez: I noticed that this didn't work well for a narrow window.  Note the xs={12} for the grid items, this means that the two items should be full width (and thus one above the other) for extra small windows (mobile browsers, and narrow desktop windows).  The scrollbar goes offscreen and you can't see all the content... basically doesn't work right for that layout.  It's going to be tricky to solve.  For narrow windows, the map goes to a fixed height, and the info item gets the rest of the window below the map.

exxonvaldez (changes requested):

frontend/src/screens/Dashboard.jsx
zachzwy: When the window width is between 600px ~ 640px, the map is squished.
Screen Shot 2019-12-09 at 9 12 34 PM

exxonvaldez: > When the window width is between 600px ~ 640px, the map is squished.

Yeah, there's some resize logic in MapSpider and MapStops where it jumps down in height from 100% to a fixed height for windows below a certain width.  It doesn't exactly match up with the Material UI breakpoints, but should be easy to fix since you can get the breakpoint values from the MUI theme object.  It's a separate issue.

hathix:
Seanny, do you need any help finishing up this PR?

SeannyPhoenix:
I'll work on this in the coming week; if I hit a wall, I'll reach out @hathix.

hathix:
Just resolved some merge conflicts so it's up to date

PR #449, opened 2019-11-22
UI proposal: hide numeric scores and star ratings; vary text color on route table

youngj:
See http://opentransit-youngj.herokuapp.com/

image

image

image

I think that showing numeric scores on an absolute scale gives too much emphasis to the preferences  of whoever defines the scoring algorithm. Each user may have different opinions on what metrics are important to them about the quality of a route, such as median wait time, speed, variability, etc.  Different riders may have different opinions about whether a route that is scheduled with long headways and is always on schedule is better or worse than a route that arrives inconsistently with shorter headways. Also, different routes and transit modes have different rider expectations.

If I was showing OpenTransit to someone who works at a transit agency, I wouldn't be able to justify the current methodology for scoring routes and trips. Instead of claiming expertise at knowing what is "good" and what is "bad", I think it would be better to stick to presenting basic metrics that users can intuitively understand without a long explanation. Then users can focus on the metrics that matter to them, and avoid being distracted by trying to understand whether our scoring algorithm matches what they actually care about.

Similar to the numeric scores, I think that the star ratings place too much confidence in our ability to tell what is good and what is bad.

Using varying colors to highlight possible good/bad metrics seems helpful, even if the colors are defined on an absolute scale (although we would probably need to customize scales for different agencies with different ranges of wait times, speeds, etc.). For example it seems less controversial to show a metric in yellow than to show something like 1.5 / 5 stars.

I think relative rankings of metrics across routes are helpful to put metrics in context, without needing to make our own judgement of how good or bad a metric is.

I like varying text colors for metrics in the route table to allow the user to easily scan the table for possible good or bad numbers without needing to parse each number, while being less visually overwhelming than changing the cell background color.

Adding units to all metrics in the route table seems to make it easier to scan the table without needing to remember the associated header of that column.

Hiding scores also reduces the number of numbers on the route table and route/trip summary, making the app somewhat less overwhelming.

### DIFF
--- a/frontend/src/components/InfoScoreCard.jsx
+++ b/frontend/src/components/InfoScoreCard.jsx
@@ -79,7 +79,7 @@ export default function InfoScoreCard(props) {
           justifyContent="flex-start"
           height="100%"
         >
-          <Typography variant="overline">{title}</Typography>
+          <Typography style={{whiteSpace:'nowrap'}} variant="overline">{title}</Typography>
 
           <Box flexGrow={1}>
             {' '}
@@ -90,9 +90,9 @@ export default function InfoScoreCard(props) {
             <Typography variant="h5" display="inline">
               {smallValue}
             </Typography>
-            {hideRating ? null : (
+            {/*hideRating ? null : (
               <Rating readOnly size="small" value={rating} precision={0.5} />
-            )}
+            )*/}
           </Box>
           <Box
             display="flex"
--- a/frontend/src/components/InfoTripSummary.jsx
+++ b/frontend/src/components/InfoTripSummary.jsx
@@ -37,7 +37,7 @@ import WatchLaterOutlinedIcon from '@material-ui/icons/WatchLaterOutlined';
  * @param {any} props
  */
 export default function InfoTripSummary(props) {
-  
+
   const [typicalAnchorEl, setTypicalAnchorEl] = useState(null);
   const [planningAnchorEl, setPlanningAnchorEl] = useState(null);
 
@@ -56,7 +56,7 @@ export default function InfoTripSummary(props) {
   function handlePlanningClose() {
     setPlanningAnchorEl(null);
   }
-  
+
   const { graphData, graphParams, routes } = props;
   const waitTimes = graphData ? graphData.waitTimes : null;
   const tripTimes = graphData ? graphData.tripTimes : null;
@@ -165,7 +165,7 @@ export default function InfoTripSummary(props) {
     getPercentileValue(tripTimes, PLANNING_PERCENTILE),
   );
   const travelVariability = Math.round(
-    (getPercentileValue(tripTimes, PLANNING_PERCENTILE) -      
+    (getPercentileValue(tripTimes, PLANNING_PERCENTILE) -
      getPercentileValue(tripTimes, TENTH_PERCENTILE)) / 2.0,
   );
 
@@ -222,7 +222,7 @@ export default function InfoTripSummary(props) {
   const popoverContentLongWait = grades ? (
     <Fragment>
       Long wait probability is the chance a rider has of a wait of twenty minutes or
-      longer after arriving randomly at the "from" stop. 
+      longer after arriving randomly at the "from" stop.
       Probability of{' '}
       {(longWaitProbability * 100).toFixed(1) /* be more precise than card */}%
       gets a score of {grades.longWaitScore}.
@@ -286,7 +286,7 @@ export default function InfoTripSummary(props) {
             <Grid container spacing={4}>
               {/* spacing doesn't work exactly right here, just pads the Papers */}
               <Grid item xs component={Paper} className={classes.uncolored}>
-                <Typography variant="overline">Typical journey</Typography>
+                <Typography variant="overline" style={{whiteSpace:'nowrap'}}>Typical journey</Typography>
                 <br />
 
                 <Typography variant="h3" display="inline">
@@ -314,9 +314,9 @@ export default function InfoTripSummary(props) {
                   </IconButton>
                  </Box>
               </Grid>
-              
+
               <Grid item xs component={Paper} className={classes.uncolored}>
-                <Typography variant="overline">Journey planning</Typography>
+                <Typography variant="overline" style={{whiteSpace:'nowrap'}}>Journey planning</Typography>
                 <br />
 
                 <Typography variant="h3" display="inline">
@@ -335,7 +335,7 @@ export default function InfoTripSummary(props) {
                   <Typography variant="body1">
                     <WatchLaterOutlinedIcon fontSize="small" style={{verticalAlign: 'sub'}} />&nbsp;
                     {planningWait} min
-                    <br/> 
+                    <br/>
                     <StartStopIcon fontSize="small" style={{verticalAlign: 'sub'}} />&nbsp;
                     {planningTravel} min
                   </Typography>
@@ -344,7 +344,7 @@ export default function InfoTripSummary(props) {
                   </IconButton>
                 </Box>
               </Grid>
-              <InfoScoreCard
+              {/*<InfoScoreCard
                 grades={grades}
                 gradeName="totalScore"
                 hideRating
@@ -353,7 +353,7 @@ export default function InfoTripSummary(props) {
                 smallValue={`/${grades ? grades.highestPossibleScore : '--'}`}
                 bottomContent="&nbsp;"
                 popoverContent={popoverContentTotalScore}
-              />
+              />*/}
               <InfoScoreCard
                 grades={grades}
                 gradeName="medianWaitScore"
--- a/frontend/src/components/RouteSummary.jsx
+++ b/frontend/src/components/RouteSummary.jsx
@@ -199,7 +199,7 @@ function RouteSummary(props) {
     <Fragment>
       <div style={{ padding: 24 }}>
         <Grid container spacing={4}>
-          <InfoScoreCard
+            {/* <InfoScoreCard
             grades={wait && speed && grades ? grades : null}
             gradeName="totalScore"
             hideRating
@@ -212,7 +212,7 @@ function RouteSummary(props) {
                 : 'No data'
             }
             popoverContent={popoverContentTotalScore}
-          />
+            />*/}
 
           <InfoScoreCard
             grades={wait && grades ? grades : null}
--- a/frontend/src/components/RouteTable.jsx
+++ b/frontend/src/components/RouteTable.jsx
@@ -94,20 +94,20 @@ function getSorting(order, orderBy) {
 
 const headRows = [
   { id: 'title', numeric: false, disablePadding: false, label: 'Name' },
-  { id: 'totalScore', numeric: true, disablePadding: false, label: 'Score' },
-  { id: 'wait', numeric: true, disablePadding: true, label: 'Median Wait (min)' },
+  //{ id: 'totalScore', numeric: true, disablePadding: false, label: 'Score' },
+  { id: 'wait', numeric: true, disablePadding: true, label: 'Median Wait' },
   {
     id: 'longWait',
     numeric: true,
     disablePadding: true,
     label: 'Long Wait %',
   },
-  { id: 'speed', numeric: true, disablePadding: true, label: 'Average Speed (mph)' },
+  { id: 'speed', numeric: true, disablePadding: true, label: 'Average Speed' },
   {
     id: 'variability',
     numeric: true,
     disablePadding: true,
-    label: 'Travel Time Variability (min)',
+    label: 'Travel Time Variability',
   },
 ];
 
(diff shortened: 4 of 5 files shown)


exxonvaldez:
Thanks for the writing up a detailed case for removing scores.  I am somewhat attached to scores, since I joined OpenTransit to add scoring.   The most important aspect of the spec  ( https://docs.google.com/document/d/19eMnShDf8f6Qitq175uYil7StwzSkuVXSNDUfyeVCMw/edit# )
is that it's adapted from the Transit Capacity and Quality of Service Manual ( http://www.trb.org/Main/Blurbs/169437.aspxl ) by the Transportation Research Board, part of the National Academies of Sciences, Engineering, and Medicine (Washington, DC), plus other ideas from the SF Transit Rider / American Planning Association Data Jam event.  Our score weighs all metrics equally, and we can add better ones and remove worse ones as we're able to, and in response to user feedback (like Tal's suggestions from the SFMTA for high frequency vs low frequency).

I have demoed OpenTransit for an SFMTA transit planner and an operations manager.  I don't think they cared too much about the score, they liked seeing historical data -- speeds and wait/trip times.  On the issue of whether scoring means 1. we're claiming expertise and judging good/bad, or is it 2. a method for making the data easier to understand and find insights, I'm hoping it's mostly number 2.  Maybe we can brainstorm how to refine the UI rather than removing scoring.

We have many UI design options: replace scores with bars or partially completed rings, use emoji or other icons, and we can make scores less prominent.  We can also add user preferences, like a "Jesse mode" that shows a lot less information, or detailed customization that lets users choose which columns and cards are visible.

The first time I see the Portland routes table on OpenTransit, I'm going to want to sort the routes by score so I can see how it compares with SF Muni and then start drilling down into the differences, and also look at the Portland routes at both ends of the list.

Neel and I were testing the colored route tables yesterday, and sorted by score it's fascinating to see which subscores get lower as you look down through the entire system.

Screen Shot 2019-11-21 at 8 12 00 PM
Screen Shot 2019-11-21 at 8 12 15 PM
Screen Shot 2019-11-21 at 8 12 25 PM

I also think it'll be interesting to find out if the scoring works well across agencies or not, and what issues come up.  I know Jesse is ready to delete the long wait metric whereas I want to upgrade it to on-time performance for infrequent lines and bunching/gapping for frequent lines.

Lastly, I see route scoring as the foundation for system health scoring and our tweetbot.  We could use the iOS 13 Health app as a model ( https://www.macrumors.com/guide/health/ ), where it offers highlights based on recent trends.  The idea here is to surface unusual days or periods of time for an agency (e.g., marathon going on this day), as well as long term trends (Stockton St. reopens, or Van Ness BRT begins).  Neel has pointed out that we don't give users things to investigate, and system health trends could be a good way to provide an overview of our data and drill down.

EddyIonescu:
My concern about grading routes is the risk that it wouldn't achieve anything more than making transit agencies look bad and give people yet another excuse to take Uber/Lyft ("my route has a C"). For example, Swiftly made themselves look pretty bad when they announced that 53% of TTC vehicles were not on-time, which the TTC got pretty defensive about and ended up trying to discredit them - https://www.cbc.ca/news/canada/toronto/programs/metromorning/ttc-responds-swiftly-1.3485236

On the other hand, I think there's a lot of potential in starting a dialogue on what would be needed to actually improve those routes, which the Bus Turnaround Project seemed to have done a pretty good job at - http://busturnaround.nyc/report-cards/, as the MTA CEO credited them when announcing the Bus Action Plan in 2018 - https://www.citylab.com/transportation/2018/05/how-riders-won-the-fight-for-better-buses-in-new-york-city/559901/

To make grades successful I think we should be really clear on what they mean and who they're for. As they're aimed at transit riders & advocates, making the grades reflect the customer experience of relying on that route seems like the best option - which would definitely work well with a Tweetbot alerting transit riders about the health of their route. As Terence mentioned, transit agency staff would ignore them as they have their own (simpler) internal metrics that are intended to reflect operations, like % of service delivered, % late pull-outs/pull-ins, average speeds, % vehicles gapped/bunched, etc, but they can't discredit the grades as long as they're clearly explained and transparently determined.

As for "claiming expertise", we're all regular transit users and so we know the difference between good and poor service. I wouldn't be against having a "transit-agency" mode that shows more detailed metrics, including ones they care most about.

As for the stars, I think they're good for seeing how a route compares to others within the network, which would also be good for starting a conversation on how to improve "the stars".

Also, for % of long-waits, as to better distinguish between routes that are scheduled to be infrequent (such as the 23 after 7pm and on weekends, which as a score of ~30% as it's every 30 minutes) vs. have service gaps due to operational issues, what are your thoughts on using 15 minutes and scheduled headway + 5, whichever's greater? This way if a route is infrequent, it means that a trip wasn't on-time, while for frequent routes, it's measuring a service-gap.

> "The first time I see the Portland routes table on OpenTransit, I'm going to want to sort the routes by score so I can see how it compares with SF Muni and then start drilling down into the differences, and also look at the Portland routes at both ends of the list."
 - same but for the TTC. They have some really long suburban routes that operate at ridiculous frequencies (literally every 2 minutes at rush hour and every 5-10 from 6am to 2am) yet suffer from awful bunching/gapping and overcrowding due to operating in mixed-traffic. The variability's gonna be huge - which with the map, would really make the case for transit improvements!

And did they put rocket boosters on the 29 Sunset?

exxonvaldez:
Interesting story links, thanks for posting those. 

> Also, for % of long-waits, as to better distinguish between routes that are scheduled to be infrequent (such as the 23 after 7pm and on weekends, which as a score of ~30% as it's every 30 minutes) vs. have service gaps due to operational issues, what are your thoughts on using 15 minutes and scheduled headway + 5, whichever's greater? This way if a route is infrequent, it means that a trip wasn't on-time, while for frequent routes, it's measuring a service-gap.

Headway + 5 is one long wait scheme mentioned in the service gaps section of https://docs.google.com/document/d/19eMnShDf8f6Qitq175uYil7StwzSkuVXSNDUfyeVCMw/edit#heading=h.51k1di46e2h6 , plus a few other variants (2 * scheduled headway for a stop at that particular time of day and 2 * average headway).  @jtanquil added collection of schedule data a couple of months ago, but that data is not used yet.  Once we're able to compare scheduled arrivals with actual arrivals, we can do both on-time performance and gap %age or a blended metric like you suggested... all kinds of possibilities.

akgupta89:
[quote]My concern about grading routes is the risk that it wouldn't achieve anything more than making transit agencies look bad and give people yet another excuse to take Uber/Lyft ("my route has a C"). For example, Swiftly made themselves look pretty bad when they announced that 53% of TTC vehicles were not on-time, which the TTC got pretty defensive about and ended up trying to discredit them - https://www.cbc.ca/news/canada/toronto/programs/metromorning/ttc-responds-swiftly-1.3485236[/quote]

I partially disagree with this because

1. We're already biased how our route performs based on our own prior experience (whether it's slow or fast, how often it arrives, and whether there's bunching, crowding, etc). I take one of the best lines in the city, and I know it out performs just about every other route in the city based on personal experience, and the app confirmed my intuition.
2. I don't think people will switch off to taking ubers/lyfts just because a metrics app says their route is terrible, it's too cost prohibitive
3. If someone is taking a different route and trying a new bus, this gives them an idea of what it'll be like, just the same as Google Maps gives time estimates to get from A to B for buses and ubers
4. I think this is the same as Yelp or Google Maps and rating your favorite pizza places in the city, one is subjective, and the other is metrics based. Regardless, scoring things is still going to happen.

I think to push for change in the transit industry, there has to be controversial features that agencies may or may not like. While there are no easy solutions to fixing public transportation, rather than saying a route is terrible, we give metrics why a route is underperforming, which is far more useful to transit agencies.

---

That being said, there is something to be said about the algorithm. I think a good base case to model scoring off is against uber/lyft, as that is primary competition, and if a route can take you to the same locations about the same time as private transportation I'd consider the route to be quite good. I'm against removing the scoring system altogether, because for the average person who isn't transit literate, this would make the application far less useful.

hathix:
Good arguments on both sides! My take is that users (especially those without much transit expertise) may not know what exactly makes a route good. Plus, people like having a clear conception of what's "best" and "worst." So it's good to show scores in some form.

However, the point stands that the criteria for our scoring (while pretty scientifically determined) might not be immediately clear. I'd propose adding help text to the scores column -- a popover, info bubble, modal when the score is clicked, etc. -- to make our criteria more transparent.

We could also let users adjust the scoring criteria (change how much each score is weighted, change the how the value of a metric maps to its point value, etc.) in case people don't like our score-calculation algorithm.

youngj:
Thanks for the feedback everyone.

I think I like Terence's prototype with varying background colors in an oval better than varying the text colors in this PR. I also like going in the the direction of using the same style for Score as other columns, instead of emphasizing the score more than the other columns.

Other things that might help if keeping the score: 

* Changing the name from "Score" to something else -- maybe something like Level of Service, Quality of Service, Overall Quality, etc.

* Changing the scale. The 0-100 scale is familiar to everyone from grades in school, but also means people will associate scores with the idea that >=90 is an "A" and 59 or below is an "F". A score of 59 or below could imply that the transit agency is failing. But a 59 or below could actually mean that the transit agency is performing exactly as scheduled. How about changing it to a 0.0 to 10.0 scale which might have less existing meaning associated with it?

* Displaying the score as the last column instead of the first column to suggest that it is derived from the other metrics.

* Making it possible to configure the metrics shown in the web app (e.g. via environment variables passed from Flask to the JS config). Different people have different ideas about how they envision OpenTransit being used, but currently we only support a one-size-fits-all UI.

* Changing the scoring thresholds to increase the domain and avoid giving zero or perfect scores in so many situations. Currently computeGrades assigns a zero score to routes with >10 minute average wait times, >33% probability of 20 minute wait time, <5 mph average speed, and >10 minute end-to-end variability. Anything worse than these numbers doesn't affect the score at all since any subscore can't be worse than zero. Conversely, computeGrades assigns a perfect score to a route with <5 minute average wait times, <10% probability of 20 minute wait time, >10 mph average speed, and <5 minute end-to-end variability. Anything better than these numbers also doesn't affect the score at all since any subscore can't be better than 100. It would be difficult to come up with thresholds that make sense in all situations because different transportation modes and transit systems will have different expectations for various metrics. Maybe instead of having a piecewise linear scoring scale with arbitrary cutoffs that correspond to zero or perfect scores, the scoring function could use a sigmoid curve (https://en.wikipedia.org/wiki/Sigmoid_function).

akgupta89:
I think we shouldn't focus too much on the computation score right now, I have a feeling that Terence designed it with SF in mind, and this doesn't translate to other cities quite as fairly, so in that sense we're going to have to be adjusting it anyways once we expand to other cities where buses travel faster and longer (think the midwest/south).

The other names suggested to me seem more official than score, so maybe we should call it "OpenTransit Score", "OpenTransit Grading", "OpenTransit Rating"  to be implied subjective and based on our criteria.

hathix:
Comments to various proposals in line.

> I also like going in the the direction of using the same style for Score as other columns, instead of emphasizing the score more than the other columns.

I like that idea.

> Changing the name from "Score" to something else -- maybe something like Level of Service, Quality of Service, Overall Quality, etc.

Arun makes a good point that we should emphasize that this is subjective. In the interest of brevity, something like "OT Rating" (out of 10) or "Quality Estimate" might help.

> How about changing it to a 0.0 to 10.0 scale which might have less existing meaning associated with it?

Good idea. I think people are more used to grading things out of 10 than out of 100. (You can probably rate Code for SF's pizza on a scale of 1-10, but rating it from 1-100 would be weird.)

> Displaying the score as the last column instead of the first column to suggest that it is derived from the other metrics.

The score is intended to be a top-level overview (and a common thing to sort by), so I'd suggest keeping it on the leftmost (first) column.

> Making it possible to configure the metrics shown in the web app (e.g. via environment variables passed from Flask to the JS config). Different people have different ideas about how they envision OpenTransit being used, but currently we only support a one-size-fits-all UI.

Perhaps each agency's subdomain could specify its own weights/buckets. This way TriMet could get scored on a different scale that's more tailored to its local context.

> Changing the scoring thresholds to increase the domain and avoid giving zero or perfect scores in so many situations... Maybe instead of having a piecewise linear scoring scale with arbitrary cutoffs that correspond to zero or perfect scores, the scoring function could use a sigmoid curve.

I agree that our thresholds are a bit weird and support changing them. But keeping them linear makes them easier to understand.

>  I have a feeling that Terence designed it with SF in mind, and this doesn't translate to other cities quite as fairly, so in that sense we're going to have to be adjusting it anyways once we expand to other cities where buses travel faster and longer (think the midwest/south).

Ooh, this brings up an interesting dilemma: do you standardize route scores across cities or not? If you do standardize, people can more easily see the variation within their city, and it's arguably fairer since each city has its own constraints and context. But if you don't standardize, people can compare across cities more easily. Not sure what's a better alternative here. Regardless, I think it would help for us to support customizing the scoring algorithm for each city.

PR #460, opened 2019-11-26
handle loops correctly when displaying list of stops in the UI issue 426

### LINKED ISSUE #426: Handle loops correctly when displaying list of stops in the UI
The Portland Streetcar A Loop and B Loop lines travel in a continuous loop. However, each line has an arbitrary beginning/end stop (which are identical). Currently, the UI only allows selecting a destination stop that appears after the origin stop in the list of stops. 

If the beginning/end stop for a direction are identical, the UI should consider the route to be a loop, and should add the stops that are after the arbitrary beginning/end stop and before the origin stop to the bottom of the destination stop dropdown list.

In addition to the UI change, this will also need backend support at https://github.com/trynmaps/metrics-mvp/issues/425 in order for the statistics to be calculated correctly.

licarpen:
Fixes #426 

* Please provide any feedback as this is my first pull request on an open source project.  

## Screenshot 1: Shows second stops extending past 9th & Lovejoy (this used to be the last stop shown)

Screen Shot 2019-11-25 at 8 59 29 PM

## Screenshot 2: Shows functionality preserved for unidirectional routes

Screen Shot 2019-11-25 at 8 57 04 PM

### DIFF
--- a/frontend/src/components/ControlPanel.jsx
+++ b/frontend/src/components/ControlPanel.jsx
@@ -63,11 +63,12 @@ function ControlPanel(props) {
       mySelectedRoute,
       directionId,
     );
-    const secondStopListIndex = stopId
-      ? secondStopInfo.stops.indexOf(stopId)
-      : 0;
-
+    
     const stopsList = secondStopInfo.stops;
+    const secondStopListIndex = stopId
+      ? stopsList.indexOf(stopId)
+      : 0;
+    
     // loop routes start and stop at same stop
     const isLoopRoute = stopsList[0] === stopsList[stopsList.length - 1];
     const oneWaySecondStopsList = stopsList.slice(secondStopListIndex + 1);
@@ -75,7 +76,7 @@ function ControlPanel(props) {
     if (!isLoopRoute) {
       return oneWaySecondStopsList;
     }
-    // loop routes display all subsequent routes up to origin stop
+    // loop routes display all subsequent stops up to origin stop
     return oneWaySecondStopsList.concat(stopsList.slice(1, secondStopListIndex));
   }
 
--- a/frontend/src/components/ControlPanel.jsx
+++ b/frontend/src/components/ControlPanel.jsx
@@ -70,9 +70,10 @@ function ControlPanel(props) {
 
     const unidirectionalSecondStopsList = secondStopInfo.stops.slice(secondStopListIndex + 1);
 
-    const circularSecondStopsList = unidirectionalSecondStopsList.concat(secondStopInfo.stops.slice(0, secondStopListIndex));
-
-    return isLoopRoute ? circularSecondStopsList : unidirectionalSecondStopsList;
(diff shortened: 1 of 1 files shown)


frontend/src/components/ControlPanel.jsx
```diff
@@ -66,7 +66,13 @@ function ControlPanel(props) {
     const secondStopListIndex = stopId
       ? secondStopInfo.stops.indexOf(stopId)
       : 0;
-    return secondStopInfo.stops.slice(secondStopListIndex + 1);
+    const isLoopRoute = secondStopInfo.stops[0] === secondStopInfo.stops[secondStopInfo.stops.length - 1];
+
+    const unidirectionalSecondStopsList = secondStopInfo.stops.slice(secondStopListIndex + 1);
+
+    const circularSecondStopsList = unidirectionalSecondStopsList.concat(secondStopInfo.stops.slice(0, secondStopListIndex));
```
youngj: Since the 0th stop is also at the end of unidirectionalSecondStopsList, could change the 0 to 1 to avoid listing the same stop twice.

frontend/src/components/ControlPanel.jsx
```diff
@@ -66,7 +66,13 @@ function ControlPanel(props) {
...
+
+    return isLoopRoute ? circularSecondStopsList : unidirectionalSecondStopsList;
```
youngj: Could return unidirectionalSecondStopsList for non-loop routes before computing circularSecondStopsList

frontend/src/components/ControlPanel.jsx
licarpen: Great point.  Made this change.

frontend/src/components/ControlPanel.jsx
licarpen: Great feedback.  Made this change.

licarpen:
Thanks for the valuable feedback.  I made the changes suggested so am assuming the most recent changes need approval before merging.  Both trimet and portland-sc routes still working after latest changes.

exxonvaldez:
Oh, I just thought of something that we might want to file another issue for.  Can you test this out?  On the map, we highlight the path between selected the “from” stop and the “to” stop in indigo.  There is code in MapStops that creates two Polylines between each stop.  The first one is wider and is either white or indigo, depending on whether the code thinks this Polyline is between the from/to stops.  The second line is thinner and indicates the speed.

I’m pretty sure this is a simple traversal where the code has a boolean flag that says whether or not we have seen the “from” stop, and maybe another flag for having seen the  “to” stop.  We’ll need different logic for loops.

If this is the case, you can click the “...” menu for this comment and “reference in new issue” to create the issue, and then either pick it up or leave it for others.

licarpen:
Thanks for the info - I will look into the highlighting issue for loop routes.  I appreciate your patience and taking the time to provide feedback.

PR #462, opened 2019-11-26
Adjust vertical alignment of date button dropdown icon and map shield…

### LINKED ISSUE #236: Minor Material UI issues
There are a bunch of small edits we need to make to the UI now that we ported the UI to use exclusively Material. These are mostly on the dashboard:

* Fix alignment of the text & dropdown button in the gray date-picker bar
* Add padding to the Routes Table and make the links not blue
* Make route number properly centered in the bubbles on the spider map 

Screenshot 2019-08-07 at 9 17 04 PM

exxonvaldez:
… text.

Fixes #236.

## Proposed changes

- Adjust vertical alignment of date button dropdown icon.
- Adjust vertical alignment of map shield text, and bump up radius to get rounder shields.

No fixes needed:
- Misaligned "Date-time range" text has been removed.
- Route table borders on left have been removed, so route name is not against a border and no padding needed.  Table is being reworked in #437 also.
- Route naes are now in the primary color (dark indigo).

## Screenshot

Screen Shot 2019-11-26 at 11 53 24 AM

## Link to demo, if any

n/a

### DIFF
--- a/frontend/src/components/DateTimePanel.jsx
+++ b/frontend/src/components/DateTimePanel.jsx
@@ -338,7 +338,7 @@ function DateTimePanel(props) {
           </Typography>
           <Typography className={classes.secondaryHeading} display="inline">
             {timeLabel}
-            <ExpandMoreIcon />
+            <ExpandMoreIcon style={{verticalAlign: '-20%'}} />
           </Typography>
         </div>
       </Button>
--- a/frontend/src/components/DateTimePanel.jsx
+++ b/frontend/src/components/DateTimePanel.jsx
@@ -50,8 +50,9 @@ const useStyles = makeStyles(theme => ({
   column: {
     flexGrow: '1',
   },
-  nowrap: {
+  dateTime: {
     whiteSpace: 'nowrap',
+    display: 'flex',
   },
   root: {
     display: 'flex',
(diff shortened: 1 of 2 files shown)


frontend/src/components/DateTimePanel.jsx
```diff
@@ -338,7 +338,7 @@ function DateTimePanel(props) {
           </Typography>
           <Typography className={classes.secondaryHeading} display="inline">
             {timeLabel}
-            <ExpandMoreIcon />
+            <ExpandMoreIcon style={{verticalAlign: '-20%'}} />
```
akgupta89: There should be a better way to fix alignment than using hack fixes like this. I'll look it up when I have some free time.

exxonvaldez: Google search says to use vertical-align: middle on both the icon and the text.  In this particular situation, though, we want to keep the text baseline-aligned since there are two font sizes involved.  So it comes out like this:

Screen Shot 2019-12-02 at 4 31 12 PM

exxonvaldez: Screen Shot 2019-12-02 at 4 47 47 PM

The shields are just bordered divs containing text (wrapped in a span).  In this particular case, `vertical-align: middle` is better than the default but doesn't quite center the text either.

akgupta89: https://blog.logrocket.com/13-ways-to-vertical-center-in-2018-cb6e98ed8a40/

We should be okay with not perfectly aligned designs because I think the design will evolve more as we add in more features and abilities, as well as design for all devices.

exxonvaldez: I went back to the Stack Overflow page that said to use vertical-align, and then looked at the second answer, which suggested three ways, of which using flexbox sounded the best to me:  https://stackoverflow.com/questions/43360526/align-material-icon-vertically .  Flexbox is something we already use, and was a big part of Bootstrap layout.

For the date-time button, I wrapped the two pieces of text in a span, put the icon next to that, and set the parent to display:flex.  That was sufficient, the default of align-items is stretch, which gets the icon where it needs to be relative to the text, since they share the same height.

Screen Shot 2019-12-02 at 9 51 20 PM

For the shield, I used the Stack Overflow example and needed both align-items: center and justify-content: center.

Screen Shot 2019-12-02 at 9 51 30 PM

hathix:
I like the flexbox approach. Arun, can you approve if you're happy with it?

PR #463, opened 2019-11-27
Speed calculation fixes

### LINKED ISSUE #407: Speed calculations should be based on distance between stops used in trip time calculations
The function getSpeedAndVariabilityForRoute calculates "median speed" by dividing `distance.direction` by the median trip time between two stops at or near the ends of the route. However, the distance between these two stops may be smaller than `distance.direction`, resulting in incorrect speed calculations.

### LINKED ISSUE #439: Fix "infinite" speed appearing
For some reason, sometimes the speed is shown as infinite. 

Screen Shot 2019-11-20 at 8 29 52 PM

It might be caused by having just 1 stop in the dataset:

Screen Shot 2019-11-20 at 8 31 11 PM

exxonvaldez:
Fixes #439 and supersonic 29-Sunset.  Fixes #407.  

## Proposed changes

- muni.yaml: Updated muni.yaml for GTFS stop ids and E/F temporary turnaround at pier 39.  Fix for supersonic 29 sunset speed by ignoring starting inbound stops that are too close (tight "U"). (#439)
- MapStops: Minor layout fix for map height to compensate for new breadcrumb area (from #415)
- TravelTimeChart and routeCalculations:  compute route speed with more accurate distance based on starting and ending stops as outlined in #407.

## Screenshot

Improved speed for E and F on the left:

Screen Shot 2019-11-26 at 6 16 11 PM

Improved speed for 29 and also routes that double back, like 36 and 39, on the left:

Screen Shot 2019-11-26 at 6 18 08 PM

Route summary layout fix on left, map not cut off vs cut off map and scroll on right:

Screen Shot 2019-11-26 at 6 24 45 PM

## Link to demo, if any

https://ot-terence.herokuapp.com/

### DIFF
--- a/backend/agencies/muni.yaml
+++ b/backend/agencies/muni.yaml
@@ -248,6 +248,22 @@ js_properties:
   #   - The 9 has multiple terminals so use the last common stop.
   #   - The 5 was reconfigured and Nextbus stop configs are out of sync with historic data.  Use last good stop.
   routeHeuristics: {
+      E: {
+        '0': {
+          ignoreFirstStop: '13095', # use Beach and Stockton as outbound start terminal until ~Oct 2020.
+        },
+        '1': {
+          ignoreLastStop: '14530', # use Embarcadero and Stockton as inbound end terminal until ~Oct 2020.
+        }
+      },
+      F: {
+        '0': {
+          ignoreFirstStop: '13095', # use Beach and Stockton as outbound start terminal until ~Oct 2020.
+        },
+        '1': {
+          ignoreLastStop: '14530', # use Embarcadero and Stockton as inbound end terminal until ~Oct 2020.
+        }
+      },
       N: {
         '0': {
           ignoreFirstStop: true, # 4th and King to 2nd and King trip times are skewed by a few hyperlong trips
@@ -256,18 +272,18 @@ js_properties:
       S: {
         ignoreRoute: true,
       },
-      '5': {
-        '1': {
-          ignoreFirstStop: '4218', # no data for 3927, and first few stop ids are now different.  Problem is even worse on outbound side, no good fix there.
-        },
-      },
       '9': {
         '1': {
-          ignoreFirstStop: '7297', # use Bayshore as actual first stop (daytime)
+          ignoreFirstStop: '17227', # use Bayshore and Visitacion as actual first stop (daytime)
         },
         '0': {
-          ignoreLastStop: '7297', # use Bayshore as actual terminal (daytime)
+          ignoreLastStop: '13783', # use Bayshore and Leland as actual terminal (daytime)
         },
+      },
+      '29': {
+        '1': {
+          ignoreFirstStop: '14783', # use Gilman and 3rd St as first inbound stop, almost doubling back
+        }
       },
       '90': {
         ignoreRoute: true,
--- a/backend/agencies/muni.yaml
+++ b/backend/agencies/muni.yaml
@@ -246,7 +246,6 @@ js_properties:
   #     should be used instead for end-to-end calculations.  Cable car lines are like this.
   # - Possibly special handling for routes with back end issues (currently 5, 9, 9R) as a temporary workaround.
   #   - The 9 has multiple terminals so use the last common stop.
-  #   - The 5 was reconfigured and Nextbus stop configs are out of sync with historic data.  Use last good stop.
   routeHeuristics: {
       E: {
         '0': {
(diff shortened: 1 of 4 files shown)


frontend/src/helpers/routeCalculations.js
```diff
@@ -77,7 +77,7 @@ function getTripTimesUsingHeuristics(
   if (!tripTimesForDir || !routes) {
     //console.log("No trip times found at all for " + directionId + " (gtfs out of sync or route not running)");
     // not sure if we should remap to normal terminal
-    return { tripTimesForFirstStop: null, directionInfo: null };
+    return { tripTimesForFirstStop: null, directionInfo: null, distanceAlongRoute: null };
```
akgupta89: I see `distanceAlongRoute` being created but used nowhere and cant find a use reference to it in the codebase.

exxonvaldez: Good catch, I renamed this return value but forgot to update it here (in this error scenario, the distance isn't actually used anyway).  Fixed.

frontend/src/helpers/routeCalculations.js
```diff
@@ -177,21 +178,33 @@ export function getEndToEndTripTime(
...
   }
 
   // if there is no trip time to the last stop, then use the highest trip time actually observed
 
   if (!tripTime) {
-    tripTime = Math.max(
-      ...Object.values(getTripTimeStat(tripTimesForFirstStop, statIndex)),
-    );
-    // console.log("No trip time found for " + routeId + " " + directionId + " to stop " + lastStop + '. max observed: ' + tripTime);
+    const obj = getTripTimeStat(tripTimesForFirstStop, statIndex);
```
akgupta89: This could be named better

backend/agencies/muni.yaml
```diff
@@ -248,6 +248,22 @@ js_properties:
   #   - The 9 has multiple terminals so use the last common stop.
   #   - The 5 was reconfigured and Nextbus stop configs are out of sync with historic data.  Use last good stop.
   routeHeuristics: {
+      E: {
+        '0': {
+          ignoreFirstStop: '13095', # use Beach and Stockton as outbound start terminal until ~Oct 2020.
```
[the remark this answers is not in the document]
exxonvaldez: It's a little bit convoluted right now... true means we calculate speed using trips starting from the 2nd stop instead of the normal first.  String id means we calculate speed using starting trips from the specified stop.  So we need both an ignoreFirstStop and ignoreLastStop.  But it was thrown together very ad hoc so good to think about a better structure or names.

frontend/src/helpers/routeCalculations.js
```diff
@@ -177,21 +178,33 @@ export function getEndToEndTripTime(
...
+    const keys = Object.keys(obj);
+    tripTime = 0;
+
+    for (let i = 0; i < keys.length; i++) {
```
akgupta89: This is completely subjective and more of a nitpick but I think gravitating towards using ES6 added functionality makes code more concise for readability, e.g. 

```
Object.entries(obj).filter([, value] => value > tripTime).forEach([key, value] => {
        tripTime = value;
        tripDistance = directionInfo.stop_geometry[key].distance - firstStopDistance;
});
```

Key and value could be better descriptions for route name? and time?. The question is if this operation is `O(N)` vs `O(2N)` I should think is valid. I can't find an answer to this for JS, except that in Java this would be a N operation. 🤷‍♂

frontend/src/helpers/routeCalculations.js
exxonvaldez: Fixed.

frontend/src/helpers/routeCalculations.js
exxonvaldez: Good thought re ES6, but that wouldn't be the right code.  Given a hash of stop ids to trip times, the original code set tripTime to the Math.max() of all the trip times (Object.values(hash)).  The new code needs to do that and also get the distance along the route, so it needs the stop id that corresponds to the max trip time.

So step one would be to use reduce (not filter) to get the single hash entry with the highest trip time.  Step two would be to then set the tripTime and tripDistance.  It'd be like this:

    const tripTimes = getTripTimeStat(tripTimesForFirstStop, statIndex);
    const highestEntry = Object.entries(tripTimes).reduce((highestTripEntry, currentTripEntry) =>
       currentTripEntry[1] > highestTripEntry[1] ? currentTripEntry : highestTripEntry,
       [0,0]);
    let stopId;
    [ stopId, tripTime ] = highestEntry;
    tripDistance = directionInfo.stop_geometry[stopId].distance - firstStopDistance;

and here's my PR code with the updated variable names:

    const tripTimes = getTripTimeStat(tripTimesForFirstStop, statIndex);
    const stopIds = Object.keys(tripTimes);
    tripTime = 0;

    for (let i = 0; i < stopIds.length; i++) {
      if (tripTimes[stopIds[i]] > tripTime) {
        tripTime = tripTimes[stopIds[i]];
        tripDistance = directionInfo.stop_geometry[stopIds[i]].distance - firstStopDistance;
      }
    }

Which do you prefer?  I don't think either is going to look that great given the messy operation we are doing here.

Issue #457, opened 2019-11-25
Indicate route type in web UI


youngj:
The route object in JS now contains an integer "type" property, with the same value as GTFS route_type field in routes.txt:

```
0 - Tram, Streetcar, Light rail. Any light rail or street level system within a metropolitan area.
1 - Subway, Metro. Any underground rail system within a metropolitan area.
2 - Rail. Used for intercity or long-distance travel.
3 - Bus. Used for short- and long-distance bus routes.
4 - Ferry. Used for short- and long-distance boat service.
5 - Cable car. Used for street-level cable cars where the cable runs beneath the car.
6 - Gondola, Suspended cable car. Typically used for aerial cable cars where the car is suspended from the cable.
7 - Funicular. Any rail system designed for steep inclines.
```

It would be useful to indicate the route type in the web UI somewhere (icons?)

hathix:
There are actually a lot of emojis we could use.

Tram: 🚊🚋🚈
Subway: 🚇
Rail: 🚅🚄🚂
Bus: 🚌🚍
Ferry: ⛴️🚢
Cable car: 🚎🚈
Gondola: 🚠🚡
Funicular: 🚞

Perhaps we could put these emojis right before or after the route name on the dashboard... Or in their own column. And on the routes page we could put the emoji in the breadcrumbs, or even replace the standard bus icon on the map with these emojis.

hathix:
Some customization per city may be needed. SF distinguishes motor coaches (buses without a connection to an overhead rail) and other coaches (forget the name but they are connected to an overhead rail). And there's historic street cars.

exxonvaldez:
> Some customization per city may be needed. SF distinguishes motor coaches (buses without a connection to an overhead rail) and other coaches (forget the name but they are connected to an overhead rail). And there's historic street cars.

Trolley coach.  During the Potrero yard tour, they said something about how San Francisco was very unusual for having such extensive overhead wires in the US, as most cities had removed their overhead wires along with their streetcar rails during the conversion to buses.  I think it’s because of the steepness of SF hills.  Custom icons sounds like a fun addition, as does the use of emojis.

youngj:
The GTFS files for SF Muni use type 3 for both motor coaches and trolley coaches. SF Muni uses the types 0, 3, and 5 (Muni Metro lines don't count as subway). 

In Portland we have an aerial tram with type 6 although we're not collecting data for it yet.

SeannyPhoenix:
[...]

PR #468, opened 2019-12-03
Draft: Emoji to represent GTFS route type

exxonvaldez:
This a partial prototype for #457, show route type in UI, to get early feedback and to hand over if anybody else is interested in taking over.

## Proposed changes

Using native (UTF-8) emojis, wrapped in a span for accessibility.  Created constants to represent emoji symbol and label by GTFS route type.

Added emoji to route names in the route table, and in the bread crumb, just to see how it looks.  Alternate emojis are noted in the source code.

## Screenshot

Route table:

Screen Shot 2019-12-02 at 11 15 31 PM

Stop to stop:

Screen Shot 2019-12-02 at 11 16 29 PM

(Could add emojis to route picker, "from stop" dropdown label, and map, depending on how far we want to take it).

## Link to demo, if any

n/a

### DIFF
--- a/frontend/src/UIConstants.js
+++ b/frontend/src/UIConstants.js
@@ -85,3 +85,26 @@ export const WEEKENDS = [
 
 // Marey chart:  how long of a dwell at a stop results in a second data point for exit.
 export const DWELL_THRESHOLD_SECS = 120;
+
+/*
+ * Route type emojis.
+ *
+ * 0 - Tram, Streetcar, Light rail. Any light rail or street level system within a metropolitan area.
+ * 1 - Subway, Metro. Any underground rail system within a metropolitan area.
+ * 2 - Rail. Used for intercity or long-distance travel.
+ * 3 - Bus. Used for short- and long-distance bus routes.
+ * 4 - Ferry. Used for short- and long-distance boat service.
+ * 5 - Cable car. Used for street-level cable cars where the cable runs beneath the car.
+ * 6 - Gondola, Suspended cable car. Typically used for aerial cable cars where the car is suspended from the cable.
+ * 7 - Funicular. Any rail system designed for steep inclines.
+ */
+export const ROUTE_TYPE_EMOJIS = {
+  '0': { symbol: '🚈', label: 'Light Rail'}, /* light rail, or railway car: 🚃, or tram: 🚊, tram car: 🚋  */
+  '1': { symbol: '🚇', label: 'Subway'}, /* metro */
+  '2': { symbol: '🚆', label: 'Rail'}, /* train, or bullet train: 🚅, high-speed train: 🚄, locomotive: 🚂 */
+  '3': { symbol: '🚌', label: 'Bus'}, /* bus, or oncoming bus: 🚍, trolley bus: 🚎 */
+  '4': { symbol: '⛴', label: 'Ferry'}, /* ferry, or ship: ️🚢 */
+  '5': { symbol: '🚋', label: 'Cable Car'}, /* tram car (nothing better available) */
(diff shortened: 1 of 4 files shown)


hathix:
Looks cool! Emojis add a pop of color and fun, which I think the app really benefits from. I like the idea of adding the emoji to the map, though then we'd probably want to replace all instances of that generic train icon with an emoji... which might get too busy. So I think your current approach is a good balance

youngj:
Emojis are different on each OS (e.g. https://emojipedia.org/bus/), so the UI would be more consistent if we used our own icons. Also I think the colors are too distracting especially when repeated for each route in the table. Single color icons would probably look cleaner and more professional.

exxonvaldez:
Here's some options for implementing emoji or icons:

- Native emoji is the easiest to implement but varies by client platform.

- This is an emoji picker component that seems to have the emoji images for 'apple', 'google', 'twitter', 'emojione', 'messenger', 'facebook' if we preferred the emoji for a specific platform (not sure if this is legal redistribution/reuse):

https://github.com/missive/emoji-mart

- We can search through the Material UI icons, these are already monochrome and available in several styles per icon (solid, outlined, etc.):

https://material-ui.com/components/material-icons/

- We can also look on Google for free icons:

https://www.google.com/search?q=bus+train+icons+free

I think the next best/easiest iteration to try would be Material UI icons in the route table (and if successful, then see how they look in the breadcrumbs and map).

hathix:
[MUI's icons](https://material-ui.com/components/material-icons/) are a good option too. A quick rundown of which icons it supports, for the record:

Tram: yes, `tram`
Subway: yes, `subway`
Rail: yes, `train`
Bus: yes, `directionsbus`
Ferry: yes, `directionsboat`
Cable car:  no
Gondola: no
Funicular: no

I think this is emblematic of the broader problem of using anything besides emojis: it'll be hard to get icons for that long tail of transport modes.

Using random icons on Google might lead to a mishmash of different styles, and the images may not be high quality.

Another option is Font Awesome, but it also lacks cable cars, gondolas, and funiculars.

We could also use Mozilla's or OpenMoji's emoji pictures ([example](https://emojipedia.org/aerial-tramway/)), which would probably be open source.

hathix:
I would say that for a "minimum viable product" it's best to go with the good-enough, platform-specific emojis. We can add a fancier alternative later.

exxonvaldez:
Re: open source emojis, good point.  A few more emoji sets that are open:  Google and Twitter (https://blog.emojipedia.org/who-owns-emoji/ )

hathix:
The Google one is more like an international font than an emoji pack. [Twemoji](https://twemoji.twitter.com/) is super nice though, and very recognizable. And it offers SVGs. We can use that.

hathix:
Oh man, you can programmatically create Twemoji! That's a great feature.

zachzwy:
I like @youngj 's idea of single color emoji, considering there are already a few different colors on the current route table. We don't want the route's emoji color to mis-indicate its quality.

Maybe all black icon?
Screen Shot 2019-12-10 at 9 35 56 PM

exxonvaldez:
I've updated this PR with the emojis commented out and added the Material UI icons @hathix suggested.  The solid icons were too dark, so I switched some to outlined.  Either way, the tram and bus icons are very similar, so the result is not compelling.  It was pretty easy to try out, at least.

Screen Shot 2019-12-11 at 1 19 56 PM

Screen Shot 2019-12-11 at 1 24 22 PM

hathix:
The UI needs some polishing, says Terence. Not ready to merge this. We'd like to get some ui people to look at this

exxonvaldez:
Closing this out.  This PR can still be found easily via the open issue #457.

PR #469, opened 2019-12-04
Msmith/399 return to dashboard

### LINKED ISSUE #399: Keep the "Clear Map" button on the route screen
Once you've clicked on a route, it's quite difficult to get back to the original map, making misclicks pretty costly.

I think #160 will make this less of an issue, but I still think the "Clear Map" button has value when you've chosen a route.

melismae:
## Proposed changes

This PR proposes to add a "Return to Dashboard" link on the route map page. 

Fixes #399 

Would like feedback suggested path on alternatives to using the `<br />` below the button wrapper on the map (this mimics the code in MapSpider). Is hacky for now. Shared component with styling? Or is this ok for right now?

...

## Screenshot
Screen Shot 2019-12-03 at 9 16 48 PM

...

## Link to demo, if any
https://trynmaps-melissa.herokuapp.com/
...


exxonvaldez:
I was thinking about this, the convention seems to be a back arrow button in the upper left corner.  For us, maybe in the left edge of the white breadcrumb area?

melismae:
Yeah I like that a lot better. The button while consistent gets lost at the bottom of the map. good suggestion!

exxonvaldez:
Cool... I was looking at Google Photos and Google Docs just now.  Both of them have a hamburger menu in the upper left.   Then when you drill into something, they replace the hamburger with a back arrow (Photos) or a home button (Docs uses an icon represent docs, as opposed to sheets/slides).

hathix:
I like the idea of replacing the hamburger with a back button when you're "drilled into" a page like the routes page. That'll make the UI more consistent and easy to use on phones (something we'll have to start thinking about).

I would say it makes the most sense to put this back button in the blue header bar, since that's closest to the corner of the page. Plus, the breadcrumbs will already have a lot of links to other pages, and adding a back button there might make it too busy.

akgupta89:
Shouldn't the breadcrumb do this? Maybe a giant home icon next to the first entry in the breadcrumb should be sufficient I hope.

hathix:
Let's make sure we give this a formal review...

hathix:
This is a false start, so let's start over with the new approach. See #399 for comments.

hathix:
I have some new comments about the new direction in #399 (see the post with screenshot).

melismae:
Closing out in favor of a new PR!

PR #479, opened 2019-12-05
Fix padding of route table headers (#437 followup)

exxonvaldez:
## Proposed changes

#437 added extra padding to the right of column headers.  This was an adjustment for when the background colors filled the entire cell.  For chips we don't need this padding.

## Screenshot

Screen Shot 2019-12-05 at 12 55 57 AM

## Link to demo, if any

n/a

### DIFF
--- a/frontend/src/components/RouteTable.jsx
+++ b/frontend/src/components/RouteTable.jsx
@@ -25,6 +25,7 @@ import {
   getAllSpeeds,
   getAllScores,
   quartileBackgroundColor,
+  quartileContrastColor,
 } from '../helpers/routeCalculations';
 
 import { handleGraphParams, fetchPrecomputedWaitAndTripData } from '../actions';
@@ -386,6 +387,9 @@ function RouteTable(props) {
                     >
                     <Chip
                       style={{
+                        color: quartileContrastColor(
+                          row.totalScore / 100,
+                        ),
                         backgroundColor: quartileBackgroundColor(
                           row.totalScore / 100,
                         ),
@@ -401,6 +405,9 @@ function RouteTable(props) {
                     >
                     <Chip
                       style={{
+                        color: quartileContrastColor(
+                          row.medianWaitScore / 100
+                        ),
                         backgroundColor: quartileBackgroundColor(
                           row.medianWaitScore / 100
                         ),
@@ -417,6 +424,9 @@ function RouteTable(props) {
                     >
                     <Chip
                       style={{
+                        color: quartileContrastColor(
+                          row.longWaitScore / 100,
+                        ),
                         backgroundColor: quartileBackgroundColor(
                           row.longWaitScore / 100,
                         ),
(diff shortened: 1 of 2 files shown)


youngj:
Although it seems that the SF team agreed in person on the darker color scheme, overriding the comments in the other PR, I still prefer the lighter color scheme because it makes the numbers easier to read. I think the darker color scheme with text in both white and black is a step backwards in terms of usability. The background colors are just a visual aid for more easily scanning the numbers. If some people have trouble distinguishing the background colors, they don't actually lose any important functionality.

exxonvaldez:
These are good points.  There is a tradeoff between prioritizing scanning (overview) and prioritizing details (text values).  I'm wondering if we should separate these displays.  Given the argument that the table has too much information currently, we could make the table primarily about scanning by not showing the numeric values other than score.

The other numbers would be shown on light colored backgrounds via row hover or row select (either within the hovered row, or a non-modal popover just below the row).  This would make it harder to compare numbers two routes (not that this was ever easy in the table), but I'm not sure if this is an important use case.  These same numbers appear on the route summary screen if the user drills in.

I can only mock up the first part quickly, where we just use colors.  The row hover/select is more work.

Screen Shot 2019-12-05 at 2 32 23 PM

youngj:
As you might guess from my comments in https://github.com/trynmaps/metrics-mvp/pull/449, I wouldn't support removing all of the numbers except for the score.

I think that the most important numbers that are relevant and understandable to most riders are Median Wait and Average Speed. 

Once we have backend support for schedule adherence (working on a PR for this coming soon) we can replace Long Wait % with On Time %.

I'd support removing Score (as previously argued), Long Wait %, and Travel Time Variability (end-to-end variability over the entire route is not as relevant for people who commute between particular stops on the route).

hathix:
My $0.02:

* I think we still haven't reached agreement on whether we should keep the score or not. I would say, for the purposes of this PR, to keep the score in, since the score debate should be localized to #449.
* I agree with Jesse that we should show all the numbers. For accessibility's sake, it's not enough to just show the colors -- plus, more detail-oriented users will probably want to see the full numbers, and we don't want to force them to have to click into every single line to see the full numbers. If we feel that strongly about offering a numbers-free option, we can make that a setting in a preferences menu or something.
* Ignoring score, I think the most useful/understandable metrics are (in descending order of usefulness): median wait > average speed > long wait % > travel time variability. I would support getting rid of travel time variability and long wait %. I think Jesse's idea to replace long wait % with on-time % is spot-on.

tldr: I agree with Jesse, but I'd say to keep the score in for this PR so we can localize the debate over score to #449.

hathix:
@exxonvaldez , can you fix the merge conflicts here?

exxonvaldez:
Jesse has already refactored this component and fixed the alignment of the table headers and the table row padding.  There are also some super subtle color tweaks in this PR to slightly lighten the black text on a yellow background, and lighten the "light green" one shade, but it's been so long that it's moot.  I'll close this PR.

PR #481, opened 2019-12-09
Add GraphQL API fields for timetables and schedule adherence; update timetable parsing

youngj:
This PR updates the methods for parsing timetables from the GTFS feed to work with multiple agencies. It changes the data format for storing timetables on disk and S3 from CSV to a JSON format that is similar to the other JSON formats we use, which will make the schedule data easier to work with. It also updates the GraphQL API to allow the frontend to get statistics about scheduled headways, wait times, and trip times, on-time performance, and bunching/gapping.

This PR involved refactoring the code for parsing route configuration from GTFS feeds, in order to avoid duplicating code when parsing timetables. Most of the logic in backend/save_routes.py is now in backend/models/gtfs.py , although it should generally work the same way as before.

### Overview of GTFS Timetables

In a GTFS feed, each trip is associated with a particular "service ID", which identifies a subset of timetables that are effective on certain dates. Depending on the agency, each date may only have one service ID in effect, or may have multiple service IDs on the same date. The GTFS feed allows clients to determine which service IDs are associated with each date. The CSV file calendar.txt defines the default service IDs that are effective on particular weekdays within a date range, e.g:

```
service_id,start_date,end_date,monday,tuesday,wednesday,thursday,friday,saturday,sunday
2_merged_9175742,20190826,20191011,0,0,0,0,0,1,0
2_merged_9175745,20191012,20191227,0,0,0,0,0,1,0
1_merged_9175740,20190826,20191011,1,1,1,1,1,0,0
1_merged_9175743,20191012,20191227,1,1,1,1,1,0,0
3_merged_9175744,20191012,20191227,0,0,0,0,0,0,1
3_merged_9175741,20190826,20191011,0,0,0,0,0,0,1
```

The CSV file calendar_dates.txt defines exceptions to these default service IDs on particular dates (perhaps holidays), either by adding a new service ID to a particular date (exception_type=1), or by removing a default service ID from a particular date (exception_type=2), e.g:

```
service_id,date,exception_type
1_merged_9175740,20190907,2
1_merged_9175743,20191128,2
1_merged_9175743,20191225,2
3_merged_9175744,20191128,1
3_merged_9175744,20191225,1
3_merged_9175741,20190907,1
```

(The SF Muni GTFS feed in this example actually has a bug because it tries to remove 1_merged_9175740 from 20190907 even though 1_merged_9175740 was not scheduled on that date.)

### Timetable JSON File Format

Since many days have identical timetables, in order to avoid storing redundant data, this PR stores each timetable with a particular set of service IDs only once. Each particular set of service IDs is identified by a "date key", which is the string representation of one of the dates where that set of service IDs is scheduled (e.g. 2019-10-14). There may be many different dates that have the same date key. A JSON file is stored in S3 which maps all dates specified in the GTFS feed to their corresponding date keys, e.g.:

```
{"version":"v1","date_keys":{"2019-08-31":"2019-08-31","2019-09-01":"2019-08-31","2019-09-07":"2019-09-07","2019-09-08":"2019-08-31","2019-09-14":"2019-08-31","2019-09-15":"2019-08-31","2019-09-21":"2019-08-31","2019-09-22":"2019-08-31","2019-09-28":"2019-08-31","2019-09-29":"2019-08-31","2019-10-05":"2019-08-31","2019-10-06":"2019-08-31","2019-10-12":"2019-10-12","2019-10-13":"2019-10-12","2019-10-19":"2019-10-12","2019-10-20":"2019-10-12","2019-10-26":"2019-10-12","2019-10-27":"2019-10-12","2019-11-02":"2019-10-12","2019-11-03":"2019-10-12","2019-11-09":"2019-10-12",...
```

When a new GTFS feed is released covering new dates, this file will be overwritten to include the new date keys, as well as the old date keys from the previous version of the file. 

The date key is used in the path of the cache files and S3 objects where the timetables are stored for each route, e.g. `timetables_v1_muni_2019-08-26_1.json`. In order to look up a timetable for any date, the code first looks up the date key for that date in order to construct the path to this file.

An individual route's timetable looks like this:

```
{"version":"v1","agency":"muni","route_id":"1","date_key":"2019-08-26","timezone_id":"America/Los_Angeles","service_ids":["1_merged_9175740"],"arrivals":{"0":{"14015":[{"t":19500,"i":378},{"t":21300,"i":221},{"t":22500,"i":187},{"t":23340,"i":313},{"t":24060,"i":150},{"t":24660,"i":287},{"t":25260,"i":423},{"t":25740,"i":140}...
```

In the "arrivals" object, the top level key (e.g. "0") is a direction ID, and the second level key (e.g. "14015") is a stop ID.

The integer "t" values stored in this file are the number of seconds past midnight on a date where the timetable is in effect, in the specified timezone. This makes it possible to easily construct timetables for a specific date by adding the timestamp at midnight to the relative "t" value from the file.

If the schedule has a departure time that is different from the arrival time, the departure time will be stored as an integer value "e" that is also the number of seconds past midnight.

The integer "i" values stored in this file identify distinct trips, with a 1-to-1 mapping to the string trip_ids in the GTFS feed. This allows the code in backend/models/trip_times.py to compute scheduled trip times in exactly the same way that it computes actual trip times. 

Currently the original string trip_id from the GTFS feed is not stored in this JSON file. It's possible that  in the future it may be necessary to store the mappings between integer "i" values and the string GTFS trip IDs if we want to associate the trip ID that may be provided in GTFS-realtime feeds with a particular scheduled arrival.

### GraphQL API Updates

The GraphQL API now allows the client to fetch a lot more statistics about timetables for a particular stop or pair of stops, as well as comparisons between the scheduled arrival times or headways and the actual values.

For example:

```
query {  
  routeMetrics(agencyId:"muni", routeId:"27") {
    trip(startStopId:"13172", endStopId:"15151") {
      interval(dates:["2019-12-03"]) {
        headways {
          median
          bunchCount: countRange(max:2)
          count
        }
        scheduledHeadways {
          median
        }
        waitTimes {
          median
        }
        scheduledWaitTimes {
          median
        }
        headwayScheduleDeltas {
          min
          median
          max
          gapCount: countRange(min:5)
          count
        }
        tripTimes {
          median
        }
        scheduledTripTimes {
          median
        }
        arrivals
        scheduledArrivals
        scheduleAdherence(earlySec:60, lateSec:300) {
          onTimeCount          
          earlyCount
          lateCount
          missingCount
          scheduledCount
          arrivalScheduleDeltas {
            min
            median
            max            
            count
          }
        }
      }
    }
  }
}
```

The scheduledWaitTimes, scheduledTripTimes, and scheduledHeadways fields are analogous to the existing waitTimes, tripTimes, and headways fields, but using timetables instead of actual arrival data. The code is structured so that the Timetable object returns a data frame that has similar columns to the data frame returned by the ArrivalHistory object. This makes it easy to compute the same metrics for scheduled vs actual data just by passing a different data frame.

The BasicStats object now contains a `countRange` field that can be used to get statistics about bunching and gapping. To see how often vehicles were bunched together (less than 2 minute headways):

```
...
headways {
          bunchCount: countRange(max:2)
          count
}
...
```

The headwayScheduleDeltas field returns statistics about "actual headway minus scheduled headway" in minutes. This makes it possible to get statistics about unscheduled gaps in service. To see gaps where headways between vehicles were 5 minutes or more longer than the scheduled headways at that time:

```
...
headwayScheduleDeltas {
   gapCount: countRange(min:5)
   count
}
```

To determine the scheduled headway associated with each actual arrival, this PR uses the headway associated with the closest scheduled arrival time (which may be before or after the actual arrival time). Multiple actual arrivals may be associated with the same scheduled arrival time.

The arrivals and scheduledArrivals fields just provide the total integer number of arrivals or scheduled arrivals at the origin stop (this should probably be changed to the destination stop, and then departures / scheduledDepartures could refer to the origin stop).

The scheduleAdherence field allows getting information about on-time performance relative to particular early/late thresholds.  The scheduleAdherence field takes each scheduled arrival time and matches it to an actual arrival time (opposite of headwayScheduleDeltas , which takes each actual arrival time and matches it to a scheduled arrival time). The process of matching scheduled arrival times to arrival arrival times is somewhat complicated. If any arrivals exist within the "on time" interval, the closest one within the on-time interval to the scheduled time will be used, which may not be the closest overall arrival. For example if the on-time interval is 60 seconds early to 300 seconds late, and a bus arrives 90 seconds before the scheduled time and 100 seconds after the scheduled time, the one that is 100 seconds after the scheduled time will be selected as the "closest" arrival.

Additionally, for computing schedule adherence, each actual arrival is only associated with at most one scheduled arrival (the closest one to the actual arrival time). If multiple scheduled arrival times are associated with the same actual arrival time, only the pair that is closest together will be considered "matching". The others will be considered "missing". (However, missing vehicles may still have arrived much earlier or later than the scheduled arrival time. It is possible that multiple vehicles may be closest to another scheduled arrival time, and only one of those vehicles will be matched. If vehicles are very early or very late so that they are closer to another scheduled arrival time than the one they were actually targeting, then they may show up as missing.)

The scheduleAdherence field lets the client fetch the number of vehicles that were on time, early, late, and missing, and the total number of scheduled vehicles at the origin stop. It also allows the client to fetch statistics about how early or late vehicles were relative to the closest scheduled arrival.

```
...
   scheduleAdherence(earlySec:60, lateSec:300) {
          onTimeCount          
          earlyCount
          lateCount
          missingCount
          scheduledCount
          arrivalScheduleDeltas {
            min
            median
            max            
            count
          }
        }
...
```

### Command line scripts

The timetable.py script has a --comparison flag that matches scheduled arrivals to actual arrivals:

```
root@e39e7d6f8301:/app/backend# python timetable.py --agency=muni --route=12 --dir=1 --stop=13476 --date=2019-11-19 --comparison
Start: 2019-12-09 00:19:48.001302
2019-11-19 05:25:00 (1574169900) +  :00  dir:1   nan min headway   actual: 05:22:31  early -2.5 min     nan min headway (- min)
2019-11-19 05:45:00 (1574171100) +  :00  dir:1  20.0 min headway   actual: 05:35:14  early -9.8 min    12.7 min headway (-7.3 min)
2019-11-19 06:04:00 (1574172240) +  :00  dir:1  19.0 min headway   actual: 05:57:09  early -6.8 min    21.9 min headway (+2.9 min)
2019-11-19 06:23:00 (1574173380) +  :00  dir:1  19.0 min headway   actual: 06:19:59  early -3.0 min    22.8 min headway (+3.8 min)
2019-11-19 06:40:00 (1574174400) +  :00  dir:1  17.0 min headway   actual: 06:47:08  late  +7.1 min    27.1 min headway (+10.1 min)
2019-11-19 06:55:00 (1574175300) +  :00  dir:1  15.0 min headway   actual: 06:52:07  early -2.9 min     0.0 min headway (-15.0 min)
2019-11-19 07:09:00 (1574176140) +  :00  dir:1  14.0 min headway   actual: 07:10:16        +1.3 min    18.1 min headway (+4.1 min)
2019-11-19 07:24:00 (1574177040) +  :00  dir:1  15.0 min headway   actual: 07:37:12  late  +13.2 min    26.9 min headway (+11.9 min)
2019-11-19 07:43:00 (1574178180) +  :00  dir:1  19.0 min headway   actual: 07:40:52  early -2.1 min     3.7 min headway (-15.3 min)
2019-11-19 07:56:00 (1574178960) +  :00  dir:1  13.0 min headway   actual: 07:51:02  early -5.0 min    10.2 min headway (-2.8 min)
2019-11-19 08:10:00 (1574179800) +  :00  dir:1  14.0 min headway   actual: 08:07:53  early -2.1 min    16.9 min headway (+2.9 min)
2019-11-19 08:25:00 (1574180700) +  :00  dir:1  15.0 min headway   actual: 08:28:46        +3.8 min    20.9 min headway (+5.9 min)
2019-11-19 08:40:00 (1574181600) +  :00  dir:1  15.0 min headway   actual: 08:43:05        +3.1 min    14.3 min headway (-0.7 min)
2019-11-19 08:55:00 (1574182500) +  :00  dir:1  15.0 min headway   actual: 08:56:47        +1.8 min    13.7 min headway (-1.3 min)
2019-11-19 09:10:00 (1574183400) +  :00  dir:1  15.0 min headway   actual: 09:07:06  early -2.9 min    10.3 min headway (-4.7 min)
2019-11-19 09:25:00 (1574184300) +  :00  dir:1  15.0 min headway   actual: 09:23:11  early -1.8 min    16.1 min headway (+1.1 min)
2019-11-19 09:40:00 (1574185200) +  :00  dir:1  15.0 min headway   actual: None
2019-11-19 09:55:00 (1574186100) +  :00  dir:1  15.0 min headway   actual: 09:54:55        -0.1 min    31.7 min headway (+16.7 min)
...
-----
65 scheduled arrivals
On-time  : 30.8% (20/65)
Late     : 7.7% (5/65) more than 5 min late
Early    : 53.8% (35/65) more than 1 min early
Missing  : 7.7% (5/65)
```

The headways.py script also has a --comparison flag that matches actual arrivals to scheduled arrivals:

```
root@e39e7d6f8301:/app/backend# python headways.py --date=2019-11-19 --agency=muni --route=1 --stop=16290 --comparison
Date: 2019-11-19
Time of Day: [None, None)
Route: 1 (1-California)
Stop: 16290 (Sacramento St & Battery St)
Direction: 0 (Outbound to Geary Blvd & 33rd Ave)
2019-11-19 05:28:16 (1574170096) +  :13 vid:5816  dir:0    8m   nan min   scheduled: 05:27:22 (+0.9 min) @ nan min headway (    - min)
2019-11-19 05:56:49 (1574171809) +  :10 vid:5805  dir:0    7m  28.6 min   scheduled: 05:57:22 (-0.6 min) @ 30.0 min headway ( -1.4 min)
2019-11-19 06:16:31 (1574172991) +  :16 vid:5795  dir:0    5m  19.7 min   scheduled: 06:17:22 (-0.8 min) @ 20.0 min headway ( -0.3 min)
2019-11-19 06:30:09 (1574173809) +  :07 vid:5860  dir:0    2m  13.6 min   scheduled: 06:31:22 (-1.2 min) @ 14.0 min headway ( -0.4 min)
2019-11-19 06:41:31 (1574174491) +  :06 vid:5851  dir:0   11m  11.4 min   scheduled: 06:43:22 (-1.9 min) @ 12.0 min headway ( -0.6 min)
2019-11-19 06:53:45 (1574175225) +  :12 vid:5807  dir:0   12m  12.2 min   scheduled: 06:53:22 (+0.4 min) @ 10.0 min headway ( +2.2 min)
2019-11-19 07:02:38 (1574175758) +  :10 vid:5776  dir:0   13m   8.9 min   scheduled: 07:03:22 (-0.7 min) @ 10.0 min headway ( -1.1 min)
2019-11-19 07:10:14 (1574176214) +  :30 vid:5816  dir:0   21m   7.6 min   scheduled: 07:11:22 (-1.1 min) @ 8.0 min headway ( -0.4 min)
....
```

### Pre-Computed Statistics (not implemented yet)

Currently the GraphQL API only provides statistics for a single stop or pair of stops. However, the web UI will likely require presenting information for all routes at once, and all stops along a route, in a similar way to how it currently presents information about actual median wait times, average speeds, travel time variability, etc. For example the frontend might want to get the average on-time percentage for all routes on a particular day. To implement this we could generate files with pre-computed statistics in the same way that we precompute statistics about actual wait times and trip times. It may be good to discuss which schedule-related statistics we want to precompute.


youngj:
Thanks for the feedback!

> If we want to plot the schedule on the Marey chart, should there be an API for getting the schedule? Or should the client reimplement the same logic as the backend to find the right timetable?

I'm thinking the client could reimplement the same logic as the backend to find the right timetable. This is the same as how it currently works for actual arrival histories, except with the added step of looking up the date key. This would allow the Python backend to do less work and send less data, so it would probably be more scalable.

> I can't tell if you've added schedule stats to the IntervalMetrics

Yes, the schedule stats are added to IntervalMetrics, so it is also available for the intervals of the day chart.

> as Tal from Muni suggested, we might prioritize OTP vs bunching/gapping differently for frequent vs infrequent service (where the cutoff is around 10 or 15 min headway). TCQSM 3rd edition also has a bunching/gapping rubric, but for frequencies of 10 min or less only. Their rubric is based on the standard deviation of headways, assuming a constant headway, and it may not be practical for us to implement.

Muni's defines bunching and gaps on https://www.sfmta.com/reports/percentage-transit-trips-bunching-or-gaps  :

> The percentage of observed time point arrivals on the Rapid Network with spacing of less than two minutes (one minute for scheduled headways of less than five minutes) or gaps of greater than 5 minutes beyond their scheduled headway (time between vehicles) times.

For gaps, `headwayScheduleDeltas` can allow computing the rate of gaps of greater than 5 minutes beyond the scheduled headway:

```
headwayScheduleDeltas {
   gapCount: countRange(min:5)
   count
}
```

For bunching, since Muni's threshold is either 2 or 1 minutes depending on whether the scheduled headway is greater or less than 5 minutes, I don't think the API currently provides a way to compute Muni's bunching statistic if the scheduled headway varies over the course of the interval. If you assume a fixed scheduled headway over an interval, then you could use a query like this:

```
        headways {
          bunchCount: countRange(max:2)
          rapidBunchCount: countRange(max:1)
          count
        }
        scheduledHeadways {
          median
        }
```

> Re: "arrivals" and "scheduledArrivals" nomenclature: Seems fine as is, since we have so many other stats that are based on just the start stop.

I updated the API to be consistent about using arrival times for the destination stop (endStopId) and departure times for the origin stop (startStopId). The fields `arrivals` and `scheduledArrivals` now return the number of arrivals at the destination stop, while `departures` and `scheduledDepartures` return the number of departures at the origin stop.

Also, instead of the `scheduleAdherence` field, there is a `departureScheduleAdherence` field which refers to the on-time performance of departure times at the origin stop, and a `arrivalScheduleAdherence` field which refers to the on-time performance of arrival times at the destination stop. It's possible the frontend may only need to use `departureScheduleAdherence`.

> So the service IDs are mapped under some date from the timetable. Does this mean that uploading multiple GTFS's means that choosing a date from the past in the client will use the scheduled trip times from back then?

That's correct.

> What if there are more actual arrivals than are scheduled? I'm guessing it'd just get ignored as it'll be too far off from other scheduled arrivals.

That's correct, `departureScheduleAdherence` and `arrivalScheduleAdherence` statistics are generated from a data table with one row per scheduled time. If multiple actual times are closest to the same scheduled time, only the closest actual time will be used in the schedule adherence stats, and the others will be ignored. 

In cases where there are more actual arrivals than scheduled, this would be visible as negative values in `headwayScheduleDeltas`  (actual headways minus scheduled headways). `headwayScheduleDeltas` is generated from a data table with one row per actual departure time, and multiple actual departure times can match the same scheduled departure time.

> Also, it looks like ScheduleAdherence will be really inaccurate for very frequent (ie. 5 minutes or better) routes so we might be better off only showing HeadwayAdherence in the frontend in those cases. 

With frequent routes you could potentially use smaller values of `earlySec`/`lateSec`. However, I agree that `headwayScheduleDeltas` would be more useful than `departureScheduleAdherence` for very frequent routes.

Would anyone like to review the latest changes before I merge?

PR #483, opened 2019-12-11
Show "stop to stop" time series data 

exxonvaldez:
## Proposed changes

This is a quick front end for #372, which is merged in.  That PR gives us "stop to stop" data by day, when the user selects a date range.  This PR is for discussing how to start showing "time series" data, which was broadly sketched out in https://docs.google.com/document/d/17I9-LRN1TLD0KIFP5a-jI_VVnO3ydgzbwj1_tYKpe1w/edit?usp=sharing

This PR, for the sake of expediency, adds a new "by day" tab (InfoByDay.jsx) to the "stop to stop" info.  This new tab has the following charts, all of which are, not surprisingly, by day:

- Median/90th percentile wait times and trip times (like the intervals of the day chart).  Note: the radio buttons only apply to this one chart, haven't made that super clear in the UI yet.
- Long wait %
- Speed
- Travel Variability
- Score

I've also added some partial code for encoding the date range settings in the URLs, just enough for my testing purposes.  It should be reusable and able to be completed once we get #347 in.

## Screenshot

In this particular example, we are seeing KT outbound from Sunnydale to Embarcadero for the last 7 days.  What's interesting is that the median wait/trip times don't reveal an apparent decrease in performance on Saturday, which you can see in the later charts (or by switching to the 90th percentile wait/trip times).

Screen Shot 2019-12-11 at 12 11 41 AM

Screen Shot 2019-12-11 at 12 12 50 AM

## Link to demo, if any

https://ot-terence-time-series.herokuapp.com/

There will be some challenges around making charts that can successfully show a few days' data as well as three months' data, if we're able to get that in a performant way.  Currently, requesting 30 days' worth of data usually fails,  after waiting for some time.

Crosshairs (hover effects) have not been implemented for the charts below the first one yet.  Same goes for showing the average value for the time period -- not for charts below the first one yet.


akgupta89:
So are we creating a new tab to display a range of data rather than augmenting the current setup to  display a date range?

frontend/src/helpers/routeCalculations.js
```diff
@@ -534,3 +534,48 @@ export function milesBetween(p1, p2) {
...
   return metersToMiles(meters);
 }
+
+/**
+ * Returns stop to stop distance based on graph params.
+ *
+ * @param {Object} myGraphParams from Redux state
+ * @param {Object} myRoutes from Redux state
+ */
+export function computeDistance(myGraphParams, myRoutes) {
```
youngj: Seems to do the same thing as getDistanceInMiles in helpers/mapGeometry.js , which also handles loop routes.

In general, it would be better for helper functions not to take `graphParams` as a parameter since it makes the function harder to reuse, and it is less clear which properties it actually depends on.

frontend/src/routesMap.js
```diff
@@ -15,20 +21,41 @@ export default {
     the ? after the : means an optional paramter variable
     ()* shows am optional parameter label
     */
-    path: `/route/:routeId/:directionId?/:startStopId?/:endStopId?`,
+    path: `/route/:routeId/:directionId?/:startStopId?/:endStopId?`
+      + `/(${DATE})*/:date?/(${START_DATE})*/:startDate?`
```
youngj: Seems weird that the start date appears after the end date in the URL.

Query string parameters would probably make the code simpler and more flexible compared to adding route segments for all of the date/time parameters.

exxonvaldez: Good feedback, will work on both items.

frontend/src/helpers/routeCalculations.js
exxonvaldez: Good catch, will update this to use the new method.  Agree on graphParams comment.

exxonvaldez:
> Could avoid serializing properties that are the same as the default so the URL isn't so long.

Went ahead and did this.  It's a little ambiguous for whether the dates should go in the query string, since the default date changes over time.  I decided it was cleaner all around to omit the dates if the default dates are being used.

> Also could hide the By Day tab if only one day is selected.

Decided to keep the tab visible for discoverability (since right now we only select one day by default).  If only one day is selected, I replaced the charts with instructions to select start and end dates.

PR #490, opened 2019-12-19
Add clear stops button

### LINKED ISSUE #438: Add button to clear to/from stops
On a page like https://muni.opentransit.city/route/F/direction/0/fromStop/15678/toStop/15680 (where the To and From stops are selected), it's hard to go back to the route overview screen. Can we add a clear button somewhere (like the map) to do that? (Getting clickable breadcrumbs will also work, but having an explicit clear button makes it even easier.)

zachzwy:
Fixes #438.

## Proposed changes
Add a clear stops button.

It will clear the selected stops. However, the URL doesn't change. Ideally, the URL should be updated as well. I guess redux-first-router should be able to handle it somehow. Not sure how to do it though. Look for any feedback and suggestions.

Also, when the screen becomes smaller, it is not well responsive. See the screenshot below. I will need to address this as well.

## Screenshot
Screen Shot 2019-12-18 at 7 45 33 PM

Screen Shot 2019-12-18 at 7 50 02 PM
...

### DIFF
--- a/frontend/src/actions/index.js
+++ b/frontend/src/actions/index.js
@@ -1,7 +1,14 @@
 import axios from 'axios';
-import { MetricsBaseURL, S3Bucket, RoutesVersion, TripTimesVersion, WaitTimesVersion, ArrivalsVersion } from '../config';
-import { getTimePath } from '../helpers/precomputed';
 import Moment from 'moment';
+import {
+  MetricsBaseURL,
+  S3Bucket,
+  RoutesVersion,
+  TripTimesVersion,
+  WaitTimesVersion,
+  ArrivalsVersion,
+} from '../config';
+import { getTimePath } from '../helpers/precomputed';
 import { MAX_DATE_RANGE } from '../UIConstants';
 
 /**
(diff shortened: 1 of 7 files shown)


hathix:
This has been sitting around for a while. @zachzwy , are you OK to finish this?

zachzwy:
I'll keep working on this PR. @hathix

zachzwy:
The below screenshot is a very react-select native "clear" button (An "X" icon right next to the select content).

Screen Shot 2020-02-05 at 8 11 50 PM

It looks good until I revisit the issue #438 we are trying to solve. I believe the goal here to enable user to easily go back to the route overview screen. With this proposal,  user needs to click "X" icon twice for from stop and to stop, which I don't think is a good idea.

We might want to think of a different UI to enable this. Or, I actually believe clicking the breadcrumb button is a pretty good way to achieve this goal.

What do you think? @exxonvaldez  @hathix

exxonvaldez:
Hmm, there may not be any quick solution here.  Seems like the more intuitive thing is group the stop pickers together into a "fieldset" (in HTML terms) or outlined inputs (in Material UI terms).  Then you can associate a clear button with the entire fieldset.

As for quick fixes, the react-selects with clear buttons look good.  If the user clears the "to" picker or the "from" picker, then the UI should be showing the route summary, so it's just one click away, although not obvious.

The breadcrumbs are working to go back to the route summary, but are also not obvious.  There are actually two breadcrumb links that can work -- one to go to the route summary with no direction chosen, and one for the route summary in a given direction.

The browser back button also works to go the route summary, if the user didn't follow a direct link with stops preselected.

Finally, there's the new #469 "return to dashboard" button, which could be enhanced to be more like the brower's back button.  Especially now that we have a clickable logo which always goes to the dashboard.

PR #491, opened 2019-12-26
Compute wait time statistics over multiple days

### LINKED ISSUE #488: Compute wait time statistics over multiple days
Currently backend/models/wait_times.py computes exact statistics for wait times in one day. However the frontend needs to compute statistics for wait times over multiple days. Currently, backend/models/schema.py only returns statistics for the first day, instead of the requested range, so the wait time statistics are not actually correct. Perhaps wait_times.py should be modified to compute percentiles and histograms over multiple days.

youngj:
Fixes #488.

The GraphQL API allows returning wait time statistics over multiple days, but previously the returned statistics were only actually from the first day.

This PR implements a MultiIntervalWaitTimeStats class that returns statistics over multiple intervals, assuming each interval is weighted equally (not weighted by their length).

Most wait time statistics (other than the average) are computed from the cumulative distribution function (CDF), which gives the probability of having a wait time less than a given number of minutes if someone arrives at a stop at a random time between the first and last departure time of the day. 

With each interval weighted equally, the probability of a wait time less than a given number of minutes (CDF value) is the average of the values of the CDFs for each interval. Since each interval's CDF is piecewise linear, the combined CDF will be piecewise linear between all of the unique domain values defining the intervals' CDFs.

### DIFF
--- a/backend/models/metrics.py
+++ b/backend/models/metrics.py
@@ -97,7 +97,10 @@ class RouteMetrics:
 
             wait_stats_arr.append(wait_stats)
 
-        return wait_stats_arr
+        if len(wait_stats_arr) == 1:
+            return wait_stats_arr[0]
+        else:
+            return wait_times.combine_stats(wait_stats_arr)
 
     def get_arrivals(self, direction_id, stop_id, rng: Range):
         return self._get_count(direction_id, stop_id, rng, self.get_history_data_frame, 'TIME')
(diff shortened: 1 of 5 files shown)


youngj:
That's correct, this code wouldn't give the correct wait times for intervals of unequal size like 6-9am and 3-7pm. 
However, currently the only usage is for an interval of the same size repeated over multiple days.

If there was a need for intervals of different sizes in the future, I think the statistics could be computed by a weighted average of the CDFs with weights corresponding to the relative sizes of each interval.

I added API documentation for the WaitTimeStats type and mentioned that intervals are equally weighted.

PR #492, opened 2019-12-27
Improve algorithm for generating trips from possible arrivals; handle loops, figure-eights, routes that double back

### LINKED ISSUE #312: Fix arrival history and stop geometry for routes that double back on themselves
I was looking at save_routes.py: get_stop_geometry method.  It seems like it could have trouble with shapes that double back on themselves.  This has to do with the calculation of closest_index and getting an index that is upstream of (wrong way from) the previous index.

Jesse concurs that we need to ignore the lines (route segments) upstream of the previous stop when finding the closest index for the next stop.

A similar fix is needed when computing arrival histories.  After a vehicle has passed a stop, we should only look for arrivals at downstream stops.  Currently we sometimes find a repeat arrival at an upstream (wrong way) stop, and discard the arrival. The vehicle was actually arriving a nearby downstream stop.

### LINKED ISSUE #425: Handle directions with loops correctly when computing arrivals and trip times
The Portland Streetcar A Loop and B Loop lines travel in a continuous loop. However, each line has an arbitrary beginning/end stop (which are identical). Currently, OpenTransit considers trips that cross this arbitrary beginning/end stop as separate trips, and is unable to compute trip times across this boundary. We should consider routes with a direction with the same beginning/end stop to be continuous loops and handle this case correctly when computing arrivals and trip times. 

We should update eclipses.py so that the trip_id stays the same for a particular vehicle each time around the loop.

Since a vehicle may arrive at the same stop multiple times with the same trip_id, trip_times.py should make sure that the trip time is based on the next arrival time when a trip_id appears at the destination stop after the departure time at the origin stop with the same trip_id.

When storing trip time statistics for loops, store all pairs of stops, rather than only pairs of stops where the destination stop index is greater than the origin stop index.

### LINKED ISSUE #465: Fix indigo/white polyline issue for loop routes
The indigo Polyline on the route map is intended to highlight the path between the selected start and end stops along a route while the white Polyline highlights the untraveled portion of the route.  In the case of loop routes such as Portland streetcars, the selected end stop may be listed before the start stop in the stops list.  In these cases, the StartStopIcon and EndStopIcons swap and the white Polyline highlights the intended route while the indigo Polyline highlights the untraveled portion of the route.  Logic should be changed to account for the case when a user selects an end stop listed before the start stop.  

Logic resides in MapStops.

youngj:
Fixes #312.
Fixes #425.
Fixes #465.

Previously, eclipses.py did not do a good job at grouping arrivals into trips for certain routes. The previous algorithm simply looked for 3 adjacent possible arrivals with ascending stop indexes. However, for routes that doubled back on themselves (or nearly so) there would be out-of-order possible arrivals between the actual arrivals.

For twisty routes where a single direction doubles back on itself (or nearly so), such as SF Muni's 39-Coit, 36-Teresita, 30-Stockton, 9-San Bruno, etc., it is possible that there may be out-of-order stop indexes in the "possible" arrivals.

As an example, suppose consecutive stop indexes of possible arrivals for some route is 1,2,6,3,7,4,5,6,3,4,7,8,9. In this case, the "best" trip has the stop indexes 1,2,3,4,5,6,7,8,9. 

The previous algorithm, which only looked for 3 or more consecutive ascending stop indexes, would likely find three separate trips: (1,2,6), (4,5,6), and (3,4,7,8,9).

Alternatively, if the algorithm only used the first ascending index it found, it would only find 1,2,6,7,8,9 and miss indexes 3,4,5.

In order to handle these cases, the algorithm keeps track of multiple possible sequences as it loops through the possible arrivals. To avoid needing to keep track of a large number of possible sequences, it drops possible sequences if they are strictly worse than another possible sequence.
A sequence is worse than another sequence, for example, if it is shorter than another sequence and ends in the same stop_index; or if it is the same size as another sequence but ends in a larger stop_index.

When all possible sequences end in a terminal, or if the algorithm processes several rows without being able to extend any of the possible sequences with an ascending stop index, it assumes that the trip has ended and chooses the longest possible sequence as the "best".

After it finds the end of a trip, it resets the possible sequences and continues processing possible arrivals where the previous sequence ended.

Various heuristics are used which seem to provide good results in practice for the routes tested:

* Each trip must contain arrivals at least 3 stops (unless the route only contains 2 stops in that direction)
* If there is a gap of more than 4 stops, the arrivals will be split into separate trips, unless the time between the two stops is less than 5 minutes. This avoids adding long gaps within a trip where a vehicle likely was out of service (which happens occasionally on various normal routes like the N-Judah), but avoids splitting trips in the case of 36-Teresita when the vehicle skips the Myra Way section after 9pm (https://www.sfmta.com/sites/default/files/36_teresita_pdf.pdf)
* If the algorithm processes 4 possible arrivals without being able to extend any possible trip sequences, it will determine that the trip has probably ended (even if it hasn't reached the terminal)

## Figure-eight routes

The 36-Teresita contains a figure-eight (https://www.sfmta.com/sites/default/files/36_teresita_pdf.pdf) where three stops in the Inbound direction along Clarendon Ave (Olympia Way, Galewood Circle, Panorama Drive) are repeated twice along the route (with identical stop IDs in the GTFS file). This PR handles this case by creating unique OpenTransit stop IDs for each subsequent time a stop is reached in a given trip. For example, the first time 36-Teresita reaches Clarendon Ave & Olympia Way, we use the stop ID 14035; the second time it reaches Clarendon Ave & Olympia Way, we use the stop ID 14035-2. This allows the frontend code to continue assuming that each stop ID uniquely identifies a single point along a route for a given direction.

## Loop routes

Portland Streetcar A-Loop and B-Loop have a defined start/end stop in the GTFS feed, which appears in stop_times.txt at the start and end of each trip with the same GTFS stop ID. However, in practice people can board the vehicle before the "end" stop and continue around the loop after the "start" stop. 

To simplify handling of loop routes, the GTFS parsing code sets a boolean `loop` property to `true` for the direction, and removes the duplicate "end" stop ID from the list of stops for that direction (instead of adding a "-2" stop like for figure-eight routes).

When parsing schedules for loop routes from the GTFS feed, arrival times at the end stop are added to the start stop. If the GTFS feed already defines an arrival time at the start stop within a few minutes, it is updated to use the arrival time at the previous end stop.

When computing arrival times for loop routes, get_arrivals_with_ascending_stop_index in eclipses.py continues appending arrivals to the same trip when the vehicle goes from stop index N-1 to stop index 0. (For loop routes, the difference of stop index is calculated mod N.)

For loop routes, the same vehicle may arrive at the same stop many different times with the same trip ID. When computing trip times for loop routes, trip_times.py finds the first arrival that appears after the departure time with the same trip ID. 

The frontend is updated to check the `loop` property in several places:
* computing the list of downstream stops after a given stop
* getting the coordinates along a route between two stops
* getting the distance along a route between two stops
* determining which segments to highlight between two stops

Some frontend code was refactored to avoid duplicating the same logic in multiple places.

## Route bounds

The route object in S3 now contains a `bounds` property which contains an array of two lat/lon coordinate objects that bound all stops along that route. The frontend uses these bounds when showing a particular route. 

## Fix to distance calculations

The code for parsing GTFS routes had a bug which transposed latitudes and longitudes when computing distances along a route, which caused all of the distances for directions and stops to be incorrect. This caused all of the statistics in the frontend based on distances, including speeds and scores, to be incorrect. The new route configuration has corrected distances in `stop_geometry` and the `distance` field for each direction.

### DIFF
--- a/backend/models/gtfs.py
+++ b/backend/models/gtfs.py
@@ -292,9 +292,6 @@ class GtfsScraper:
 
             route_config = gtfs_route_id_map[gtfs_route_id]
 
-            if route_config.id != '36':
-                continue
-
             arrivals_by_service_id = {}
             trip_ids_map = {}
 
@@ -475,7 +472,92 @@ class GtfsScraper:
 
                 direction_arrivals[stop_id].append(arrival_data)
 
+        for dir_info in route_config.get_direction_infos():
+            if dir_info.is_loop():
+                direction_arrivals = arrivals_by_direction[dir_info.id]
+                self.clean_loop_schedule(dir_info, direction_arrivals)
+
         return arrivals_by_direction
+
+    def clean_loop_schedule(self, dir_info, direction_arrivals):
+        stop_ids = dir_info.get_stop_ids()
+        first_stop_id = stop_ids[0]
+        last_stop_id = stop_ids[0] + "-2"
+
+        sort_key = lambda arr: arr['t']
+
+        first_stop_arrivals = sorted(direction_arrivals[first_stop_id], key=sort_key)
+        last_stop_arrivals = sorted(direction_arrivals[last_stop_id], key=sort_key)
+
+        prev_trip_ints = {}
+
+        # avoid creating duplicate arrivals at the first stop
+        # when the previous trip did not complete the entire loop.
+        non_duplicate_arrivals = []
+        prev_stop_arrival_time = None
+        for i, stop_arrival in enumerate(first_stop_arrivals):
+            stop_arrival_time = stop_arrival['t']
+            if stop_arrival_time != prev_stop_arrival_time:
+                non_duplicate_arrivals.append(stop_arrival)
+            else:
+                prev_trip_ints[stop_arrival['i']] = first_stop_arrivals[i-1]['i']
+
+            prev_stop_arrival_time = stop_arrival_time
+
+        first_stop_arrivals = direction_arrivals[first_stop_id] = non_duplicate_arrivals
+
+        first_stop_arrival_index = 0
+        num_first_stop_arrivals = len(first_stop_arrivals)
+
+        unmatched_last_stop_arrivals = []
+
+        for last_stop_arrival in last_stop_arrivals:
+            last_stop_arrival_time = last_stop_arrival['t']
+
+            is_matched = False
+
+            while first_stop_arrival_index < num_first_stop_arrivals:
+                first_stop_arrival = first_stop_arrivals[first_stop_arrival_index]
+                first_stop_arrival_time = first_stop_arrival['t']
+
+                first_stop_arrival_index += 1
+
+                if first_stop_arrival_time >= last_stop_arrival_time:
+                    # assume the next loop is a continuation of the same trip if the times are close
+                    if first_stop_arrival_time <= last_stop_arrival_time + 360:
+                        prev_trip_ints[first_stop_arrival['i']] = last_stop_arrival['i']
+                        is_matched = True
+                        if 'e' not in first_stop_arrival:
+                            first_stop_arrival['t'] = last_stop_arrival_time
+                            first_stop_arrival['e'] = first_stop_arrival_time
+
+                    break
+
+            if not is_matched:
+                unmatched_last_stop_arrivals.append(last_stop_arrival)
+
+        print(unmatched_last_stop_arrivals)
+
+        if len(unmatched_last_stop_arrivals) > 0:
+            print(first_stop_id)
+            print(len(direction_arrivals[first_stop_id]))
+            print(direction_arrivals[first_stop_id])
+            direction_arrivals[first_stop_id].extend(unmatched_last_stop_arrivals)
+            print(len(direction_arrivals[first_stop_id]))
+            print(direction_arrivals[first_stop_id])
+
+        orig_trip_ints = {}
+        for trip_int, prev_trip_int in prev_trip_ints.items():
+            orig_trip_int = orig_trip_ints.get(prev_trip_int, prev_trip_int)
+            orig_trip_ints[trip_int] = orig_trip_int
+
+        del direction_arrivals[last_stop_id]
+
+        for stop_id, stop_arrivals in direction_arrivals.items():
+            for stop_arrival in stop_arrivals:
+                orig_trip_int = orig_trip_ints.get(stop_arrival['i'], None)
+                if orig_trip_int is not None:
+                    stop_arrival['i'] = orig_trip_int
 
     def get_stop_times_for_trip(self, trip_id):
         if self.stop_times_by_trip is None:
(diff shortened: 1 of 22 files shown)


exxonvaldez:
Changes sound pretty amazing, in particular the handling for doubling back and figure eights.  Will be nice to see these old bugs gone.

I looked over the UI changes, and they look good.  I appreciate the refactoring of repeated logic around handling of stop lists.  The addition of route bounds to the route config is a good idea.

backend/models/gtfs.py
```diff
@@ -687,7 +827,7 @@ def project_xy(lon, lat, z=None):
         shape_prev_lat = np.r_[shape_lat[0], shape_lat[:-1]]
 
         # shape_cumulative_dist[i] is the cumulative distance in meters along the shape geometry from 0th to ith coordinate
-        shape_cumulative_dist = np.cumsum(util.haver_distance(shape_lon, shape_lat, shape_prev_lon, shape_prev_lat))
+        shape_cumulative_dist = np.cumsum(util.haver_distance(shape_lat, shape_lon, shape_prev_lat, shape_prev_lon))
```
EddyIonescu: Thanks for finding this!

frontend/src/helpers/mapGeometry.js
```diff
@@ -67,23 +102,47 @@ export function getDistanceInMiles(
...
+            stops[stopIds[(i + 1) % numStops]]
+          );
+        }
+        startIndex = 0;
+      }
 
-    distance = milesBetween(fromStopInfo, toStopInfo);
+      for (let i = startIndex; i < toStopIndex; i++) {
+        miles += milesBetween(
+          stops[stopIds[i]],
```
EddyIonescu: So here you're calculating the straight-line distance between two stops while going through the intermediary stops, while before it was just a straight line directly between them?

youngj: That's correct, although I think this case wouldn't normally occur because stop_geometry should be available for all stops displayed in the UI.

backend/models/gtfs.py
youngj: I also created the bug in the first place :)

EddyIonescu: It's surprising that it took so long to find, as all the values in the dashboard seemed reasonable (other than the UI bug that showed super high speeds).

frontend/src/helpers/mapGeometry.js
EddyIonescu: I see, although it'll be useful for agencies that don't provide any geometry, so thanks for adding that now.

backend/models/gtfs.py
youngj: Yeah, it was hard to notice because the distances were typically only ~10% off from the actual distances. I found it while investigating why one of the routes didn't have strictly ascending x-values for some stops in the travel time chart.

PR #493, opened 2019-12-27
Run backend unit tests via GitHub Actions

youngj:
This PR automatically runs backend tests on every push using GitHub Actions. 

Unit tests can be added in the backend/tests directory with file names like test*.py, following the same template as test_util.py.

Individual tests can be run in the docker shell like `python tests/test_util.py`. 

To run all tests in backend/tests, run `python -m unittest discover -s tests`.

The status of the tests will be shown in the GitHub pull request. We could optionally require all tests to be passing before a pull request can be merged.


youngj:
test*.py is the default search pattern for the Python unittest module to discover test packages, although I wouldn't mind using *_test.py instead: https://docs.python.org/3/library/unittest.html#unittest-test-discovery

youngj:
Showing test coverage in pull requests sounds good to me. I haven't looked into how to do that yet, but can open an issue for it.

PR #498, opened 2020-01-06
reorganize Redux store; rename state properties and actions; only store currently visible precomputed stats to simplify code

youngj:
The frontend Redux store has several properties with confusing/inconsistent names, such as `state.fetchGraph.graphData`, `state.routes.routes`, `state.routes.graphParams`, etc. 

This PR splits out some properties from routesReducer.js into separate top-level objects and moves most reducer code into reducers/index.js instead of creating new files for each top-level object, since having separate files seems like unnecessary overhead in most cases.

This PR reorganizes/renames state properties as follows:

`state.routes.routes` => `state.routes.data`
`state.routes.graphParams` => `state.graphParams` (`graphParams` perhaps isn't the best name, but it was used in so many places that I just left it as is)
`state.routes.arrivals` => `state.arrivals.data`
`state.routes.spiderSelection` => `state.spiderSelection.stops`
`state.routes.spiderLatLng` => `state.spiderSelection.latLng`
`state.fetchGraph.graphData` => `state.tripMetrics.data.interval`
`state.fetchGraph.intervalData` => `state.tripMetrics.data.timeRanges`

In addition, the structure of `state.routes.waitTimesCache` and `state.routes.tripTimesCache` added complexity throughout the codebase, since it was necessary to know the current `graphParams` in order to look up data from these caches.

This PR updates the Redux store so it only stores the precomputed wait times and trip times for the current graphParams (similar to how it works for arrivals and tripMetrics), using `state.precomputedStats.waitTimes` and `state.precomputedStats.tripTimes`. This allows simplifying a lot of code in helpers/precomputed.js and helpers/routeCalculations.js, which no longer needs `graphParams`.

(The structure of `state.precomputedStats` will probably change soon if we put wait times, trip times, on-time rates, and anything else in a single JSON file instead of multiple files, and add support for displaying data from multiple agencies.)

There was also inconsistency between different components which have different conventions for naming action functions (e.g. `handleGraphParams`, `myHandleGraphParams`, or `onGraphParams`). It appears that these naming inconsistencies were generally created in order to create local variables using shorthand like `const { myHandleGraphParams } = props;`, which could then be passed to `useEffect` for functional components, without shadowing the `handleGraphParams` from the module scope.

I thought that the inconsistent naming for action functions was confusing, so I changed the property names in most places to be the same name as the name of the action functions (although there is variable shadowing in a few cases).

Renamed `fetchGraphData` to `fetchTripMetrics` to match terminology used by GraphQL API.

Renamed `fetchPrecomputedWaitAndTripData` to `fetchPrecomputedStats` since it will soon be updated to fetch other things like on-time rates.

Updated some of the action functions to keep track of which URL it last loaded data from, to avoid making duplicate requests if the URL is the same as last time.


frontend/src/screens/RouteScreen.jsx
```diff
@@ -67,12 +66,12 @@ function RouteScreen(props) {
             {agency ? agency.title : null}
           </div>
           <div style={{flexGrow: 1}}/>
-          <DateTimePanel dateRangeSupported={graphData || graphError}/>
+          <DateTimePanel dateRangeSupported={tripMetrics || tripMetricsError || tripMetricsLoading}/>
```
exxonvaldez: Good catch for fixing the label here.

frontend/src/components/Info.jsx
```diff
@@ -343,9 +341,14 @@ function Info(props) {
...
-      {graphError ? (
+      {tripMetricsError ? (
         <Box p={2}>
-          <code>{graphError}</code>
+          <code>Error: {tripMetricsError}</code>
+        </Box>
+      ) : null}
+      {tripMetricsLoading ? (
+        <Box p={2}>
+          Loading...
```
exxonvaldez: Nice enhancement.

frontend/src/actions/index.js
```diff
@@ -173,73 +177,91 @@ export function fetchRoutes(params) {
...
-      const s3Url = generateTripTimesURL(agencyId, dateStr, statPath, timePath);
-
-      dispatch({ type: 'REQUEST_PRECOMPUTED_TRIP_TIMES' });
+    const agencyId = agency.id;
+    const waitTimesUrl = generateWaitTimesURL(agencyId, dateStr, 'median-p90-plt20m', timePath);
+    if (getState().precomputedStats.waitTimesUrl !== waitTimesUrl)
+    {
+      dispatch({
+        type: 'REQUEST_PRECOMPUTED_WAIT_TIMES',
+        agencyId: agencyId,
```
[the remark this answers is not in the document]
youngj: Removed for now, although this will probably be needed when we add support for displaying multiple agencies at once. state could store a map of agency ID => precomputed stats.

youngj:
Removed some calls to fetchPrecomputedStats since it is called from handleGraphParams. Added a call to handleGraphParams({}) in the Dashboard so that precomputed stats are fetched for the initial graph params (instead of doing this in the RouteTable component). 

Also reintroduced some 'my' prefixes to avoid shadowing.

youngj:
> Also, I have not tested this PR yet, how much testing should I do?

Probably don't have to do much testing

PR #506, opened 2020-01-12
Simplify code for computing and displaying route stats

youngj:
This PR simplifies the code for computing displaying route-level statistics and scores. It uses a Redux action to store computed statistics and scores for each route (for the current date/time selection). This allows UI components like RouteTable and RouteSummary to easily display route-level statistics and scores without needing to repeat the code to compute them. This will make it easier to add or remove statistics from the route table and route summary in the future.

The unused A/B/C/D letter grade calculations were removed to simplify the code, and the term 'grade' was replaced by 'score'.

Minor functionality changes include always sorting "--" rows to the bottom of the RouteTable in both ascending/descending sorts, and using medianWaitScore instead of waitRank in MapSpider to compute the width of the lines used to draw each route.

### DIFF
--- a/frontend/src/components/MapSpider.jsx
+++ b/frontend/src/components/MapSpider.jsx
@@ -201,8 +201,7 @@ class MapSpider extends Component {
 
         const stats = routeStats[startMarker.routeId] || {};
 
-        // scale wait score to 0, 1, or 2
-        let waitScaled = (stats.medianWaitScore != null) ? Math.trunc((stats.medianWaitScore / HighestPossibleScore) * 3) : 0;
+        let waitScaled = stats.waitRankCount ? Math.trunc((1 - stats.waitRank / stats.waitRankCount) * 3) : 0;
 
         for (let i = 0; i < downstreamStops.length - 1; i++) {
           // for each stop
--- a/frontend/src/components/MapSpider.jsx
+++ b/frontend/src/components/MapSpider.jsx
@@ -199,7 +199,7 @@ class MapSpider extends Component {
         // Add a base polyline connecting the stops.  One polyline between each stop gives better tooltips
         // when selecting a line.
 
-        const stats = routeStats[startMarker.routeId];
+        const stats = routeStats[startMarker.routeId] || {};
 
         // scale wait score to 0, 1, or 2
         let waitScaled = (stats.medianWaitScore != null) ? Math.trunc((stats.medianWaitScore / HighestPossibleScore) * 3) : 0;
--- a/frontend/src/components/MapSpider.jsx
+++ b/frontend/src/components/MapSpider.jsx
@@ -18,9 +18,9 @@ import {
   getDownstreamStopIds
 } from '../helpers/mapGeometry';
 import {
-  getAllWaits,
   filterRoutes,
   milesBetween,
+  HighestPossibleScore
 } from '../helpers/routeCalculations';
 import { handleSpiderMapClick } from '../actions';
 import { Agencies } from '../config';
@@ -182,10 +182,7 @@ class MapSpider extends Component {
    * Rendering of route from nearest stop to terminal.
    */
   DownstreamLines = () => {
-    const allWaits = getAllWaits(
-      this.props.precomputedStats.waitTimes,
-      this.props.routes,
-    );
+    const routeStats = this.props.routeStats;
 
     // One polyline for each start marker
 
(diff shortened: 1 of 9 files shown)


frontend/src/helpers/routeCalculations.js
```diff
@@ -503,41 +458,54 @@ export function getAllScores(routes, waits, speeds) {
...
 
-export const quartileBackgroundColor = d3
-  .scaleThreshold()
-  .domain([0.25, 0.5, 0.75])
-  .range([red[300], yellow[300], lightGreen[700], green[900]]);
+export const scoreBackgroundColor = score => {
+  if (score == null || Number.isNaN(score)) {
+    return null;
+  }
+  if (score <= 25) {
```
exxonvaldez: Should the thresholds of 25, 50, and 75 be relative to HighestPossibleScore (e.g., 0.25 * HighestPossibleScore), so that they scale with HighestPossibleScore?   The d3 APIs are nice if we ever need support for linear gradients or curves, as opposed to a simple step function, any particular reason for replacing them?

youngj: I created this function in order to handle `null` scores differently than the d3 API. With only 4 possible colors it seems easier to write in plain JavaScript, and easier to understand for anyone who is not already fluent in the d3 API.  If the functionality changes to output color gradients we could use a library for that. Could change to be relative to HighestPossibleScore.

exxonvaldez: I’d prefer to keep the d3 form, as it’s faster for making experimental changes (e.g., iterating through color schemes, or changing the number of thresholds, or the function type).  For custom null handling, you can add a separate wrapper function that takes one argument and invokes this function.

youngj: OK, done.

exxonvaldez:
I got a javascript crash on the Dashboard.  I selected some lines on the map and then changed the end date in the date picker.

Screen Shot 2020-01-14 at 6 29 47 PM

youngj:
Thanks for finding that, should be fixed now.

frontend/src/components/MapSpider.jsx
```diff
@@ -202,16 +199,10 @@ class MapSpider extends Component {
...
-        );
+        const stats = routeStats[startMarker.routeId] || {};
 
-        // scale wait rank to 0, 1, or 2
-        let waitScaled = Math.trunc((waitRank / allWaits.length) * 3);
-        if (!isFinite(waitScaled)) {
-          waitScaled = 0;
-        }
+        // scale wait score to 0, 1, or 2
+        let waitScaled = (stats.medianWaitScore != null) ? Math.trunc((stats.medianWaitScore / HighestPossibleScore) * 3) : 0;
```
exxonvaldez: This change is interesting, now that I've tested it for Muni on a range of days, I see that almost all lines probably have median wait scores greater than 33.  The net result is the map shows lots of thick and medium lines, and almost no thin ones.  Is this change meant to show actual "wait" health day to day, almost like the colored backgrounds on the median wait table cells?

My idea of using rankings to determine widths was to try to sort lines into a visual hierarchy.  This was inspired by the SFMTA Muni map:  https://www.sfmta.com/sites/default/files/pdf_map/2019/06/sfmta-metro-06.12.2019-web.pdf  .  Notice that they bucket bus headways into 0-10, 10-20, and 20-30 and make blue lines of different widths for them.   I don't know how accurate those headways really are when they colored the lines, but they did a good job of making major routes stand out vs a lot of the minor ones.   Seems like real world Muni has lots of lines with similarly good headways... when I click near Market and Montgomery for 1/9 all day, I get 17 lines scoring 75 or more, 13 scoring 50-75, 4 scoring 25-50, and 3 scoring 0-25.

So I guess it depends on what we are going for here.  It is true that the original code was a bit of an outlier for using rank, but visually I liked the result.  Also because the rankings into "third-iles" was generally consistent from day to day, it made it easier to find specific routes.

youngj: It seems more consistent to me if the visual display of a route is only dependent on that route's performance (like the scores and colors in the route table, as well as the linked SFMTA PDF map), and not the performance of other routes (as happens when using rankings).  I don't have a strong preference for exactly what wait times correspond to what widths, but how about just using 3 buckets based on median wait times of 0-5 min, 5-10 min and 10+ min like the SFMTA PDF map?

exxonvaldez: From looking at the routes table, those thresholds would put most routes in the 5-10 bucket, with only a few routes in 0-5 and some in 10+.  To clarify what I saying previously, the visual display of routes on the spider map is not meant to show route performance.  If it were, we would also use color to show performance as we do on the route summary map.  The spider map is focused on differentiating between lines, especially when they have overlapping segments.  This is the use case where a user is exploring routes near a point.  So I still think the wait ranking is best way to distribute the three line widths across the routes, a scheme inspired by the Muni map but not literally copying it.

In the future, I would like to have more data visualizations that use the map (for one route or multiple routes).  Then we could use every aspect of the route display (color, line patterns, widths, etc) to show performance.

youngj: I still think an absolute measure would be better than a relative measure, but since this UI change is unnecessary to the refactoring, I have restored the previous behavior for now.

PR #507, opened 2020-01-12
dispatch actions to update URLs using redux-first-router instead of manually updating path

### LINKED ISSUE #408: Path class does not update URL correctly
In ControlPanel.jsx, the following code in onSelectFirstStop does not add the "from" stop to the URL if secondStopId is already set:

```
    const path = new Path();
    path.buildPath(FROM_STOP, stopId);

    if (secondStopId) {
      path.buildPath(TO_STOP, secondStopId);
    }

    path.commitPath();
```

It seems like the Path class in routeUtil.js is a fragile and limited way of updating paths since it requires all path parameters (like route ids) to be preceded by a static path token like /route/. It doesn't allow adding parameters that are not preceded by static path tokens like /muni/1/0/2234/3332 . It would also fail if the static token like "route" was itself the value of a path parameter.

routesMap.js already defines the template string ROUTESCREEN.path so there should be an easy way to construct paths from this by passing in the parameters (perhaps there is a function in Redux first router to do this?)

This approach would allow simplifying the code by removing the Path class entirely, add more flexibility in how paths can be structured, and avoid bugs.

youngj:
Fixes #408.

To update the URL when using redux-first-router, we can dispatch an action with the `type` equal to the key in routesMap.js and the `payload` object containing the named parameters in the path. 

This also allows removing the unneeded static path segments `direction`, `fromStop`, and `toStop`.


frontend/src/components/ControlPanel.jsx
```diff
@@ -97,17 +94,14 @@ function ControlPanel(props) {
...
-    path
-      .buildPath(ROUTE, routeId)
-      .buildPath(DIRECTION, directionId)
-      .commitPath();
-    props.handleGraphParams({
-      routeId,
-      directionId,
-      startStopId: null,
-      endStopId: null,
+    props.dispatch({
```
ckrajewski: @youngj 
Ah so you can dispatch the prop directly so that redux-first-router handles it yeah?
Instead of using actionToPath right?

youngj: Correct, actionToPath is not needed.

PR #510, opened 2020-01-18
update readme to make it more applicable to non-SF developers, and add more information on the backend stack

youngj:
The readme was a bit confusing for new Portland developers since it sounded like it was only for SF.

Also, it had a lot of information about the technologies used in the frontend stack, but not much information about the backend stack.


SeannyPhoenix:
This looks great, Jesse.

PR #508, opened 2020-01-16
Adding Breadcrumbs UI

ckrajewski:
## Proposed changes

Taking @youngj changes and adding UI  portion (breadcrumbs)

Fixes #160

## Screenshot

image

## Link to demo, if any

...

PR #514, opened 2020-01-30
471 - add official link icon

### LINKED ISSUE #421: Add links to official pages for routes and stops when available
Some GTFS feeds contain URLs for each route (http://trimet.org//schedules/r100.htm) and each stop (e.g. https://trimet.org/#tracker/stop/9826). 

backend/save_routes.py is now storing these URLs in the route configuration in S3, so that it is available in JavaScript as the "url" property of the route and stop info objects. 

If these URLs are available for a route or stop, it would be nice to put links somewhere in the UI so that people could easily go to the official page for a route or stop.

JuanCaicedo:
**Note**: This PR builds on #508 and is blocked until that merges. To make reviewing easier, I opened this PR against that one. After #508 merges, I'll change the base branch to master 👍

## Motivation
- Fixes #421

## Proposed changes

- Adds a clickable icon to next to the route name for routes that have official links

## Screenshot

image


frontend/src/screens/RouteScreen.jsx
```diff
@@ -35,7 +36,15 @@ const useStyles = makeStyles(theme => ({
...
       paddingRight: '0'
-    }
+    },
+  externalLink: {
+    color: 'rgba(0, 0, 0, 0.54)',
+    marginLeft: '5px'
+  },
+  externalLinkIcon: {
+    position: 'relative',
+    top: '5px'
```
JuanCaicedo: I'm not really sure what's going on here, but adding the icon svg leads to some weird alignments 😅

It causes the whole line to grow to `33px` from the original `28px`.
image

And also the `a` tag doesn't seem to wrap completely around the icon, which sits higher than the link box
image

This makes the icon not aligned with the text next to it.
image

Thought I tried to figure all this out, it seemed easier to just visually center the icon 😅

hathix:
One UX question: does that link icon go to the same place as the blue link? My initial expectation as a user is that they both go to the same place. If they don't, you may want to change the icon to make it more obvious that it's a different link. Perhaps changing it to a circled (I) would make it clearer.

hathix:
We merged #508 so you can rebase this :)

exxonvaldez:
Looks pretty good, thanks for doing this!  I found that using the SVG icons with text was pretty difficult also, and the relative position "trick" seems like the best way to go that I've found.  I noticed that the breadcrumbs are just slightly out of alignment vertically, I think it's because the icon makes the first breadcrumb taller than the others.  Not sure what the best fix is for this.

Screen Shot 2020-01-29 at 7 45 25 PM

youngj:
Agree with @hathix - with this UI, I would expect that clicking either the icon or the adjacent link will open a page in a new window. It would make more sense to me to have a text link that says something like "Official Site" next to the "open in new window" icon. This would probably be too much to display within the breadcrumb text itself, but could be floating on the right side of the white bar.

JuanCaicedo:
I made a few changes here 😀
- Change the open in new window icon to `Official Site`
- Align the link to the right side of the nav bar

image

frontend/src/screens/RouteScreen.jsx
```diff
@@ -34,7 +34,15 @@ const useStyles = makeStyles(theme => ({
   },
   breadCrumbsWrapper: {
     padding: '1%',
-    paddingRight: '0',
+    display: 'flex',
+    justifyContent: 'space-between',
```
JuanCaicedo: This looks a little weird on mobile 
image

I think solving it could be as simple as adding a media query that sets `flexDirection: column` for mobile sizes, but I'm not sure how to do media queries within Material UI 😅 The strategy I found [in their docs](https://material-ui.com/components/use-media-query/) seems complicated, which makes me think I might be looking at the wrong approach

hathix: Nah, don't overcomplicate it.

The more elegant solution, to me, is to make that text label only visible on medium and large screens (you can do that with tools from [here](https://material-ui.com/components/grid/#fluid-grids), i think) and replace it with a small icon on small screens.

hathix:
You got a lint error, by the way. See the `test / test` result

hathix:
Anyway, I think your approach of using the "Official Site" on the right side was great. Much better from a UX perspective.

PR #522, opened 2020-01-30
made entire row clickable

### LINKED ISSUE #396: Clicking on entire row should open route information, instead of just the route name.
The only way to get to route data is to click on the route name. Instead, any click on the entire row should open the route data.

image

allisonbusse:
Fixes #396 

## Proposed changes
Added link to table row instead of just first table cell
...

## Link to demo, if any
https://open-transit-row.herokuapp.com/
...


hathix:
It looks like you can click on the colored chips, but the mouse stays looking like a "pointer" instead of a "hand". The mouse does look like a "hand" over the rest of the row. Perhaps some small CSS tweak is needed.

Issue #511, opened 2020-01-22
GraphQL API for precomputed statistics over a date range


youngj:
Currently, the frontend directly requests JSON files of precomputed statistics from S3. These statistics are shown in the RouteTable on the Dashboard, and are also used in the RouteSummary cards, the route segments in MapStops, and the travel time chart on the RouteScreen. 

However, the frontend only requests statistics for the end date.  To display statistics within a date range, one option would be for the frontend to request the precomputed statistics for each date from S3, but this would likely not have good performance because the precomputed wait times and trip time files are fairly large, and this would require more and more data to be sent from S3 to each user as the date range gets larger.

To allow displaying these statistics for multiple dates, the backend should provide a GraphQL API that allows the frontend to request statistics for a range of dates. The backend could request the precomputed stats from S3 for each date, compute the median of the requested statistics over each date, and return the requested data to the frontend (without sending a large amount of data for wait times and trip times that the frontend doesn't actually use). This would allow the backend to cache the files containing precomputed stats instead of downloading them from S3 each time. We could also potentially cache the result of each GraphQL query in nginx.

This also requires moving the averageSpeed computation from the frontend to the backend, and using the heuristics for ignoring start/end stops in the backend.

PR #526, opened 2020-01-30
Temporarily limit date range queries on the front end to 10 days

exxonvaldez:
## Proposed changes

Temporarily limit date range queries on the front end to 10 days instead of 90.  Date range queries of 30 days or more seem to tie up the back end for too long and may be causing Flask worker timeouts.

I think Jesse is planning to optimize the graphql code, possibly as part of #511 , so these changes can be reverted when that goes in.

## Screenshot

n/a

## Link to demo, if any

n/a

### DIFF
--- a/frontend/src/UIConstants.js
+++ b/frontend/src/UIConstants.js
@@ -68,14 +68,18 @@ export const TIME_RANGES = [
   },
 ];
 
-export const MAX_DATE_RANGE = 90; // largest date range allowed, in days (30 might be more performant)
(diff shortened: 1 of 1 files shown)


hathix:
Would #528 make this obsolete?

exxonvaldez:
> Would #528 make this obsolete?

Hard to know for sure.  30 days seems to work pretty well on the heroku demo.  60 days was on the slow side.  There isn't enough data to test 90 days.

exxonvaldez:
Looks like we don't need this.

PR #527, opened 2020-01-30
fix react-select event listener cleanup

trxaphion:
Replaces @exxonvaldez 's hotfix (#525 ).

## Proposed changes

I didn't understand at the time that the global object was quite so...global. As a result only the second react-select's listeners were being cleaned up.

## Screenshot
Both screenshots are after returning to dashboard
<table>
<tr><td>Before</td><td>After</td></tr>
<tr>
<td></td>
<td></td>
</tr>

### DIFF
--- a/frontend/src/components/ReactSelect.jsx
+++ b/frontend/src/components/ReactSelect.jsx
@@ -408,11 +408,11 @@ function filterValue(stopId) {
   return option => option.value.stopId === stopId;
 }
 
+const reposition = {};
 /**
  * updates textfield location on scroll/resize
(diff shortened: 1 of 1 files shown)


trxaphion:
Yea I'm currently not on the team. I did leave my username on the Slack channel at one point but I guess it got drowned out.

exxonvaldez:
Just did a quick test, looks good.  Usually we have the "opener" of the request do the merge, in case any last minute tweaks (to account for the effects of any recent merges to master) or testing are needed.  Other times Neel will just merge approved PRs in all at once, but that can lead to bugs in production.

PR #528, opened 2020-02-02
GraphQL API for precomputed statistics over a date range; replace Long Wait % with On Time %; improve date range performance

### LINKED ISSUE #297: Show on-time performance
Say early buses are <1min early, late buses are >4min late. Let users see this for each route. Let them slice & dice by:

* Time of day
* Day of week
* Week of year
* Stop
* Direction

Also let users see a histogram of the delta between real & scheduled arrival.

This obviously depends on us drawing in scheduled data better.

### LINKED ISSUE #511: GraphQL API for precomputed statistics over a date range

### LINKED ISSUE #512: Precompute on-time rate and show in route table
It would be useful to display the average on-time rate for each route on the table on the Dashboard. In order to do this we would need to precompute the on-time rates for each interval (similar to how we precompute wait times) and find the median on-time rate for all stops in a route. This could be combined with https://github.com/trynmaps/metrics-mvp/issues/511 so the frontend could request on-time rate statistics for a date range from a GraphQL API without the frontend needing to fetch the precomputed stats directly from S3.

### LINKED ISSUE #513: When a route and direction is selected, the Summary tab should show statistics for that direction
Currently when a route and direction is selected on the route screen, the Summary tab displays averages for all directions for that route, not for the selected direction.

youngj:
Fixes #511.
Fixes #512.
Fixes #513.
Fixes #297.

Previously, the frontend directly requested JSON files of precomputed statistics from S3. These statistics were shown in the RouteTable on the Dashboard, the RouteSummary cards, the route segments in MapStops, the travel time chart on the RouteScreen.

However, the frontend only requested statistics for the end date. To display statistics within a date range, it would likely have poor performance if the frontend requested the precomputed statistics for each date from S3, because the precomputed wait times and trip time files are fairly large, and this approach would require more and more data to be sent from S3 to each user as the date range gets larger.

To allow all of these screens to support date ranges, this PR updates the GraphQL API to allow the Python backend to compute statistics over a date range and return the data to the frontend.

Since it would require additional work to implement Long Wait % with the new GraphQL API, and we were already planning to replace Long Wait % with On Time % anyway, this PR implements On Time % using the updated GraphQL API and replaces all instances of Long Wait % with On Time %.

Since this kind of architectural change is difficult to do in small pieces, this PR ended up being quite large.

**Updated GraphQL API**

The fields in the GraphQL API were reorganized to support the various queries that are needed by the frontend, while also allowing the backend to cache data that may be needed to compute multiple fields. 

The below annotated example query illustrates the overall structure of the GraphQL API:

```graphql
query {
  agency(agencyId:"muni") { 
    # top-level field lets all subfields use the same metrics.AgencyMetrics object
    agencyId
    interval(dates:["2020-01-27","2020-01-28"]) { 
      # high-level stats for all routes in agency shown in RouteTable on dashboard
      routes {
        # list of all routes in the agency
        routeId
        directions {
          # list of stats for each direction of this route.
          # frontend averages these stats over all directions when displaying data for a route.
          directionId      
          medianWaitTime
          averageSpeed
          travelTimeVariability
          onTimeRate
        }
      }
    }
    route(routeId:"1") {
      # more detailed metrics for an individual route.
      # used to be the top level field routeMetrics.
      interval(dates:["2020-01-27","2020-01-28"]) { 
        # overall statistics for the entire route in an interval, e.g. trip times
    	directions {
          # list of stats for each direction in the route
          directionId
          segments {
            # information about trip times between adjacent stops
            # (used by MapStops)
            fromStopId
            toStopId
            medianTripTime
          }
          cumulativeSegments {
            # information about trip times from first stop in route
            # (used by TravelTimeChart)
            fromStopId
            toStopId
            medianTripTime
          }
        }
      }
      trip(startStopId:"16304", endStopId:"16309", directionId:"0") {
        # metrics for a particular trip, like before
        byDay(dates:["2020-01-27","2020-01-28"]) {
          # used on 'by day' tab.
          # byDay stats now have same fields as interval and timeRanges
          waitTimes {
            median            
          }
          tripTimes {
            median
          }
          # etc.
        }
        interval(dates:["2020-01-27","2020-01-28"]) {
          # same as before
          waitTimes {
            median            
          }
          tripTimes {
            median
          }
        }        
        timeRanges(dates:["2020-01-27","2020-01-28"]) {
          # used for 'by time of day' tab, same as before
          waitTimes {
            median            
          }
          tripTimes {
            median
          }
        }        
      }
    }
  }
}
```

In practice, the frontend actually uses 3 different GraphQL queries with different fields, instead of requesting everything in one GraphQL query:

* The action `fetchAgencyMetrics` fetches the top level statistics shown in the route table and route summary whenever the date range changes.
* The action `fetchRouteMetrics` fetches the route-level data shown in MapStops and TravelTimeChart when the selected route or date range changes.
* The action `fetchTripMetrics` fetches the trip-level data when the start/end stops or date range change.

The top-level fields for returning route configuration were removed because the frontend currently does not use them (since it gets the route configuration from S3).

Previously the byDay field used precomputed stats, which profiling showed was faster than computing statistics from the raw arrival times. However, in practice, the frontend requests the "interval" field and the "byDay" field at the same time. Since the interval field was already computing statistics from the raw arrival times for the entire date range, it seems like we might as well just let byDay work the same way. By updating metrics.RouteMetrics object to cache the values that were calculated for each date in the range, byDay and interval fields can reuse the results of computations for wait times, trip times, etc. without having to redo them twice. 

This change lets the byDay field have all the same properties as the interval field, rather than being limited to only precomputed statistics.

The API documentation has not been updated yet, but if this structure looks good I would update the API documentation before merging.

**Changing how precomputed statistics are computed and stored**

Previously there were separate scripts for precomputing wait time statistics and trip time statistics, as well as separate functions for storing/retrieving them from S3. Adding on-time rates required storing a new kind of precomputed statistics. In the future we'll probably want to add more precomputed statistics for things such as headways, dwell times, bunching/gapping, as well as scheduled wait times, trip times, headways, etc. It would be unwieldy if all of these types of statistics needed their own scripts and helper functions. 

This PR combines the functionality in compute_wait_times.py and compute_trip_times.py into compute_stats.py . Each file containing precomputed statistics may contain multiple different types of statistics, like wait times, trip times, and on time rates. However, it is still possible to store different precomputed statistics in separate files. 

Currently there is one "combined" file which contains all the precomputed stats used by the dashboard and route screen (including median wait times and on time rates for all stops, as well as median trip times, 10th/90th percentile trip times, and number of trips for a subset of stop pairs). There is another "median-trip-times" file which contains only median trip times for all stop pairs and is used by the isochrone screen. 

**Performance improvements**

Based on profiling, much of the backend execution time is involved loading and parsing large JSON files such as the precomputed stats or the route configuration. To improve performance, it is important to avoid loading/parsing each JSON file more than once per request, and reducing the size of these files if possible.

The new metrics.AgencyMetrics object caches route configuration and precomputed stats so that each file only needs to be loaded once per request.

Storing trip times for every pair of stops creates quite large JSON objects, but currently only the isochrone map actually uses trip times for each pair of stops. The other screens only need trip times for adjacent stops, or between the first stop of the route and other stops (although the "first" stop is sometimes changed by the ignoreFirstStop setting). To improve performance, the precomputed stats used by the GraphQL API now only contain trip times for a subset of stop pairs. However, the ignoreFirstStop setting makes things more complex because it might change over time, which could require recomputing old stats if the stats did not contain all stop pairs starting from the new first stop. This PR stores all pairs of stops starting with the first 3 stops in the route, so that it is possible to change ignoreFirstStop to any of the first 3 stops without recomputing old statistics.

**Using on-time rate in total score**

Since Long Wait % is removed, this PR replaces it in the "total score" calculation with On Time %, where the component of the score is equal to the on-time percentage (already on a scale from 0 to 100). It should concern anyone that something as prominently displayed as the total score could be changed so arbitrarily. Yet this change is no more arbitrary than the original calculation (where 25% of the total score was based on the percentage of wait times that were longer than 20 minutes, where a rate of 10% or less resulted in a perfect score and a rate of 33% or more resulted in a zero score). This change should not be considered as an endorsement of showing scores in the app or the particular scoring algorithm that is currently used. 

Having a "total score" makes it difficult to customize the fields that are displayed in the app or optimize performance because the frontend needs to always fetch all of the properties that have been selected as components of the total score. However, this PR leaves the total score for now so that we can continue the discussion about removing scores later.

**Show direction-specific statistics**

Previously, when a route and direction was selected on the route screen, the Summary tab displayed averages for all directions for that route, not for the selected direction. Now it shows the data for the selected direction, as the user would expect.

**Updated travel time chart**

Added the number of trips above the travel time chart to help figure out if the ignoreFirstStop/ignoreLastStop heuristics for SF Muni were working as expected. (Previously, for some routes, the travel time chart and the average speed and travel time variability statistics were only based on a very small percentage of the total trips, but there was no way to see that this was happening.)

Since the frontend must request the cumulativeSegments GraphQL field for a specific route, it is no longer possible for the data diagnostic screen to show travel time charts for all routes at once (unless the frontend is changed to fetch data for all routes at once, which didn't seem worth doing only for the sake of the data diagnostic screen).

**Muni first/last stop heuristics**

Updated the ignoreFirstStop/ignoreLastStop heuristics for Muni so that the Travel Time chart, Average Speed, and Travel Time Variability statistics are not affected by outliers at the start and end of the route. The frontend JS no longer uses the ignoreFirstStop/ignoreLastStop heuristics directly, but I didn't refactor them in this PR. 

The way the ignoreFirstStop/ignoreLastStop properties work is confusing (i.e., if it is `true`, then ignore the first/last stop; if it is a string, then use that stop ID as the first/last stop). There should probably be an issue to make these clearer.


youngj:
Demo is at http://ot-agency-metrics.herokuapp.com/ . It currently has precomputed stats for 1/22/2020 to 2/1/2020, so if you are viewing later you may need to change the date in order to see any data.

To compute stats like median wait time, average speed, on time rate, and travel time variability over a date range, the backend computes the median of the precomputed statistics for each date in the range (e.g. the median wait time for the range is the median of the median wait times for each date). The median of medians for each date is not actually the overall median. However, I don't think it would be feasible to compute the actual median over an arbitrary date range in a reasonable amount of time. 

An alternative would be to use the mean of the statistics for each date, which would be more sensitive to dates with outlier values.

exxonvaldez:
This is impressive, nice work!  I like the approach of using slimmed down precomputed json files for faster computation of route statistics over date ranges with GraphQL.

I've finished a first reading of the front end changes, and also a close reading of the above writeup, thanks for providing that.  I've also tested out the demo.  So far I don't see any immediate or high level issues.

Do I understand correctly that the "by day" query field (not sure what the right GraphQL term is for the `byDay(...)` part of the query is) can be used with agency metrics and route metrics?

It's neat to see the 30 day and 90 day queries working on the dashboard, route, and trip screens.   We may need to show a progress bar for longer queries (can be a new Github issue).

The substitution of OTP for long wait makes sense.  On my backlog is a task to look at how to combine or blend OTP with bunching and gapping when scoring, as Tal from SFMTA suggested.

We can create a new issue for refining how OTP is scored.  The TCQSM (http://www.trb.org/Main/Blurbs/169437.aspx) page 5-30, has rough tiers for OTP:

Screen Shot 2020-02-05 at 4 21 56 PM

What's interesting is this metric is far more context sensitive than the other metrics.  The SF City Charter calls for an OTP of 85%.  In practice the system stat is in the 50s.  The TCQSM says 70% may be the best that's possible where congestion is a factor, which would be true for SF.  The TCQSM also talks about "headway adherence" for frequent lines. It's a pretty complicated stat, so I'll just show the table of tiers:

Screen Shot 2020-02-05 at 4 29 01 PM

I'll continue reviewing this PR during tonight's hack night.

youngj:
> Do I understand correctly that the "by day" query field (not sure what the right GraphQL term is for the byDay(...) part of the query is) can be used with agency metrics and route metrics?

"field" is the correct term, "query field" is also good https://graphql.org/learn/queries/#fields

I haven't implemented the byDay field on the AgencyMetrics or RouteMetrics GraphQL types, but it would be straightforward to add if needed.

> It's neat to see the 30 day and 90 day queries working on the dashboard, route, and trip screens. We may need to show a progress bar for longer queries (can be a new Github issue).

I've only computed stats for 1/12/2020 to 2/4/2020 (24 days) so far. 

The first time the GraphQL API needs precomputed stats for a particular date, it will fetch them from S3. On subsequent requests, it will use the cache file from the local disk. However, if the precomputed stats files don't exist yet for certain days in the range, the backend will try to load them from S3 on every request. The average load time will likely be different depending on whether the statistics are already cached locally or not, or if the statistics don't exist in S3.

Based on some limited performance testing on http://ot-agency-metrics.herokuapp.com/ it seems like each day currently adds roughly 65ms to the runtime for the agency metrics query on average if the precomputed stats are already cached locally. In other words, each additional 15 days in the date range would be expected to add about 1 second to the load time for the dashboard. There are probably ways we can make this faster, like using Redis to cache the precomputed stats in memory so the JSON data doesn't need to be loaded from disk and parsed on every request.

The route screen will be slower than the dashboard because there is another GraphQL query for precomputed route stats and possibly another GraphQL query for trip metrics.

exxonvaldez:
> I haven't implemented the byDay field on the AgencyMetrics or RouteMetrics GraphQL types, but it would be straightforward to add if needed.

By day for the route summary is something I wanted to have for sure.  For the dashboard, I'm also interested in using for displaying system health and how it varies by day.

> > It's neat to see the 30 day and 90 day queries working on the dashboard, route, and trip screens. We may need to show a progress bar for longer queries (can be a new Github issue).
> 
> I've only computed stats for 1/12/2020 to 2/4/2020 (24 days) so far.

Oh, okay.  So if a query spans uncomputed (or future) days and computed days, it will silently shorten the date range to the available days.  We should show a warning in that case, so that users know about the effective date range.

akgupta89:
Just a note that the `metrics_api.md` should probably be updated with this PR

frontend/src/components/TravelTimeChart.jsx
```diff
@@ -90,10 +131,18 @@ function TravelTimeChart(props) {
...
+      &nbsp;&nbsp; Average speed:{' '}
+      {tripTimeForDirection > 0
+        ? ((60 * distanceForDirection) / tripTimeForDirection).toFixed(1)
+        : '?'}{' '}
+      mph
+      <br />
+      Distance:{' '}
+      {distanceForDirection != null ? distanceForDirection.toFixed(1) : '?'} mi
+      &nbsp;&nbsp; Stops: {numStops > 0 ? numStops : '?'} &nbsp;&nbsp; Completed
+      trips: {numTrips != null ? numTrips : '0'}
```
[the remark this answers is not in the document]
youngj: OK, done

frontend/src/components/RouteSummary.jsx
```diff
@@ -297,21 +295,23 @@ function RouteSummary(props) {
...
         <TravelTimeChart />
       </Box>
-      <Box
-        p={2}
-        hidden={tabValue !== MAREY_CHART}
-        style={{ overflowX: 'auto' }}
-      >
-        <MareyChart hidden={tabValue !== MAREY_CHART} />
-      </Box>
+      {mareyChartSupported ? (
```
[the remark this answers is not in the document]
youngj: OK, done

frontend/src/components/RouteSummary.jsx
```diff
@@ -61,8 +56,8 @@ function RouteSummary(props) {
                 <TableCell align="right">{stats.medianWaitScore}</TableCell>
               </TableRow>
               <TableRow>
-                <TableCell>Long wait probability</TableCell>
-                <TableCell align="right">{stats.longWaitScore}</TableCell>
+                <TableCell>On-time rate</TableCell>
```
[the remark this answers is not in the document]
youngj: done

frontend/src/actions/index.js
```diff
@@ -429,20 +405,16 @@ export function handleGraphParams(params) {
...
     // for debugging: console.log('hGP: ' + graphParams.routeId + ' dirid: ' + graphParams.directionId + " start: " + graphParams.startStopId + " end: " + graphParams.endStopId);
     // fetch graph data if all params provided
     // TODO: fetch route summary data if all we have is a route ID.
 
     if (graphParams.date) {
-      dispatch(fetchPrecomputedStats(graphParams));
+      dispatch(fetchAgencyMetrics(graphParams));
+    }
+
+    if (graphParams.agencyId && graphParams.routeId) {
```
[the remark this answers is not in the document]
youngj: This is because fetchAgencyMetrics will eventually be updated to fetch metrics for all supported agencies at once (e.g. trimet and portland-sc in Portland), not just Agencies[0]. But in order for fetchRouteMetrics to look up metrics for an individual route, it needs to know which agency the route is in.

youngj:
Updated the GraphQL API documentation.

frontend/src/components/RouteSummary.jsx
```diff
@@ -100,33 +95,24 @@ function RouteSummary(props) {
...
-              { label: '33% or more', value: 0 },
-            ]}
-          />
-        </Box>
+        The on-time percentage is the percentage of scheduled departure times
+        where a vehicle departed less than 5 minutes after the scheduled
+        departure time or less than 1 minute before the scheduled departure
+        time. Probability of{' '}
+        {(stats.onTimeRate * 100).toFixed(1) /* be more precise than card */}%
+        gets a score of {stats.onTimeRateScore}.
```
EddyIonescu: I think this should be made clearer, as I first thought that the on-time performance was calculated for all stop arrivals in the range, rather than just for the first stop. Would this work better?
> On-Time % is the share of scheduled trips that departed less than 5 minutes after and less than 1 minute before their scheduled departure time from the first stop of the route section.

frontend/src/components/MareyChart.jsx
```diff
@@ -249,6 +251,12 @@ function MareyChart(props) {
     }
   }, [arrivals, routes, timezoneId]);
 
+  if (!mareyChartSupported) {
+    return (
+      <div>The Marey chart is only available when viewing a single date.</div>
+    );
+  }
```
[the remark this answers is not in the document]
youngj: In the original PR, the Marey chart was originally hidden if not supported, but now the Marey Chart tab is always visible, after Terence's feedback.

frontend/src/components/RouteSummary.jsx
youngj: It's not just the first stop -- the on-time percentage for the entire route is the median of the on-time percentage for each stop along the route. Updated the text to clarify.

EddyIonescu: Hey, so I'm a bit confused here. When I set a stop range but change the end stop, the on-time % never changes - but it always does when I change the start stop.
Could you point me to where the median's being calculated? All I found is the resolver (https://github.com/trynmaps/metrics-mvp/blob/cf21ce076e9776fa2fa9e0c55ec3859a58315881/backend/models/schema.py#L407) which looks like it's only getting schedule adherence for the start stop?

youngj: That is correct, the On-Time % shown here is the departure schedule adherence, so it only varies when the start stop changes. The GraphQL API also provides arrival schedule adherence (https://github.com/trynmaps/metrics-mvp/blob/master/docs/metrics_api.md#tripintervalmetrics) but it is not currently shown in the web app.

youngj: My comment from two days ago ("It's not just the first stop") was in reference to the RouteSummary component where the On-Time % is the median departure schedule adherence over all stops. When you are looking at an individual trip, it only shows the departure schedule adherence for the one stop that you are actually departing from.

EddyIonescu: That was fast! I see, so by all stops do you mean all the stops in the route/direction or just the stops from first stop to last stop?

So I was referring to RouteSummary, as I don't see where the frontend would be making calls for each stop as to get the on-time median (and calculating it) for all stops - the RouteSummary component seems to be showing the on-time performance for the first stop in the range regardless of the end stop.

youngj: It seems like you are actually referring to the InfoTripSummary component. The RouteSummary component is not displayed when you select a first stop / last stop.

The RouteSummary component uses precomputed statistics with the median on-time percentage over all stops. These statistics are calculated at https://github.com/trynmaps/metrics-mvp/blob/master/backend/compute_stats.py#L120 .

PR #509, opened 2020-01-16
399 return to home

melismae:
## Proposed changes

This PR proposes to add a "Return to Dashboard" link on the route map page. 

Fixes #399 

Testable here: https://trynmaps-melissa.herokuapp.com

Screen Shot 2020-01-15 at 9 00 58 PM

PR #529, opened 2020-02-02
add scripts for running lint within docker container; run lint via github actions

youngj:
Currently it seems that most frontend developers (including me) haven't been running `npm run lint` on their PRs. If one developer runs lint on their PR, lint will then make a lot of changes unrelated to their PR, which makes reviewing and merging difficult (e.g. https://github.com/trynmaps/metrics-mvp/pull/509).

However, in order for a developer to run `npm run lint`, they need npm installed on their host, but many developers do not have npm installed on their host, because normally npm runs inside a Docker container.

This PR adds scripts to allow developers using Docker to run lint inside a Docker container, so it is no longer necessary for developers to have npm installed on their host.

It also runs the lint task via Github Actions on each push so that we can see failing lint checks when reviewing PRs, and fixes all existing lint errors.

### DIFF
--- a/frontend/.eslintrc.json
+++ b/frontend/.eslintrc.json
@@ -24,11 +24,17 @@
         "ignoreUrls": true
       }
     ],
+    "no-param-reassign": [
+      "error",
+      {
+        "props": false
+      }
+    ],
     "func-names": "off",
     "prefer-destructuring": "off",
     "no-plusplus": [
       "error",
-      { "allowForLoopAfterthoughts": true } 
+      { "allowForLoopAfterthoughts": true }
     ],
     "no-underscore-dangle": "warn",
     "react/destructuring-assignment": "off",
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,3 +11,7 @@ jobs:
       run: docker build --target flask -t opentransit/flask:latest .
     - name: Run tests
       run: docker run opentransit/flask:latest python -m unittest discover -s tests
+    - name: Build Docker react-dev image
+      run: docker build --target react-dev -t opentransit/react-dev:latest .
+    - name: Run lint
+      run: docker run opentransit/react-dev:latest npm run lint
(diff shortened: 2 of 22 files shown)


youngj:
I just fixed the ~40 or so errors that were reported by `npm run lint`, so the checks will be passing after this is merged into master. This will create some merge conflicts in https://github.com/trynmaps/metrics-mvp/pull/528 which I'll fix once someone approves this.

frontend/.eslintrc.json
```diff
@@ -24,11 +24,17 @@
         "ignoreUrls": true
       }
     ],
+    "no-param-reassign": [
+      "error",
+      {
+        "props": false
```
exxonvaldez: Is this change needed?  If I understand the docs right  for this, isn't props: false the default?  https://eslint.org/docs/rules/no-param-reassign

youngj: I noticed that the docs say props: false is the default, but the docs don't seem to be correct because lint was reporting several errors before adding this (https://github.com/trynmaps/metrics-mvp/runs/422180446): 

```
/app/frontend/src/components/ReactSelect.jsx
  427:9  error  Assignment to property of function parameter 'menuTransition'  no-param-reassign

/app/frontend/src/reducers/index.js
  22:7  error  Assignment to property of function parameter 'map'  no-param-reassign
  26:7  error  Assignment to property of function parameter 'map'  no-param-reassign

/app/frontend/src/routesMap.js
  43:11  error  Assignment to property of function parameter 'newDaysOfTheWeek'  no-param-reassign
  45:11  error  Assignment to property of function parameter 'newDaysOfTheWeek'  no-param-reassign
```

exxonvaldez: Oh, I see... it's because we extend airbnb and they use props: true.  https://github.com/immerjs/immer/issues/189 .  Ok, so we definitively want this.

youngj:
Sounds good, just added a section about the docker-lint scripts to README.md.

PR #537, opened 2020-02-06
render route on map on table row hover

### LINKED ISSUE #389: Dashboard: draw line on map when you hover over line name
On the right side of the homepage (dashboard), you see a bunch of line names. It would be cool to draw that line on the map (or highlight it, if the line is already up) when you hover over that line's name.

So if you hover over 38R in this case, the 38R's line would be drawn.

Screen Shot 2019-11-06 at 6 58 32 PM

trxaphion:
Fixes #389 

## Proposed changes

Render route on map when hovering over route table. When no spider exists both directions are rendered. When spider exists only render directions already on the spider. 

Currently on hover, the starting stop (closest stop to spider marker) of the route and stops are fully opaque whilst polylines are slightly less transparent (only when spider exists), and of course everything is slightly larger. I find that when the polylines are too transparent you sometimes see a mess of colors underneath making it harder to see but when it's too solid you can't see the stops. Thoughts?

Also the map shields of other routes currently stay on top of the hovered route (except the shield). Is this preferable?

## Link to demo, if any

https://open-transit-hover-line.herokuapp.com/

### DIFF
--- a/frontend/src/components/MapSpider.jsx
+++ b/frontend/src/components/MapSpider.jsx
@@ -15,16 +15,18 @@ import Control from 'react-leaflet-control';
 import * as d3 from 'd3';
 import { Snackbar } from '@material-ui/core';
 import {
-  getDownstreamStopIds
+  getDownstreamStopIds,
+  getTripPoints,
+  isInServiceArea,
 } from '../helpers/mapGeometry';
 import {
   filterRoutes,
   milesBetween,
-  HighestPossibleScore
+  HighestPossibleScore,
 } from '../helpers/routeCalculations';
 import { handleSpiderMapClick } from '../actions';
 import { Agencies } from '../config';
-import { getTripPoints, isInServiceArea } from '../helpers/mapGeometry';
+
 import MapShield from './MapShield';
 
 const CLICK_RADIUS_MI = 0.25; // maximum radius for stops near a point
@@ -52,6 +54,7 @@ class MapSpider extends Component {
     // for now, only supports 1 agency at a time.
     // todo: support multiple agencies on one map
     this.agency = Agencies[0];
+    this.getDownstreamLayers = this.getDownstreamLayers.bind(this);
 
     this.state = {
       // Should be true by default, so that we don't display the snackbar
@@ -101,10 +104,12 @@ class MapSpider extends Component {
    * https://medium.com/@nikjohn/creating-a-dynamic-jsx-marker-with-react-leaflet-f75fff2ddb9
    */
   generateShield = (startMarker, waitScaled) => {
+    const { hoverRoute } = this.props;
     const lastStop =
       startMarker.downstreamStops[startMarker.downstreamStops.length - 1];
     const shieldPosition = [lastStop.lat, lastStop.lon];
     const routeColor = this.routeColor(startMarker.routeIndex % 10);
+    const zIndex = hoverRoute ? 200 : 0;
 
     const icon = L.divIcon({
       className: 'custom-icon', // this is needed to turn off the default icon styling (blank square)
@@ -121,6 +126,7 @@ class MapSpider extends Component {
         position={shieldPosition}
         icon={icon}
         riseOnHover
+        zIndexOffset={zIndex}
         onClick={e => {
           e.originalEvent.view.L.DomEvent.stopPropagation(e);
           this.props.dispatch({
@@ -129,8 +135,8 @@ class MapSpider extends Component {
               routeId: startMarker.routeId,
               directionId: startMarker.direction.id,
               startStopId: startMarker.stopId,
-              endStopId: lastStop.stopId
-            }
+              endStopId: lastStop.stopId,
+            },
           });
         }}
       ></Marker>
@@ -140,15 +146,10 @@ class MapSpider extends Component {
   /**
    * Rendering of stops nearest to click or current location
    */
-  getStartMarkers = () => {
+  getStartMarkers = (selectedStops, radius = '8', opacity = 0.2) => {
     let items = null;
 
-    /* eslint-disable react/no-array-index-key */
-
-    var selectedStops = this.props.spiderSelection.stops;
-
     if (selectedStops) {
-
       items = selectedStops.map((startMarker, index) => {
         const position = [startMarker.stop.lat, startMarker.stop.lon];
         const routeColor = this.routeColor(startMarker.routeIndex % 10);
@@ -157,9 +158,9 @@ class MapSpider extends Component {
           <CircleMarker
             key={`startMarker-${index}`}
             center={position}
-            radius="8"
+            radius={radius}
             fillColor={routeColor}
-            fillOpacity={0.2}
+            fillOpacity={opacity}
             stroke={false}
           >
             <Tooltip>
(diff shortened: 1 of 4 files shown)


exxonvaldez:
Thanks for doing this.  This seems like a good enhancement overall, and it is very responsive (which I know can be a challenge with react-leaflet).  I haven't used React.memo before but this seems like a natural place to use it.

I wanted to start with some high level feedback before getting into a detailed review.  On the technical side, I was wondering if we can avoid the use of setTimeout/clearTimeout.  These kinds of global function calls are difficult to understand and debug.  Also, pushing state from Dashboard down to its children is a bit of an antipattern that Redux was meant to solve.  Can you use Redux state to communicate between components and avoid the use of setTimeout?  The spider selection state is done using Redux, you should be able to add stuff alongside that.

On the UI design side, everything a matter of subjective opinion but my two cents: I'd like the see the hover effect done a little differently.  On MapStops, we use a wider white polyline to provide some contrast against the map, and then a wider indigo polyline between the selected stops.  So can we do the hover effect using a white or indigo outline?  I'd also prefer the shield sizes and colored line widths to stay as they were.  The width conveys the frequency of the line, so that information gets lost when the width is scaled up, especially if there is no spider selection.  I'm hoping the white or indigo outline lets you keep the opacities for the colored polylines where there were originally, and no changes to the shields.

It seems natural that the hover effect should go both ways, so when you highlight a route segment, the route's row in the table is also highlighted (the light gray background in the table row, which might be a css effect).

trxaphion:
Thanks for the detailed feedback. 

My rationale behind the route only showing directions within range of the marker was that having a marker down already eliminated whole routes from the table making them unselectable, so the user would equate what is shown on the map to what remains on the table. On the other hand clicking the route involves choosing either direction so if you still think it would be more natural to always show both directions then I don't disagree.

Regarding the timeout, removing it won't have a huge impact but it will be noticeable when moving through the routes consecutively at a somewhat fast rate.

noTimeout

exxonvaldez:
> My rationale behind the route only showing directions within range of the marker was that having a marker down already eliminated whole routes from the table making them unselectable, so the user would equate what is shown on the map to what remains on the table and vice versa. On the other hand clicking the route involves choosing either direction so if you still think it would be more natural to always show both directions then I don't disagree.

I'm not sure what part of my feedback you're referring to here?  It looks the visibility of routes when a selection is made hasn't changed in this PR, and that is good.  Aside from stylistic changes to how the route highlights are done, I was suggesting that when the user mouses over the map (triggering the tooltip), we also highlight the route's row in the table.

> Regarding the timeout, removing it won't have a huge impact but it will be noticeable when moving through the routes consecutively at a somewhat fast rate.

For this particular feature, I think that is fine.  Let's try it out.

trxaphion:
My mistake, after I read the paragraph again it seems I misunderstood.

Removed timeout on heroku demo.

hathix:
Hey Raymond,

We were all going crazy about your feature at the meeting just now. In the words of one of our members, it's "tight."

Couple pieces of feedback:

Screen Shot 2020-02-12 at 7 10 12 PM

Terence was saying that the highlighting is very subtle now, making it hard to pick out the route. (Before it was a bit too much?) Perhaps something in between would be better.

Screen Shot 2020-02-12 at 7 10 28 PM

In the "live" version of OpenTransit, each line is a different width based on how often the bus comes. I don't see that in this version of yours. Can you put back that feature?

hathix:
> I was suggesting that when the user mouses over the map (triggering the tooltip), we also highlight the route's row in the table.

Also a good suggestion from Terence. How hard would that be to add?

hathix:
Only change that's a "must" is putting the variable line widths back. Other things can be put in separate issues so we can address them later.

trxaphion:
I did already undo the changes to the scale, it just seems that the route stats didn't update beyond a point as you can see there are no scores on the route table. I didn't look into it as I had assumed it was from a conflict on the backend. If you go back to Feb 9 or earlier you will see there are stats and the lines reflect them.

As for highlighting the row it sounds possible.

exxonvaldez:
The code looks pretty good, but I can't test it effectively.  I think there's some problems with the latest GTFS feed for Muni, from looking at some of the newest PRs, so we have duplicate routes.

When I click a point on the map, I still see most or all routes in the table, where I'm expecting to see just the routes that are visible on the map.  I don't see a code change in this PR that would cause that, or maybe I missed it.

It might be helpful to address the git conflicts, as this should simplify the diffs a bit.  A lot of changes are due to the linter/prettier fixing things, which are now also fixed on master.

frontend/src/components/MapSpider.jsx
```diff
@@ -152,85 +157,163 @@ class MapSpider extends Component {
...
+      polylines.push(this.generateShield(startMarker, waitScaled));
 
-    return <Fragment>{items}</Fragment>;
+      return polylines;
+    });
+    return items;
+  };
+
+  /**
+   * Get additional scale when near max zoom to account for increased periphery
```
exxonvaldez: I'm not sure I understand this comment, what is "increased periphery"?  Also, I tried having this function always return 0 vs its current behavior at various zoom levels below and above 17.  I couldn't actually see a difference, what is the issue this code solves?  This method seems kind of hard to understand and maintain if we want to adjust how the map renders later.

frontend/src/actions/index.js
```diff
@@ -389,6 +389,18 @@ export function handleSpiderMapClick(stops, latLng) {
...
 
+export function handleRouteTableHover(hoverRoute) {
+  return function(dispatch) {
+    dispatch({ type: 'RECEIVED_ROUTE_TABLE_HOVER', hoverRoute });
+  };
+}
+
+export function handleSegmentHover(segmentRouteId) {
+  return function(dispatch) {
+    dispatch({ type: 'RECEIVED_SEGMENT_HOVER', segmentRouteId });
```
exxonvaldez: The name "segment" threw me off for a bit, until I realized this was just the id of the route being hovered in the map.  Can we change it to something like "spider hover", "map hover", or "spider map hover"?

frontend/src/components/MapSpider.jsx
```diff
@@ -492,7 +609,12 @@ class MapSpider extends Component {
             opacity={0.3}
           />
           {/* see http://maps.stamen.com for details */}
-          <this.DownstreamLines />
+          <this.HoveredLine />
+          <this.MemoizedDownstreamLines
+            latLng={spiderSelection.latLng}
+            stops={spiderSelection.stops}
+            statsByRouteId={statsByRouteId}
```
exxonvaldez: From what I understand of the React.memo docs, the idea is that the memoized component skips re-rendering if a shallow compare of the props matches the props used for rendering?  So the props you're passing in are the ones that, if different, force rerendering of downstream lines?  Can you add some comments to that effect?  It's not intuitive at first, since DownstreamLines doesn't consume all these props directly.

frontend/src/components/RouteTable.jsx
```diff
@@ -252,7 +256,15 @@ function RouteTable(props) {
...
-  const { statsByRouteId } = props;
+  const {
+    statsByRouteId,
+    onRowHover,
+    spiderSelection: { segmentRouteId },
+  } = props;
+
+  useEffect(() => {
+    return () => onRowHover(null);
+  }, [onRowHover]);
```
exxonvaldez: What is the purpose of this useEffect?  Looks like it only runs on first render?

frontend/src/components/RouteTable.jsx
```diff
@@ -414,14 +426,21 @@ function RouteTable(props) {
             rowCount={displayedRouteStats.length}
             columns={columns}
           />
-          <TableBody>
+          <TableBody onMouseLeave={() => onRowHover(null)}>
             {stableSort(displayedRouteStats, order, orderBy).map(row => {
```
exxonvaldez: This is sort of unrelated to your PR. We have this duplicate route id thing going on right now on the back end, and I noticed that it generates a bunch of console warnings due to duplicate TableRow keys.  Also this breaks the spider map's filtering of the table rows.  I noticed that filtering works again if we use unique keys -- so instead of keying on route id, we could just key on row index, like "routerow" + index.
```suggestion
            {stableSort(displayedRouteStats, order, orderBy).map((row, index) => {
```

frontend/src/actions/index.js
hathix: Yeah, I think the term "segment" is going to take on a more specialized meaning now (as "the route between two adjacent stops"). Let's not use that term for other things.

frontend/src/components/MapSpider.jsx
trxaphion: It's really subtle. May just be me being nitpicky after staring at it for so long. Basically I find that when I'm zoomed out I'm looking near the marker so it's easy to spot when the route gets outlined but when I'm zoomed in to a hot spot with many routes on screen it's harder to notice, particularly when the route is less traveled and near the edge of the map.

frontend/src/components/RouteTable.jsx
trxaphion: I believe when I navigated to a different page while still having the route hovered (via back or forward on mouse) and then returning to the page the route is still hovered.

frontend/src/components/MapSpider.jsx
exxonvaldez: Hmm, okay.  I guess some change is needed here, either add the above text as a comment so others will better understand why this function is needed, or remove the function to keep things more maintainable and easier to understand.

frontend/src/components/RouteTable.jsx
exxonvaldez: I see, was able to reproduce that.  Can you add some comments to that effect?  useEffect is a bit of swiss army knife, meaning it's hard to know what the effect is being used for.  My interpretation is that this effect runs on mount and unmount to reset the table hover state.

frontend/src/actions/index.js
trxaphion: Given Neel's description segment seems increasingly appropriate to me since "the route between two adjacent stops" is precisely what is being hovered. If anything shouldn't just the parameter be changed to routeId since the routeId is not specific to each segment? Although the term doesn't seem to be used much so I guess at a glance it may not be understood.

trxaphion: Or we could throw all of that into a blender and have handleSpiderSegmentHover. I guess it really just depends how specific you want it.

exxonvaldez: Agree that segmentRouteId should just be routeId.  Going by memory, I don’t remember the term “segment” being used in the front end code, so “handleSegmentHover” doesn’t mean much to me.  handleSpiderSegmentHover is a lot better and is acceptable to me, but I’d prefer handleSpiderHover more.

frontend/src/components/MapSpider.jsx
trxaphion: zoomScale

frontend/src/components/RouteTable.jsx
trxaphion: So are we just going to chuck it in here so we wont need a separate PR?

exxonvaldez: Well, the duplicate problem is fixed now, so I guess it’s not needed.  It was mainly just to get the spider filtering working again to make testing this PR easier.

frontend/src/components/MapSpider.jsx
exxonvaldez: Is this without the additional scale function?  I’m fine with this, I’d rather not have additional logic for zoom levels greater than 17.

PR #551, opened 2020-02-13
Max colors

### LINKED ISSUE #422: Display routes in the recommended colors from the GTFS feed
Some routes are associated with colors, e.g. the MAX lines in Portland. However, we currently don't display these routes on the map using the same color, which is confusing.

GTFS feeds define colors for certain routes in the route_color and route_text_color columns in routes.txt. If present, we should save these colors in the route configuration in S3.

After colors are saved in the route configuration in S3, the frontend should be updated use those colors if available when rendering routes on the map.

licarpen:
## Proposed changes

Adds a conditional to check for existing color in startMarker.routeInfo and sets color to corresponding color.  If null, uses color generated by index.  

## Screenshot

Screen Shot 2020-02-12 at 8 42 48 PM

...

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -13875,9 +13875,9 @@
       "integrity": "sha1-fjL3W0E4EpHQRhHxvxQQmsAGUdc="
     },
     "qs": {
-      "version": "6.9.1",
-      "resolved": "https://registry.npmjs.org/qs/-/qs-6.9.1.tgz",
-      "integrity": "sha512-Cxm7/SS/y/Z3MHWSxXb8lIFqgqBowP5JMlTUFyJN88y0SGQhVmZnqFK/PeuMX9LzUyWsqqhNxIyg0jlzq946yA=="
+      "version": "6.5.2",
+      "resolved": "https://registry.npmjs.org/qs/-/qs-6.5.2.tgz",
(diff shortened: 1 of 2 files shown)


exxonvaldez:
Looks interesting, did you mean to close the pull request?

licarpen:
Hi Terence,

Yes - I closed the pull request because I realized I failed the linting
tests.  I am resolving the issues and will reopen the pull request once the
issue is resolved.  Thanks for your patience!

youngj:
Thanks for working on this! 

Since the code for determining routeColor is repeated in a few places, could you refactor that logic into a method?

Is the change to package-lock.json intended? If it's not necessary, probably better to revert that file.

licarpen:
I reverted the changes to the package-lock file as they were unintentional.  I considered refactoring given that the routeColor variable was declared 4 times, but held off as it seems the entire file could benefit from being refactored to eliminate repetitive code and doing so would require several major changes that go beyond the scope of the issue I worked on.  I was considering opening a new issue for this refactor. Is that sometime I should do or should I go ahead and just refactor for the routeColor?

youngj:
Thanks! I'd refactor just the routeColor calculation in this PR. If you're interested in refactoring the file to eliminate other repetitive code, feel free to work on that in another PR.

licarpen:
getRouteColor method has been added and routeColor variable renamed to routeColorOptions.

Issue #553, opened 2020-02-13
Let users customize stats shown on dashboard


hathix:
We're currently limited to showing just a few stats on the dashboard:

<img width="1552" alt="Screen Shot 2020-02-12 at 9 12 19 PM" src="https://user-images.githubusercontent.com/728084/74403548-720fbf80-4ddc-11ea-9379-a095da162ce4.png">

But you can imagine that different user groups may prefer different stats. Transit planners, for instance, love headways but don't really care about wait time. So it would be good to offer a "settings" panel for the user (store settings in LocalStorage or something) that lets them customize the stats they see on the dashboard. Such a settings panel would be useful for additional customization for power users going forward.

PR #556, opened 2020-02-14
refactor route table so columns are defined in one place

youngj:
Adding/removing a column in the route table required updating code in 3 different places: the column header definitions, the TableCell definitions for normal rows, and the help tooltip content. This made it difficult to modify the columns in the table. Also, there was some repetitive code for setting cell styles for each column. This PR refactors the RouteTable class so that each column is only defined in one place in the code. This should make it easier to make the columns configurable in the future (#553).


youngj:
Could someone review this?

Issue #554, opened 2020-02-13
Precompute and display median service frequency (headway)


youngj:
It would be helpful to be able to display median service frequency (headway) as one of the top-level statistics in the route table and route summary. This is a simple metric that most riders probably understand, and is used by transit agencies (see e.g. https://www.sfmta.com/routes/1-california). 

In order to display the median headway, we would need to precompute the median headway for each stop in compute_stats.py and store the median of the median headway for each direction, similar to how we are storing median wait times.

This issue should also include storing the scheduled median headways If #544 is implemented first. Otherwise, #544 should include storing the scheduled median headways. 

(Note that SF Muni refers to the service frequency as the "wait time between vehicles". Since we are currently using "wait time" to mean something different, it is likely that some users will incorrectly interpret our wait time column as service frequency. This confusion would be avoided if we displayed service frequency instead of, or in addition to, the median wait time for riders arriving randomly at a stop.)

PR #560, opened 2020-02-15
Precompute median service frequency (headway) for each direction, and add to GraphQL API

youngj:
This change partially implements #554, adding a medianHeadway field on the DirectionIntervalMetrics GraphQL type to make it possible to show median service frequency for each route on the dashboard and route screen. This PR doesn't update the UI to actually show median service frequency yet.

To avoid increasing the size of the "combined" precomputed stats file, it doesn't store the median headways for each stop, but only the median of the median headways for all stops in each direction.

Similarly, the "combined" precomputed stats file now only stores the median of the median wait time and on-time rate for each stop. This reduces the file size from about 700 kB to about 650 kB. Most of the file size is due to the stats used for the travel time chart, average speed, and travel time variability.

Only the isochrone map needs precomputed median wait time stats for each stop, so these were moved into the median-trip-times file that the isochrone map is already using.

### DIFF
--- a/backend/compute_stats.py
+++ b/backend/compute_stats.py
@@ -1,13 +1,16 @@
-from models import util, arrival_history, trip_times, constants, config, timetables, precomputed_stats, wait_times
+from models import util, arrival_history, trip_times, constants, config, timetables, precomputed_stats, wait_times, metrics
 import argparse
 from datetime import date
 import collections
 import numpy as np
 import time
 
+combined_stat_id = precomputed_stats.StatIds.Combined
+median_trip_times_stat_id = precomputed_stats.StatIds.MedianTripTimes
+
 all_stat_ids = [
-    'combined',
-    'median-trip-times',
+    combined_stat_id,
+    median_trip_times_stat_id,
 ]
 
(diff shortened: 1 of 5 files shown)


youngj:
Thanks for the review! 

We initially needed all the wait times only for the isochrone map. We never actually used all the on-time rates for each stop. When viewing metrics for a particular trip, the GraphQL API can compute metrics from the raw arrivals data pretty fast. The GraphQL API also needs to provide more detailed metrics about the distribution of wait times and schedule adherence than what is stored in the precomputed stats file.

DefaultVersion should be changed if the data format changes in a backwards-incompatible way (i.e. if old code would not understand the new data format). Since the old code relies on the medianWaitTimes property having a dict of stop IDs, and the new data format no longer provides this, it would be necessary to change either the version code or the stat ID. Adding new properties to a file would not necessarily require changing the version code or the stat ID because the old code would just ignore the new properties. However, it might still be useful to change the version code even if the format is backwards-compatible, so that it is clear what files were created with each version of the code. For example, in arrival_history.py the version code (v4c) ends with a letter that is changed each time we make significant changes to the algorithm for computing arrival times, even though the file format is the same.

PR #562, opened 2020-02-17
Precompute scheduled metrics and add to GraphQL API

### LINKED ISSUE #544: Precompute scheduled wait time and trip time metrics
On the route summary screen, it would be useful for users to be able to compare the median wait time, average speed, and travel times based on actual historical data with the expected values based on the timetable.

In order to enable this, the backend would need to precompute median wait times and trip times based on the timetable in a similar way to how it precomputes statistics based on the actual historical data. The backend could then provide GraphQL fields with the scheduled metrics.

Since the scheduled metrics will typically be the same on multiple days, it would only be necessary to precompute metrics for each of the "date keys" for that agency's timetable.

youngj:
Fixes #544.

This PR updates precomputes scheduled median wait times, trip times, and headways based on the timetable, using the same functions in compute_stats.py that compute statistics based on the observed historical data. It also updates the GraphQL API to return the scheduled metrics, which would make it possible to compare scheduled and observed metrics on the route screen. (This PR doesn't implement the UI to actually display the scheduled metrics.)

Since the scheduled metrics are typically the same on multiple days, metrics are only precomputed for each of the "date keys" for that agency's timetable.

compute_stats.py now takes the `--scheduled` command line flag to compute scheduled metrics based on the timetable. If no date is provided at the command line, it will compute metrics for all of the dates in the current GTFS feed.

save_routes.py (which is run daily via a cron job) also has a `--scheduled-stats` command line flag which will automatically precompute statistics for new dates that are added to the GTFS feed.

The precomputed metrics are available in GraphQL in the DirectionIntervalMetrics fields scheduledMedianWaitTime, scheduledMedianHeadway, scheduledAverageSpeed, and scheduledCompletedTrips; and the SegmentIntervalMetrics fields scheduledMedianTripTime and scheduledTrips.

Most of the code in this PR is refactoring models/schema.py and models/metrics.py to implement scheduled metrics without duplicating code. Also, several GraphQL types now have corresponding classes with the same name in the metrics package which are used as the `parent` object in the GraphQL definition (instead of dicts). This change reduces the amount of repetitive parameter lists and makes the GraphQL API easier to understand.


backend/timetable.py
```diff
@@ -19,7 +19,6 @@
     parser.add_argument('--verbose', dest='verbose', action='store_true', help='verbose output')
     parser.set_defaults(verbose=False)
     parser.set_defaults(comparison = False)
-    parser.set_defaults(thresholds = '5,10')
```
EddyIonescu: Was this ever used anywhere?

frontend/public/isochrone-worker.js
```diff
@@ -132,9 +132,9 @@ async function getTripTimes(agencyId, dateStr, timeStr)
     {
         let timePath = getTimePath(timeStr);
 
-        let s3Url = 'https://'+S3Bucket+'.s3.amazonaws.com/precomputed-stats/'+PrecomputedStatsVersion+'/'+agencyId+'/'+
+        let s3Url = 'https://'+S3Bucket+'.s3.amazonaws.com/observed-stats/'+PrecomputedStatsVersion+'/'+agencyId+'/'+
             dateStr.replace(/\-/g, '/')+
-            '/precomputed-stats_'+PrecomputedStatsVersion+'_'+agencyId+'_median-trip-times_'+dateStr+timePath+'.json.gz';
+            '/observed-stats_'+PrecomputedStatsVersion+'_'+agencyId+'_median-trip-times_'+dateStr+timePath+'.json.gz';
```
EddyIonescu: Can't wait to see the scheduled transit vs. actual transit isochrones!

backend/models/metrics.py
```diff
@@ -548,40 +562,109 @@ def get_average_speed(self, route_id, direction_id, rng: Range, units):
...
+                    start_time_str = start_time_str, end_time_str = end_time_str,
+                    scheduled=scheduled
+                )
+            except FileNotFoundError as e:
+                self.precomputed_stats[key] = None
+
+        return self.precomputed_stats[key]
+
+    def get_route_ids(self):
+        return self.get_route_configs().keys()
```
EddyIonescu: This file's getting pretty big, what are your thoughts on splitting the classes into a metrics module/folder? (not now, just in a future PR)

backend/timetable.py
youngj: A previous version of this file had a --thresholds argument, but it no longer exists. It was similar to the current arguments --early-min and --late-min.

backend/models/gtfs.py
```diff
@@ -276,7 +276,7 @@ def save_timetables(self, save_to_s3=False, skip_existing=False):
 
         if skip_existing and date_keys == old_date_keys:
             print("No new dates in GTFS feed, skipping")
-            return
+            return False
```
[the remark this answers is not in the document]
youngj: Sounds good, done.

backend/save_routes.py
```diff
@@ -53,4 +56,8 @@
         scraper.save_routes(save_to_s3)
 
         if args.timetables:
-            scraper.save_timetables(save_to_s3=save_to_s3, skip_existing=True)
+            updated_timetables = scraper.save_timetables(save_to_s3=save_to_s3, skip_existing=True)
```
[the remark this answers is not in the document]
youngj: That seems like a better variable name, done.

backend/models/metrics.py
youngj: Maybe, although in this case I think having several small files might be more annoying to maintain than one big file. Although compute_headway_minutes is an outlier in metrics.py so probably that should be moved somewhere else.

Issue #127, opened 2019-06-11
Latest GTFS route configs are not in sync with historic data


exxonvaldez:
As Jose mentioned in https://github.com/trynmaps/metrics-mvp/pull/89 , we need to archive GTFS data because it updates regularly.

For example, between May 24 and June 11, the F outbound direction id changed from F____O_E00 to F____O_F00, with a new terminal ("Castro" instead of "17th and Noe") and new last stop id (33311).  So attempts to show anything about F outbound trips will not find any data .  Even if we add logic to fall back on older direction ids, we still have the problem of the stop list being out of sync also.

exxonvaldez:
There seems to be a GTFS archive here that has each new version:  https://transitfeeds.com/p/sfmta/60

exxonvaldez:
With #207 we can process shapes for older GTFS data just by pointing the script at the archive mentioned above.  However, that leaves the question of whether we can get historic Nextbus route configs, or if we should derive the equivalent of them from GTFS data.

PR #576, opened 2020-02-22
Exclude routes in GTFS but not Nextbus when sorting in save_routes

### LINKED ISSUE #575: save_routes fails for routes in GTFS but not in Nextbus
Muni added some routes (see https://www.sfmta.com/blog/service-changes-coming-february-22) that exist in GTFS but not in Nextbus as they start on February 22 (the GTFS was downloaded on February 21, which had the new routes added so that Google/others will have them loaded in time).

This is because service changes have to be manually made in Nextbus, so there's usually a discrepancy between GTFS and Nextbus until the next business day; this is a common issue for the TTC as schedule changes start on Sundays but Nextbus sometimes changes on the Monday after.

EddyIonescu:
Fixes #575 

## Proposed changes

- If there are routes that don't have sort_order when the first one does, it means that they are in the GTFS but not in Nextbus (at the time of this PR). This could be when a new route is added, and so it appears in the GTFS ahead of time so that it'll be on Google Maps (as that takes a few days to update GTFS) but not on Nextbus until the day it starts service. This PR puts the new routes at the end (sort order of 9999), unless they're a subway line (special case for TTC), in which case they go to the front (sort order of 0).
- Previously the sort_order from Nextbus was used if the first route had it, which was not the case for TTC. This PR uses the sort_order from Nextbus if any route has it, which is determined before sorting.

Expected Result / Testing:
Now when running save_routes.py for Muni it no longer crashes and prints the following:
```
Excluded Route, in GTFS but not in Nextbus: KTBUS
Excluded Route, in GTFS but not in Nextbus: JBUS
Excluded Route, in GTFS but not in Nextbus: NBUS
Excluded Route, in GTFS but not in Nextbus: MBUS
Excluded Route, in GTFS but not in Nextbus: LBUS
```
As well, TTC routes are now sorted properly and subway lines appear at the top.

### DIFF
--- a/backend/models/gtfs.py
+++ b/backend/models/gtfs.py
@@ -8,6 +8,7 @@ import boto3
 import gzip
 import hashlib
 import zipfile
+from collections import OrderedDict
 
 from . import config, util, nextbus, routeconfig, timetables
 
@@ -995,11 +996,34 @@ class GtfsScraper:
                 if route_data['sort_order'] is None:
                     if route_data['type'] == 1:
                         # place subways at the front if they don't have a sort_order
-                        return -1
+                        return 0
                     return 9999
                 return route_data['sort_order']
             return route_data['id']
         return sorted(routes_data, key=get_sort_key)
+
+    def newer_route(self, newest_route_data, route_data):
+        try:
+            return int(
+                route_data['gtfs_route_id']
+            ) > int(newest_route_data['gtfs_route_id'])
+        except ValueError:
+            return route_data['gtfs_route_id'] > newest_route_data['gtfs_route_id']
+
+    def remove_duplicate_routes(self, routes_data):
+        """Remove routes with the same id that appear multiple times.
+        Keeps the route with the highest gtfs_route_id, converting it to
+        a number of possible."""
+        routes_dict = OrderedDict()
+        for route_data in routes_data:
+            if route_data['id'] not in routes_dict or self.newer_route(
+                routes_dict[route_data['id']],
+                route_data,
+            ):
+                routes_dict[route_data['id']] = route_data
+        return [
+            routes_dict[route_id] for route_id in routes_dict
+        ]
 
     def save_routes(self, save_to_s3=True):
         agency = self.agency
@@ -1007,6 +1031,7 @@ class GtfsScraper:
 
         routes_df = self.get_gtfs_routes()
         routes_data = self.sort_routes(routes_df)
+        routes_data = self.remove_duplicate_routes(routes_data)
         
         routes = [routeconfig.RouteConfig(agency_id, route_data) for route_data in routes_data]
 
--- a/backend/models/gtfs.py
+++ b/backend/models/gtfs.py
@@ -967,56 +967,47 @@ class GtfsScraper:
 
         return route_data
 
-    def save_routes(self, save_to_s3=True):
+    def sort_routes(self, routes_df):
         agency = self.agency
-        agency_id = agency.id
-
         routes_data = []
-
-        routes_df = self.get_gtfs_routes()
-
         if agency.provider == 'nextbus':
-            nextbus_route_order = [route.id for route in nextbus.get_route_list(agency.nextbus_id)]
-
+            nextbus_route_order = [
+                route.id for route in nextbus.get_route_list(agency.nextbus_id)
+            ]
+        use_sort_order = False
         for route in routes_df.itertuples():
             route_data = self.get_route_data(route)
-
             if agency.provider == 'nextbus':
                 try:
                     sort_order = nextbus_route_order.index(route_data['id'])
+                    use_sort_order = True
                 except ValueError as ex:
                     print(ex)
                     sort_order = None
             else:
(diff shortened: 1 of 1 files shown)


EddyIonescu:
Thanks for noticing that - no way Muni has 160 routes. I'll fix that later today. I also changed the use of sort_order to be when any route has it vs. just the first one, as the first TTC route isn't on Nextbus as it's a subway line.

backend/models/gtfs.py
```diff
@@ -967,40 +968,71 @@ def get_route_data(self, route):
...
-            sort_key = lambda route_data: route_data['id']
-
-        routes_data = sorted(routes_data, key=sort_key)
+    def save_routes(self, save_to_s3=True):
+        agency = self.agency
+        agency_id = agency.id
 
+        routes_df = self.get_gtfs_routes()
+        routes_data = self.sort_routes(routes_df)
+        routes_data = self.remove_duplicate_routes(routes_data)
```
youngj: Currently remove_duplicate_routes is called at the end, after the work has already been done to extract duplicate route objects from the GTFS feed. I think get_route_data is responsible for most of the running time of save_routes.py, so if duplicates could be removed before calling get_route_data on each route, save_routes.py would probably be almost 2x faster. Not a big deal since we don't expect to see GTFS feeds with duplicate routes very often.

backend/models/gtfs.py
```diff
@@ -967,40 +968,71 @@ def get_route_data(self, route):
...
+            return int(
+                route_data['gtfs_route_id']
+            ) > int(newest_route_data['gtfs_route_id'])
+        except ValueError:
+            return route_data['gtfs_route_id'] > newest_route_data['gtfs_route_id']
+
+    def remove_duplicate_routes(self, routes_data):
+        """Remove routes with the same id that appear multiple times.
+        Keeps the route with the highest gtfs_route_id, converting it to
+        a number of possible."""
```
youngj: Is something missing after "converting it to a number of possible"?

backend/models/gtfs.py
```diff
@@ -967,40 +968,71 @@ def get_route_data(self, route):
...
+        try:
+            return int(
+                route_data['gtfs_route_id']
+            ) > int(newest_route_data['gtfs_route_id'])
+        except ValueError:
+            return route_data['gtfs_route_id'] > newest_route_data['gtfs_route_id']
+
+    def remove_duplicate_routes(self, routes_data):
+        """Remove routes with the same id that appear multiple times.
+        Keeps the route with the highest gtfs_route_id, converting it to
```
youngj: It seems like using the highest GTFS route ID doesn't always work for the new Muni GTFS feed. 

For example, in routes.txt:
```
CALIFORNIA A EXPRESS,3,,,SFMTA,15223,https://SFMTA.com/1AX,,1AX
CALIFORNIA A EXPRESS,3,,,SFMTA,15444,https://SFMTA.com/1AX,,1AX
```

trips.txt:
```
b_5141,15223,1,the Financial District,175675,1_merged_9191498,9054637
...
a_5139,15444,1,the Financial District,177423,1_merged_9191495,9189828
```

calendar.txt:
```
1_merged_9191498,20200222,20200529,1,1,1,1,1,0,0
...
1_merged_9191495,20200203,20200221,1,1,1,1,1,0,0
```

Basically, 15223 has a trip with service ID 1_merged_9191498 which is valid between 2/22/2020 and 5/29/2020,
and 15444 has a trip with service ID 1_merged_9191495 which is valid between 2/3/2020 and 2/21/2020. In this case the newer version of the route has a smaller GTFS route ID.

Potentially we might not even always want to save the newer route. For example, if the route shape/stops are scheduled to change in the future, we wouldn't want the current metrics to be computed based on the shapes/stops that aren't active yet.

Eventually we should store different route configurations for different time ranges (#127), but that hasn't been implemented yet. 

For now, how about we find a trip for each route ID, get the associated service ID, and check to see the current date is between the start/end date for that service ID?

EddyIonescu: Thanks for noticing this, forgot about the issue where the new changes are in the GTFS but aren't yet active. Weirdly Muni didn't change any routes except for moving the last trip on the 1AX/BX earlier.
This would also solve the issue of where transit agencies add or remove a route and put that in the GTFS ahead of the service changes. I'll make a separate issue and PR for that later today.

EddyIonescu: Just wondering, where are GTFS and the nextbus_Configs stored in S3? Or are they stored on server?

I also noticed that while TTC updates their GTFS file a few days before the changes come into effect, the new GTFS doesn't contain the old routes (more specifically, the old trips and service_id's) as it's assumed that consumers will keep using their existing version of the GTFS, except then we'd have no routes as they'd be excluded (the old GTFS would have to be used, which save_routes just overwrote). This isn't an issue now (I'll make a new issue for it) but should be resolved by mid-March. In that case, we'd need to download the GTFS, but use the existing one if the new one doesn't have any calendar ranges covering the given date (or throw an error otherwise?) before overwriting it.

PR #578, opened 2020-02-23
Fix file write concurrency

### LINKED ISSUE #565: Avoid concurrency errors when saving cache files
Some backend Python modules have functions that write cache files in the backend/data directory. However, in production it's possible that multiple Python processes could try to write the same cache file at the same time, which could cause invalid data to be stored in the file. This issue could be prevented if the cache files were written atomically, such as by writing a temporary file and then renaming it to the desired file name.

These methods include arrival_history.get_by_date, precomputed_stats.get_precomputed_stats, routeconfig.get_route_list, timetables.get_data_by_date_key, and timetables.get_date_keys . 

Perhaps the similar logic in these functions could also be refactored to reduce duplicated code.

JulianPinzaru:
Fixes #565.

## Proposed changes
Using native python module `fcntl` to put locks on writing data to a file. This way we can prevent multiple processes writing to the same file simultaneously and erasing each other's data.


backend/models/util.py
```diff
@@ -1,8 +1,34 @@
...
+    assert mode == 'w' or mode == 'a'
+
+    with open(filepath, mode, **kwargs) as f:
+        fcntl.flock(f, fcntl.LOCK_EX)
+        written_characters = f.write(data)
+        fcntl.flock(f, fcntl.LOCK_UN)
+    return written_characters
+
+
+def append_to_file(filepath=None, data=None, mode='a', **kwargs):
```
bmblumenfeld: Since it is strictly append -- might be good to take out kwarg and just pass in mode='a' util.to write_to_file

backend/models/util.py
```diff
@@ -1,8 +1,34 @@
...
+    Concurrency safe function for writing data to the file.
+    Param mode: file open mode ('a' for appending, 'w' for writing)
+    '''
+    assert mode == 'w' or mode == 'a'
+
+    with open(filepath, mode, **kwargs) as f:
+        fcntl.flock(f, fcntl.LOCK_EX)
+        written_characters = f.write(data)
+        fcntl.flock(f, fcntl.LOCK_UN)
+    return written_characters
```
bmblumenfeld: This question comes from my lack of knowledge:
If another process tries to access file while it is locked does it lead to an IOError? Or does it automatically wait for lock to be released?

backend/models/util.py
```diff
@@ -1,8 +1,34 @@
...
+    '''
+    Concurrency safe function for writing data to the file.
+    Param mode: file open mode ('a' for appending, 'w' for writing)
+    '''
+    assert mode == 'w' or mode == 'a'
+
+    with open(filepath, mode, **kwargs) as f:
+        fcntl.flock(f, fcntl.LOCK_EX)
+        written_characters = f.write(data)
+        fcntl.flock(f, fcntl.LOCK_UN)
```
bmblumenfeld: Maybe worth explicitly closing file

bmblumenfeld:
@JulianPinzaru awesome work! I left a couple comments. Should be good to go after addressing those! Let me know if you have any questions!

youngj:
I did some tests with this approach. Although it ensures that one process finishes writing before the other process starts, the output file could still contain data from both processes if the first process writes more data than the second process.

Here's an example script to show how this can happen:

foo.py:
```
import argparse
import time
import fcntl

def write_to_file(filepath=None, data=None, mode='w', **kwargs) -> int:
    with open(filepath, mode, **kwargs) as f:
        fcntl.flock(f, fcntl.LOCK_EX)
        for i in range(1, 1000):
            f.write(data)
        time.sleep(3)
        for i in range(1, 1000):
            f.write(data)
        fcntl.flock(f, fcntl.LOCK_UN)

if __name__ == '__main__':

    parser = argparse.ArgumentParser()
    parser.add_argument('--foo', required=True, help='foo')
    args = parser.parse_args()

    write_to_file('foo.txt', args.foo + '\n')
```

Test two processes writing at once; the first process that opens the file writes more data than the second:
`(python foo.py --foo=aaaaaaaaaaaaaaaaaaaaaa &); sleep 1; python foo.py --foo=b`

The output file contains a bunch of b's followed by a bunch of `\0` characters, followed by a bunch of a's.

Instead of using `flock`, I think the best approach is to write a temporary file with a unique name in the same directory, then rename it to the desired filename when it is complete. With this approach, `flock` isn't necessary because each writer has a unique filename to write to, and the filesystem will ensure that the rename is atomic.

This approach also ensures that read_from_file will always see a consistent version of the file without needing to acquire a lock even if it is being written to concurrently by another process.

Issue #423, opened 2019-11-20
Avoid running out of memory when fetching vehicle location data for large agencies


youngj:
When running compute_arrivals.py or get_state.py, vehicle location data is loaded from tryn-api in chunks (e.g. 15 minutes at a time) in order to prevent the tryn-api server from running out of memory. 

However, the local Python process keeps all of the loaded chunks in memory at once until all chunks are fetched, before splitting the data by route and writing the day's worth of data to disk for each route. For transit agencies with a lot of vehicles, such as TTC (agency id: ttc), it is possible that the Python process may be killed due to running out of memory before writing the vehicle location data to disk.

One way to avoid this would be to write each chunk to the backend/data directory after it is fetched, instead of keeping them all in memory. 

This could also be used as a cache to avoid fetching identical chunks from tryn-api again if running compute_arrivals.py or get_state.py multiple times.

However it would still be necessary to load data from all chunks in order to save 1 day's worth of data for a particular route, which could also lead to running out of memory. To avoid this, chunks could be split by route before saving on disk.

The relevant code is in the get_state function in backend/models/trynapi.py .

EddyIonescu:
https://github.com/trynmaps/metrics-mvp/pull/580

PR #580, opened 2020-02-24
Store trynapi data by route on disk instead of in-memory

EddyIonescu:
Resolves Issue #423, as to not longer store all vehicle
states for a chunk in memory, as this can cause out-of-memory
issues for agencies with a lot of vehicles (such as TTC).

## Proposed changes

This approach involves writing the chunks for each route for each date into a CSV file, which allows for appending vehicle rows. It avoids writing many small JSON files, which would take longer to load and assemble, and also uses less space.

Performance-wise the `get_state` function of compute_arrivals now takes about 10x longer for individual routes, excluding time spend on calls to `get_raw_state`. It shouldn't be an issue to deploy as it is, as 10x is, from about 0.5 to about 5 seconds for TTC route 32 (has 25-50 vehicles all day), and about 0.3 to about 2 seconds for tiny routes like TTC routes 30 (1 vehicle all day) and 176 (1 vehicle peak periods); vast majority of the time is still spent getting the data from tryn-api and processing the arrival data. Merging this PR as it is will add somewhere around 2-4 minutes to a daily Muni run.

When removing the conversion of the CSV temp to JSON, the time takes only slightly more than before (0.5 to 0.6 seconds for TTC route 32), so performance would be similar if we switch to CSV caches. This doesn't account for time saved from memory allocations, which take longer when running the old version for all the routes at once.

Space-wise, json.gz takes up slightly less space (about 10% less, samples of two TTC routes) than csv.gz, which was pretty surprising, so it'd increase AWS data-transfer costs.

## Testing

JSON cache files created by running compute_arrivals.py in master and in this branch are identical.

### DIFF
--- a/backend/models/trynapi.py
+++ b/backend/models/trynapi.py
@@ -7,6 +7,7 @@ from . import config
 import time
 from pathlib import Path
 from datetime import date
+import pandas as pd
 
 class CachedState:
     def __init__(self):
@@ -45,8 +46,6 @@ def get_state(agency_id: str, d: date, start_time, end_time, route_ids) -> Cache
         print('state already cached')
         return state
 
-    route_state_map = {}
-
     # Request data from trynapi in smaller chunks to avoid internal server errors.
     # The more routes we have, the smaller our chunk size needs to be in order to
     # avoid getting internal server errors from trynapi.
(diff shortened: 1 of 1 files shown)


EddyIonescu:
I've updated the PR message with some performance comparisons. Good point that using CSV's would simplify things in eclipses and improve performance there - I'll make an issue for it. I'll let you merge this as to ensure it works properly on the Kubernetes/GCloud environment.

hathix:
@youngj you good to merge this?

PR #583, opened 2020-02-27
Remove routes without any service_ids that cover today's date

### LINKED ISSUE #581: Duplicate Muni routes
Muni’s current GTFS file contains each route twice, but with a different route ID. This is because, looking at the service_id, one is for before February 22, and one is for once it occurs. Note that February 22 was the rollout date of their most recent service changes, which were moving the 1AX/BX last trip time earlier and adding early-morning Muni Metro bus routes, in addition to the existing OWL service.

### LINKED ISSUE #584: Route sorting on homepage doesn't work
You can use the buttons to try sorting the routes table, but nothing really gets reordered:

image

Also it seems that there are 2 entries for each route...

EddyIonescu:
Fixes #581. Fixes #584.

## Proposed changes

Muni's GTFS contains each route twice, once from before February 22, and again starting from February 22, when they modified service on a few routes. This PR modifies save_routes to remove routes that don't have any service_ids that cover the date in which the script is being run, which is set to the current date (when versioned GTFS is added the date will be made an argument).

## Testing

Removes the duplicate routes, keeping the ones valid after February 22. Running Muni is still blocked by #576, which would need to be merged beforehand. 

- For TriMet, after removing the calendar.txt file and setting the date to a Sunday, all routes including the 92 South Beaverton Express (weekday peak-periods only) appear.
- For Muni, setting the date to February 22 (first date of changes) results in only the new routes being shown.

### DIFF
--- a/backend/models/gtfs.py
+++ b/backend/models/gtfs.py
@@ -191,8 +191,9 @@ class GtfsScraper:
 
     def get_services_by_date(self, ignore_day_of_week=False):
         """Returns map of a date object to a list of service_ids.
-        Can optionally supply ignore_day_of_week, which specifies that service_ids
-        for any day of week to be included as long as the date is within their range."""
+        Can optionally supply ignore_day_of_week, which includes all service_ids
+        that are active on a date within a date range provided in calendar.txt
+        regardless of the day of the week."""
         calendar_df = self.feed.calendar
         calendar_dates_df = self.feed.calendar_dates
 
@@ -1052,9 +1053,6 @@ class GtfsScraper:
     def save_routes(self, save_to_s3, d):
         agency = self.agency
         agency_id = agency.id
-
-        routes_data = []
-
         routes_df = self.get_gtfs_routes()
         routes_df = self.get_active_routes(routes_df, d)
         routes_data = [
--- a/backend/models/gtfs.py
+++ b/backend/models/gtfs.py
@@ -978,13 +978,24 @@ class GtfsScraper:
 
         return route_data
 
-    def get_active_routes(self, routes_df, date):
+    def get_active_routes(self, routes_df, d):
         """Returns routes in routes_df whose service_ids all have a
         start_date that is at or before and an end_date that is at or
         after the given date."""
         trips_df = self.get_gtfs_trips()
         dates_map = self.get_services_by_date(ignore_day_of_week=True)
-        active_services = set(dates_map[date])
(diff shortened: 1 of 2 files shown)


backend/save_routes.py
```diff
@@ -50,10 +51,11 @@
     agencies = [config.get_agency(args.agency)] if args.agency is not None else config.agencies
 
     save_to_s3 = args.s3
+    date = date.today()
```
youngj: This variable shadows the class named `date`. To avoid this we have been using the variable name `d` for datetime.date instances.

backend/models/gtfs.py
```diff
@@ -967,13 +976,44 @@ def get_route_data(self, route):
...
         return route_data
 
-    def save_routes(self, save_to_s3=True):
+    def get_active_routes(self, routes_df, date):
+        """Returns routes in routes_df whose service_ids all have a
+        start_date that is at or before and an end_date that is at or
+        after the given date."""
+        trips_df = self.get_gtfs_trips()
+        dates_map = self.get_services_by_date(ignore_day_of_week=True)
+        active_services = set(dates_map[date])
```
youngj: Trimet's GTFS file doesn't have a calendar.txt to specify date ranges, only a calendar_dates.txt listing all dates when the service is active. It seems like if save_routes.py was run on a weekend or holiday where there were no service IDs associated with a particular route, this would say the route is not active. Perhaps instead it could check if today is between the first and last date for a service ID?

EddyIonescu: Thanks for noticing this, didn't know that calendar.txt was optional. Also TriMet's GTFS does have calendar.txt now, I guess they added it recently (although their calendar_dates.txt still includes every date). Your approach sounds good - I've implemented it and tested it after deleting TriMet's calendar.txt file.

EddyIonescu:
@youngj Hey, thanks for catching that, sorry for making you do another review cycle. It looks like I forgot to rename `calendar.txt` even though I said I did, so that's why the 92 showed up on a Sunday. I've tested it again for March 8, 2020 with today's GTFS and with `calendar.txt` renamed and the 42 and 78 didn't appear, while the 92 and the 1 did. The 78a did appear, although looking in `calendar_dates.txt` it's supposed to.
I've also added a comment and renamed `_d` to something harder to miss.
Also let me know if you can run the job from February 22 - if not I can run it locally for Muni for the dates between then and now.

PR #461, opened 2019-11-26
UI for date range comparisons

exxonvaldez:
Hi, I wanted to get some feedback on what date range comparisons would look like in the UI and how they're implemented with respect to our GraphQL API.

Fixes #78.  Partially addresses #321.

## Proposed changes

Front end only for #321 .  Fully copies Uber's date range comparison UI.  To do this, there is now a second date picker in the app bar that starts out empty/unused.  Once populated it can be removed.

The date picker now has a label showing the selected days of the week.  The date picker uses local state as the user makes changes, only applying them to the Redux state once the picker is closed.  This reduces the number of graphQL APIs calls.

I'm thinking that date range comparisons only make sense where date ranges are currently supported, which is the stop to stop stats.

Next steps:  To finish implementing comparisons for stop to stop stats, we can make two graphQL API calls, and have the UI compare summary stats.  The summary stats would be shown as percent differences where it makes sense, like median wait time 15% longer, median speed 5% slower... or side by side numbers if we like that better.  We can also draw the time of day/headway/wait time/trip time bar charts side by side in different colors.

It looks like uber movement makes only one API call, and what comes back is differences along with data for side by side bar charts.  I'm not sure if we want to optimize our API to support this also, or not.

If we implement date range support for the other screens (isochrone, route table, route summary), there is the same question of what do comparisons look like and how are they implemented.  Via two data fetches, and the UI calculating differences, or one API call for comparisons -- I'm picturing the precomputed wait/trip times json but every stat where applicable is a ratio between the value for date range A and the value date range B?

## Screenshot

Screen Shot 2019-11-26 at 12 30 08 AM

Screen Shot 2019-11-26 at 12 29 54 AM

## Link to demo, if any

https://ot-terence-date-range-compare.herokuapp.com/

PR #589, opened 2020-02-28
Feedback on #461

exxonvaldez:
Quick feedback items from #461.

## Proposed changes

- index.js: Add documentation about Redux thunks.
- Info.jsx: Add documentation about why there currently is no "by day" data for date range comparisons.
- DataDiagnostic.jsx: Fix typo that broke page and remove unneeded text.

## Screenshot

n/a

## Link to demo, if any

n/a

### DIFF
--- a/frontend/src/components/Info.jsx
+++ b/frontend/src/components/Info.jsx
@@ -44,7 +44,14 @@ function Info(props) {
     tripMetrics && tripMetrics.interval2
       ? tripMetrics.interval2.tripTimes
       : null;
-  // By day data is not requested for the second date range.
+  /*
+   * By day data is not requested for the second date range.
+   *
+   * The second range can have the same data the first, as they are using the same GraphQL query API.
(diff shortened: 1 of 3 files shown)


frontend/src/components/Info.jsx
```diff
@@ -44,7 +44,14 @@ function Info(props) {
...
       : null;
-  // By day data is not requested for the second date range.
+  /*
+   * By day data is not requested for the second date range.
+   *
+   * The second range can have the same data the first, as they are using the same GraphQL query API.
+   * It gets tricky for the "by day" tab, because how do you chart two date ranges by day when they
+   * could have different numbers of days in them?  Does this chart only work when the ranges have
+   * the same number of days?  Do we "scale" the time axis so both are the full width of the chart?
+   */
```
hathix: You're right that this is tricky. My first guess is to use the same axis for both, but make the lines different length (so the shorter interval ends earlier). We can iterate on this later though, once we get the big items done.

PR #593, opened 2020-02-28
DateTimePopover fixes from #461.

### LINKED ISSUE #550: Add "apply"/"done" button to date picker
Currently it's not obvious how you "apply" any changes you made to the date range dropdown. The changes actually happen immediately, but it's a bit jarring because you're not sure exactly how to exit the dropdown. In my head, I was thinking "if I hit X or click outside the dropdown, will it reset my changes?" 

Screen Shot 2020-02-12 at 8 37 33 PM

Now, obviously my gut reaction wasn't accurate, but it still shows that UX is confusing. We could use an "Apply" or "Done" button that literally just closes the popup but gives the user peace of mind.

My comment in #461 shows some random bugs that could be fixed by this.

We can copy Uber Movement for this. I think Uber movement doesn't have a reset button, just a Done button.

exxonvaldez:
Fixes #550 

## Proposed changes

-DateTmePopover: Add an "Apply" button that does what "x" used to do, as in Uber Movement.  Make "x" into a cancel button that reverts the local form state.

- routesMap: Fix bug noted in #550 where second date range cannot be the default (yesterday) by adding a flag on the query string that signifies that the second range is active.

## Screenshot

Screen Shot 2020-02-27 at 7 00 50 PM

## Link to demo, if any

https://ot-terence-date-range-compare.herokuapp.com/

### DIFF
--- a/frontend/src/components/DateTimePopover.jsx
+++ b/frontend/src/components/DateTimePopover.jsx
@@ -71,12 +71,16 @@ function DateTimePopover(props) {
     graphParams[targetRange] || initialGraphParams.firstDateRange,
   );
 
-  // Whenever targetRange changes, we need to resync our local state with Redux
-
-  useEffect(() => {
+  function resetLocalDateRangeParams() {
     setLocalDateRangeParams(
       graphParams[targetRange] || initialGraphParams.firstDateRange,
     );
+  }
+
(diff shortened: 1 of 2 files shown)


hathix:
Much better ux, clever use of the grid too. And thanks for fixing that bug on #550. Can't approve on my phone but I will when I'm at computer

PR #596, opened 2020-03-01
Add custom directions and route heuristics for TriMet

youngj:
Handles loop routes, routes with multiple branches, directions that only occur at certain times of day, hides some routes from the dashboard, and excludes certain stops at the beginning and end of routes to make end-to-end statistics more representative of typical performance.

### DIFF
--- a/backend/agencies/trimet.yaml
+++ b/backend/agencies/trimet.yaml
@@ -6,3 +6,142 @@ js_properties:
   title: TriMet
   initialMapCenter: { lat: 45.525, lng: -122.677 }
   initialMapZoom: 12
+  routeHeuristics: {
+    '12': {
+      '1': {
+        ignoreLastStop: true,
+      },
+    },
+    '15': {
+      '1': {
+        ignoreFirstStop: true,
+        ignoreLastStop: true,
+      },
+    },
+    '20': {
+      '0': {
+        ignoreFirstStop: true,
+      },
+    },
+    '20a': { # apparent duplicate of route 20 in GTFS feed
+      ignoreRoute: true,
+    },
+    '22': {
+      '1': {
+        ignoreLastStop: true,
+      },
+    },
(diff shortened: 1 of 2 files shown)


backend/agencies/trimet.yaml
```diff
@@ -2,7 +2,225 @@ id: trimet
 gtfs_url: https://developer.trimet.org/schedule/gtfs.zip
 gtfs_agency_id: TRIMET
 timezone_id: America/Los_Angeles
+custom_directions:
+  '15':
+    - id: "0"
```
EddyIonescu: Running save_routes with the recentmost GTFS from TriMet (downloaded today), I got the following error:
```Exception: 2 shapes found for route 15 with GTFS direction ID 1 including 6470: 439404,437165```.

youngj: Thanks, fixed. We should probably update the custom direction logic to use the most common matching shape rather than requiring only one match.

youngj:
We don't have location data for WES Commuter Rail, but we do have location data for the night buses. It might not really be necessary to hide the night buses on the Dashboard. (The Muni config hides OWL routes as well.)

backend/agencies/trimet.yaml
EddyIonescu: Definitely agree, this'll be needed for adding TTC as for most routes each branch has two shapes: the regular one that enters the subway station (vast majority of trips) and another one where a few early-morning or late-night trips serve the station on-street as the station's closed, so it has a different shape and stop-pattern even though it's the same route/branch. It'd make sense to use the most common shape for each branch in that case.

Also, I think the Muni config hid the Owl routes because of the date splitting issue, which is no longer the case so that could be changed in the future. Also those routes can definitely have reliability issues because runtimes tend to be a lot tighter at night, so they're worth processing.

PR #597, opened 2020-03-01
Fix regression and warning from compare date range PR.

exxonvaldez:
## Proposed changes

- index.js: Fixed references to date/time parameter that should have been updated in the compare date range PR.  This fixes time of day options that were not being sent in the route metrics query.
- DateTimePopover: Fix react-hooks/exhaustive-deps compiler warning, which was

```
metrics-react-dev | ./src/components/DateTimePopover.jsx
metrics-react-dev |   Line 84:  React Hook useEffect has a missing dependency: 'resetLocalDateRangeParams'. Either include it or remove the dependency array  react-hooks/exhaustive-deps
```
which when fixed results in

```
metrics-react-dev | ./src/components/DateTimePopover.jsx
metrics-react-dev |   Line 74:  The 'resetLocalDateRangeParams' function makes the dependencies of useEffect Hook (at line 89) change on every render. To fix this, wrap the 'resetLocalDateRangeParams' definition into its own useCallback() Hook  react-hooks/exhaustive-deps
```
which is now fixed in this PR.

## Screenshot

n/a

## Link to demo, if any

n/a

### DIFF
--- a/frontend/src/actions/index.js
+++ b/frontend/src/actions/index.js
@@ -395,8 +395,8 @@ export function fetchAgencyMetrics(params) {
     const variablesJson = JSON.stringify({
       agencyId: Agencies[0].id,
       dates,
-      startTime: params.startTime,
-      endTime: params.endTime,
+      startTime: params.firstDateRange.startTime,
+      endTime: params.firstDateRange.endTime,
     });
 
     if (getState().agencyMetrics.variablesJson !== variablesJson) {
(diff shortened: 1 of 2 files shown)


hathix:
Can confirm that the bug was present before (see https://muni.opentransit.city/?firstDateRange%5BstartTime%5D=03%3A00&firstDateRange%5BendTime%5D=07%3A00 and notice that its stats are the same as https://muni.opentransit.city/). This should do it.

PR #603, opened 2020-03-05
Improve SEO by rendering route name as H1

### LINKED ISSUE #391: Set a <h1> tag for the route after selection
We should have an `<h1>` tag at the top of the information frame, but underneath the navigation component for better user experience and SEO

hathix:
Fixes #391 .

## Proposed changes

For SEO reasons, it helps to make the most important heading in your page an H1. For us, that would be the route name (e.g. "38-Geary") on a route page. 

This PR makes the route name render as an H1, but actually keep the same styling as all the other breadcrumbs.

## Screenshot

Observe how the route name is an H1 but everything else is rendered as the default tag, H6; all the breadcrumb tags have the same styling regardless.

H6:
Screen Shot 2020-03-04 at 8 45 29 PM

H1:
Screen Shot 2020-03-04 at 8 45 19 PM

### DIFF
--- a/frontend/src/screens/RouteScreen.jsx
+++ b/frontend/src/screens/RouteScreen.jsx
@@ -91,10 +91,17 @@ function RouteScreen(props) {
         const updatedPayload = Object.assign({ ...link.payload }, payload);
         link = Object.assign({ ...link }, { payload: updatedPayload });
         const { label, specialLabel } = labels(param, path.title);
+
+        // For SEO purposes, make the very first breadcrumb (which is the
+        // route name) render as an H1, but be styled like everything else
+        // (as a regular, small subtitle).
+        // if it's not the first one, render as default HTML tag.
+        const renderAsTag = index === 0 ? 'h1' : null;
+
         return hasNextValue ? (
           <Typography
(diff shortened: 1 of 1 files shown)


hathix:
can someone review this?

frontend/src/screens/RouteScreen.jsx
```diff
@@ -91,9 +91,17 @@ function RouteScreen(props) {
         const updatedPayload = Object.assign({ ...link.payload }, payload);
         link = Object.assign({ ...link }, { payload: updatedPayload });
         const { label, specialLabel } = labels(param, path.title);
+
+        // For SEO purposes, make the very first breadcrumb (which is the
+        // route name) render as an H1, but be styled like everything else
+        // (as a regular, small subtitle).
+        // if it's not the first one, render as default HTML tag.
+        const renderAsTag = index === 0 ? 'h1' : null;
```
akgupta89: I would prefer this named as renderAsH1 so the variable name tells exactly what it does

hathix: Aha, good catch. Will address in a commit.

PR #607, opened 2020-03-10
Improve user interface design; add new metrics and charts; refactor frontend components

### LINKED ISSUE #246: isochrone crashes while computing if user switches to Dashboard
Steps to reproduce:

1.  Go to the Isochrone
2.  Click on the map
3.  While isochrones are being computed, use the hamburger menu to go to the Dashboard

Actual result:

See screenshot below.

Screen Shot 2019-08-18 at 4 49 22 PM

Expected result:

Dashboard should appear.

Note:  I took a quick look at the code, did not see a quick fix.  Not sure if we should be trying to stop the worker when navigating away, or just ignoring its updates.

### LINKED ISSUE #394: Map fixed when scrolling down the routes list

### LINKED ISSUE #438: Add button to clear to/from stops

### LINKED ISSUE #542: Show histogram of minutes early/late
The On-Time % stat is useful to learn how often the departure times were close to the schedule, but it doesn't show how early or late the vehicles were. 

The GraphQL API already provides the `closestDeltas` field (https://github.com/trynmaps/metrics-mvp/blob/master/docs/metrics_api.md#scheduleadherence), which provides metrics about the number of minutes early/late relative to the scheduled time. The front end could display a histogram of this metric, similar to the existing histograms of wait times and trip times.

### LINKED ISSUE #543: Compare actual to scheduled trip metrics
Currently it is difficult for users to tell how well routes are performing relative to the intended schedule.

The GraphQL API provides information about the scheduled wait times, headways, trip times, and number of arrivals (https://github.com/trynmaps/metrics-mvp/blob/master/docs/metrics_api.md#tripintervalmetrics), but this information is currently not displayed in the web app. 

It would be useful to see scheduled values next to the actual values on the cards on the Summary tab, as well as the Headways, Wait Times, and Trip Times histograms, and the By Day / By Time of Day charts.

### LINKED ISSUE #545: Histogram tooltips show statistics for previous bucket when cursor over left half of bucket
The tooltips for the histograms of headways, wait times, and travel times are updated by the onNearestX property, but the nearest x coordinate is different depending on whether the cursor is over the left or right half of the bucket.

### LINKED ISSUE #554: Precompute and display median service frequency (headway)

### LINKED ISSUE #582: Date range comparison UI improvements
Jesse [said](https://github.com/trynmaps/metrics-mvp/pull/461#issuecomment-586020958) on #461:

> 
> I don't think cards are an effective way of comparing two sets of metrics. Visually it appears haphazard because each statistic is aligned somewhat differently and the two ranges have different font sizes. Also, the cards don't make it very clear which statistic refers to which date range (you have to refer to the ranges shown in the visually disconnected top right corner of the screen). Clarifying this within the card UI would require additional redundant text in each card.
> 
> The same issues would apply when using cards to compare actual/scheduled metrics within a single date range.
> 
> Copied from #549:
> 
> It seems like a more straightforward approach would be to use a table where each row is a statistic, with columns for the first date range, the second date range, and percentage increase/decrease 
> 
> For example:
> 
> |  | 1/1/2020  | 2/1/2020 | % Change | |
> | --- | --- | --- | --- | --- |
> | Median Service Frequency |  5 min |  8 min | +60% | `__________0******____` | 
> | Median Trip Time |  10 min |  8 min | -20% | `________**0__________` |
> 
> (The last column shows an optional horizontal bar to visually represent the % change.)
> 
> Another approach would be could be a bar graph with series like Median Service Frequency, Median Trip Time, etc. where each series has two side by side bars for each date range.
> 
> Both of these UI approaches would also work for comparing Actual vs Scheduled statistics for a single date range.
> 
> The date ranges should also be labeled in the various charts.
> 
> In the histograms, the bars from the top histogram often block the view of the bottom histogram. The bars should probably be displayed side by side instead of overlaid.

### LINKED ISSUE #590: Let users see all routes on homepage's spider map
Currently, there's no way to see all routes on the homepage's spider map. You can only click in certain areas and see the nearby routes.

Screen Shot 2020-02-27 at 6 48 00 AM

We should either (1) change the behavior such that the map shows all routes when no point on the map is clicked, or (2) add a "show all routes" button to the bottom of the map near the "clear" button.

I prefer (1) because:

* It makes the homepage more engaging by letting new users see all the routes as soon as they open the app. Currently it looks bland, since all you see on the left is a gray map.
* It makes the spider map / table behavior more consistent. When you click on a spot on the map, the table shows just the routes that are visible on the spider map. But when you haven't clicked, the table shows all routes while the map shows no routes. That's inconsistent.
* It avoids adding yet another button to the map.

However, (2) avoids putting a lot of information in front of new users.

A compromise could be to only show the most common routes on the map by default, but that would make the map-table behavior even more inconsistent.

### LINKED ISSUE #620: With date ranges, page too wide on mobile
Screenshot_20200329-195709

On my phone, the header bar is so wide that the page has to scroll horizontally - it can't all fit on the phone screen at once. that's not good ui.

youngj:
This pull request updates several parts of the user interface to fix various usability issues. It also adds new metrics and charts that were already returned by the GraphQL API but not yet displayed, including comparing observed and scheduled metrics. It also refactors the code for many of the frontend components, making it easier to add charts and reorganize the user interface in the future.

The design changes are described in https://docs.google.com/document/d/1RpVZsbBHFiPLXbKcmsa4gi68ab-C_yi1Ar3lHshu6i4/edit# . This document includes detailed descriptions of usability issues that were identified with the current design, as well as the proposed user interface changes. Please review this document and feel free to add comments.

With input from several people at Code for PDX, I also drafted a more general document proposing design principles for OpenTransit that all contributors could refer to when considering design changes in the future: https://docs.google.com/document/d/19soN0I69lygHuQbKbWlx7OKgHOek38MU2eMS0y3rH1s/edit# . Since OpenTransit has a growing number of contributors in multiple cities, agreeing on shared design principles would help ensure that teams in different cities can continue to benefit from each other’s work, without needing to fork the codebase and work on changes on their own.

The TriMet version of the app with these changes can be seen at http://opentransit-youngj.herokuapp.com/ . 

In the process of implementing these user interface design changes, a lot of the frontend code was refactored as well.

* Moved the main navigation bar into the App component, to avoid duplicating navigation UI components in every screen.
* Created a SimpleChart component, a wrapper around react-vis XYPlot, to make it easier to create charts with crosshairs (tooltips showing the values at the mouse cursor) and appropriately formatted axis labels, ticks, and legends, with a consistent UI with much less code. The react-vis library provides low-level functionality which previously required a lot of tedious duplicative code for each chart, such as managing the state to keep track of which data point the mouse is currently over. Most of the existing charts were refactored to use SimpleLineMarkChart, SimpleVerticalBarChart, or SimpleVerticalRectChart, and several new charts were implemented: OnTimePerformanceByDayChart, OnTimePerformanceByTimeChart, OnTimePerformanceHistogram, ServiceFrequencyByDayChart, ServiceFrequencyByTimeChart, and ServiceFrequencyGapHistogram.
* Each chart is refactored into its own component, which is connected to the redux state. This allows the screens that display charts to render each chart like <ServiceFrequencyByTimeChart /> without needing to pass props to each chart, and simplifies the code by avoiding mixing logic for several charts in one screen. This will make it easier to rearrange charts and screens in the future, and make it easier to customize which charts are displayed via configuration. If a chart is not applicable for the current graphParams, it will render as empty. This lets the parent screen not need to know the conditions when each chart is displayed.
* Updated most charts to support comparing observed vs scheduled metrics, and comparing metrics from two date ranges.
* Previously, the x-axis labels for "By Day" charts were shown for each date, which made it necessary to render each date vertically to avoid overlap. Now, the x-value is set to the index of the date, which allows react-vis to avoid showing every date label on the x-axis.
* Renamed components with names starting with "Info" to more appropriate names.
* Used GraphQL `@include` directives to avoid creating separate queries for single/dual date ranges, and avoid fetching timeRanges field when the time range is not "All Day"
* Added updateQuery action to simplify logic in date/time range controls. To change the query string while keeping the rest of the URL the same, it calls `dispatch` with the `type` set to state.location.type, and `payload` set to state.location.payload. (It isn't necessary to look up the `type` from state.page, since the `type` is already part of the state.)
* The RouteSummary component now only shows data from state.routeMetrics, instead of combining data from state.agencyMetrics and state.routeMetrics. In the future, when the user changes the date range, we could avoid making the API request to update state.agencyMetrics unless the user is viewing the Dashboard.
* Split the content of each RouteScreen tab into separate components (e.g. OnTimePerformanceStats, ServiceFrequencyStats)
* Tabs in the RouteScreen component are refactored so that each tab and its associated component are defined only in one place. This would make it easier to customize the RouteScreen tabs in the future via configuration.
* When a RouteScreen tab is not active, its content is no longer rendered. Previously, inactive tabs were rendered but the associated content was set as hidden.

We have already been discussing these changes over the past few weeks in Portland. I'd expect that there may also be consensus support in SF for many of these changes, but some contributors in SF may not agree with some of these proposed changes, such as the removal of scores, stars, colored chips, cards, and certain metrics. However, we would support changes that allow SF to customize the UI via configuration, e.g. if SF still wants to display scores. 

Since these changes involve significant refactoring that will result in many merge conflicts as other developers make frontend changes, I would recommend merging these changes relatively quickly and creating issues in GitHub for non-trivial fixes. If the SF team still wants to show scores, stars, cards, etc, I would suggest making this functionality available via configuration options after the changes are merged.

Fixes #590. Fixes #582. Fixes #554. Fixes #545. Fixes #543. Fixes #542. Fixes #438. Fixes #394. Fixes #246. Fixed #620.


hathix:
A doozy! Great bundle of usability improvements, and I'm glad we have a more fleshed-out design framework. Comments:

* I don't worry too much about removing the scores as long as we address #553 at some point, which would let us bring back the score if desired. 
* However, the color coding is a useful way to let people skim and extract key information at a glance. I would support a PR that lets teams add that back if desired. But, as you said, merging this quickly is important, and I think your PR is generally 80+% of the way there, so the best move would be to merge this and file a bunch of issues afterward.
* Refactors look fine to me, though Terence/Arun/others are better qualified to speak about it.
* re: moving isochrone/about from hamburger menu to header: not bad, but I worry that it may not scale well as we add more tabs. Perhaps just getting rid of the icon labels is enough.
* I assume the homepage would be fleshed out, instead of being a few blue links and text. Busstat.nyc had a handy homepage that we could use.
* These touch handles (the plus / X) are pretty small; I would suggest having bigger ones

image

* Also, on that view, the fields move a lot when I select something on the dropdown. A bit disorienting. I see how it is compact, but I would prefer if the controls were spread out across the page at the start so they didn't constantly shuffle around whenever I picked something. The Grid View may be of use?
* On the [Marey chart view](http://opentransit-youngj.herokuapp.com/route/85), the left and right side of the screen don't scroll together. Should we change that?
* When I choose 2 date ranges, the comparison table goes from observed vs scheduled to observed vs observed. Makes sense, but was jarring at first. An icon to call out more explicitly that the table is showing different kind of data from before would help.
* This might be a backend thing, but the sorting here is busted:

image

* The new design guide talks a lot about how the table on the metrics homepage is explicitly about the *observed* data without any kind of value judgment, but I think users would like to see the comparison to the *schedule* at a glance too. A switch on that page that lets users see comparisons to the schedule (like, turn all the numbers into +/- percentages showing the observed vs schedule data) would be useful. Obviously a big feature, and would belong in a new issue.
* It took me a second to figure out what the green numbers ("26 min shorter", etc) meant on this view. Perhaps add something that notes that these numbers compare Right vs Left, or add some visual styling that more closely associates the green numbers with the 2nd date range data. For instance, the "37 min" could say "37 min (26 shorter)".

image

* Contextual hints on error messages would be useful. For instance, here, it would be nice to have a blue link in the error text that, when clicked, removes the second date.

image

hathix:
Oops, accidentally closed

hathix:
There are a bunch of other little things, but overall, I like the new direction and the effort that the PDX team has put in. I'm not qualified to review all the actual code, but functionality and design wise, this is a strong start. We may need to file a bunch of issues once this is merged, but that's fine.

I would ask some of the frontend engineers to give the code a proper review before approving.

frontend/src/App.css
```diff
@@ -64,6 +71,77 @@
     opacity:1.0;
 }
 
+.app-tab-text {
+    display:none;
+}
+
+@media (min-width: 1050px)
```
hathix: Does Material come with preset breakpoints that you can use? Or can you use constants to create new consistent breakpoints for this entire app? Wouldn't want this to eventually evolve into something that has a bunch of different, arbitrary breakpoints.

frontend/src/App.css
```diff
@@ -64,6 +71,77 @@
...
+}
+
+.app-logo {
+    max-height: 38px;
+    padding-top: 5px;
+    padding-left: 10px;
+    display:inline;
+}
+
+@media (min-width: 700px)
```
hathix: Again, using preset breakpoints would be useful, if possible.

frontend/src/App.jsx
```diff
@@ -4,25 +4,42 @@ import { createMuiTheme } from '@material-ui/core/styles';
...
   Dashboard,
   RouteScreen,
   DataDiagnostic,
   NotFound,
 };
 
 const theme = createMuiTheme({
+  breakpoints: {
+    values: {
+      md: 1050,
```
hathix: Ah, I see -- can constants like this be filled into the CSS file?

frontend/src/App.jsx
```diff
@@ -32,15 +49,85 @@ const theme = createMuiTheme({
...
   },
 });
-const App = ({ page }) => {
-  const Component = Components[page];
+
+const App = props => {
+  const { page, dispatch, type } = props;
+  const Screen = Screens[page];
+
+  const agency = Agencies[0];
```
hathix: Are there flags that developers and/or users can set to switch between agencies? How would they do that once this PR is merged, anyway?

frontend/src/App.jsx
```diff
@@ -32,15 +49,85 @@ const theme = createMuiTheme({
...
+          <Toolbar variant="dense" disableGutters>
+            <AppBarLogo />
+            <div className="page-title">{agency.title}</div>
+            <LoadingIndicator />
+            <div className="flex-spacing"></div>
+            <Tabs value={tabValue} onChange={handleTabChange}>
+              <Tab
+                label={
+                  <div>
+                    <PollIcon style={{ verticalAlign: '-6px' }} />
```
hathix: can the -6px be set in CSS instead of in `style`?

frontend/src/App.jsx
```diff
@@ -32,15 +49,85 @@ const theme = createMuiTheme({
...
+                    <InfoRoundedIcon style={{ verticalAlign: '-6px' }} />
+                    <span className="app-tab-text"> About</span>
+                  </div>
+                }
+                value="ABOUT"
+              />
+            </Tabs>
+          </Toolbar>
+        </AppBar>
+        <div style={{ height: 48 }}>&nbsp;</div>
```
hathix: Can this also be set in CSS?

frontend/src/components/ControlPanel.jsx
```diff
@@ -152,121 +155,182 @@ function ControlPanel(props) {
...
+            ),
+            InputLabelProps: {
+              htmlFor: 'route',
+              shrink: true,
+            },
+          }}
+          options={(routes || []).map(route => ({
+            value: route.id,
+            label: route.title,
+          }))}
```
hathix: Clever solution

frontend/src/components/ControlPanel.jsx
```diff
@@ -152,121 +155,182 @@ function ControlPanel(props) {
...
+      {selectedDirection ? (
+        <>
+          <FormControl className={classes.formControl}>
+            <ReactSelect
+              onChange={onSelectFirstStop}
+              inputId="fromstop"
+              textFieldProps={{
+                label: (
+                  <span style={labelStyle}>
+                    Origin Stop{' '}
```
hathix: What's the `{' '}` for? Is that just to insert an `&nbsp;`?

frontend/src/screens/Home.jsx
```diff
@@ -0,0 +1,69 @@
...
+            }}
+          >
+            Metrics
+          </Navlink>
+        </h3>
+        <div>
+          Statistics and visualizations of the historical performance of each
+          route.
+        </div>
+      </div>
```
hathix: Is this section a placeholder? In a future PR, it'd be great to see this changed to something more eye-catching -- some graphics or icons, interesting charts, and/or a sleeker design.

youngj: Yes, the Home screen is just a placeholder.

frontend/src/App.jsx
youngj: Developers can already switch between agencies by setting the OPENTRANSIT_AGENCY_IDS environment variable, as documented in README.md. Users can't switch between agencies. This PR doesn't change how that works.

frontend/src/components/ControlPanel.jsx
youngj: It just adds a normal space, since JSX removes whitespace next to tags.

youngj:
> re: moving isochrone/about from hamburger menu to header: not bad, but I worry that it may not scale well as we add more tabs. Perhaps just getting rid of the icon labels is enough.

Agreed -- this potential issue is noted in the Google doc: "Replaced hamburger menu with 3 tabs on right side of toolbar, so users can easily discover them and switch in one click. Text labels are hidden on small screen widths. With this design, we could probably add 1 more tab in the toolbar if needed. If we need to add more tabs in the future, could switch to a drawer on the left side of the screen, although for now using this unused space in the top-right of the toolbar allows more horizontal space for content."

> I assume the homepage would be fleshed out, instead of being a few blue links and text. Busstat.nyc had a handy homepage that we could use.

Yeah, the home page content is just a placeholder. This could be something to work on next.

> These touch handles (the plus / X) are pretty small; I would suggest having bigger ones

I agree. The handles are the same size as the inline field labels. Could increase the text size a bit.

> Also, on that view, the fields move a lot when I select something on the dropdown. A bit disorienting. I see how it is compact, but I would prefer if the controls were spread out across the page at the start so they didn't constantly shuffle around whenever I picked something. The Grid View may be of use?

This inline layout allows the route, direction, origin stop, destination stop, and date-time range controls to fit on one row of a typical laptop screen for typical route/direction/stop names. We don't know how long each of the text values of each field will be (route/direction/stop names are sometimes quite long), so if the components are not allowed to move then it would likely require two rows of controls, which reduces the screen space for statistics without scrolling. For me, it doesn't seem disorienting for the subsequent fields to move when you change the selection of a previous field. A lot of the page content changes at the same time.

> On the Marey chart view, the left and right side of the screen don't scroll together. Should we change that?

They do both scroll together. Are you referring to the map not having a fixed position? In order for the map on the route screen to have a fixed position, we would probably need to make everything above it including the tabs and dropdown controls have a fixed position too.

> When I choose 2 date ranges, the comparison table goes from observed vs scheduled to observed vs observed. Makes sense, but was jarring at first. An icon to call out more explicitly that the table is showing different kind of data from before would help.

Could you give an example of what you mean by an icon? I think the heading text and chart legends are pretty clear about what data is being shown.

> This might be a backend thing, but the sorting here is busted:

Not sure how to reproduce this, it works for me.

> The new design guide talks a lot about how the table on the metrics homepage is explicitly about the observed data without any kind of value judgment, but I think users would like to see the comparison to the schedule at a glance too. A switch on that page that lets users see comparisons to the schedule (like, turn all the numbers into +/- percentages showing the observed vs schedule data) would be useful. Obviously a big feature, and would belong in a new issue.

Agreed, it would be useful to compare observed/scheduled metrics in the routes table somehow. From the Google doc: "In the future, we may want to allow users to compare both observed and scheduled metrics in the same table. There would probably be too many columns if we displayed all observed and scheduled metrics at once, but could have a dropdown that allows you to select one metric at a time (e.g. Median Service Frequency) and shows Observed/Scheduled columns just for that metric. The same approach could work for comparing a metric for multiple date ranges."

> It took me a second to figure out what the green numbers ("26 min shorter", etc) meant on this view. Perhaps add something that notes that these numbers compare Right vs Left, or add some visual styling that more closely associates the green numbers with the 2nd date range data. For instance, the "37 min" could say "37 min (26 shorter)".

I'm open to different ways of presenting this data. Maybe saying "26 min shorter" in text isn't really necessary. 

> Contextual hints on error messages would be useful. For instance, here, it would be nice to have a blue link in the error text that, when clicked, removes the second date.

Sounds good to me.

> Nit, but it's unclear whether this arrow belongs to the Route or Median Service Frequency columns.

The arrows are on the left side of right-aligned columns (not changed in this PR). I think users will understand because the arrow changes when they click on the column header.

> The About page could use some links to our GitHub, the project page, the Code for America pages, etc.

Sounds good to me, this content is just a placeholder.

> Seems that this new version doesn't change the <title> at all, whereas the old version did.

Changing the page title sounds like a good idea. It doesn't look like the old version changed the page title either, though.

frontend/src/App.jsx
youngj: I don't know if constants can be passed to App.css. Perhaps the responsive style rules with media queries could be refactored to use https://material-ui.com/components/use-media-query/ to avoid duplicating constants in the CSS file. Probably not necessary to do it in this PR.

frontend/src/App.jsx
youngj: Done.

frontend/src/App.jsx
youngj: Done

youngj:
Thanks for the feedback! I increased the size of the +/x buttons in the control panel and also renamed the Trip Chart tab (formerly Marey Chart) to Time-Distance Chart, as suggested by Eddy in the Google doc.

http://opentransit-youngj.herokuapp.com/route/100?firstDateRange%5BstartDate%5D=2020-03-25&firstDateRange%5Bdate%5D=2020-03-25

It looks like there is an unrelated issue with computing statistics starting on March 26, probably related to https://muni.opentransit.city/ being offline.

frontend/src/components/MareyChart.jsx
```diff
@@ -253,7 +253,7 @@ function MareyChart(props) {
 
   if (!mareyChartSupported) {
     return (
-      <div>The Marey chart is only available when viewing a single date.</div>
+      <div>The trip chart is only available when viewing a single date.</div>
```
EddyIonescu, line 256:
```suggestion
      <div>The time-distance chart is only available when viewing a single date.</div>
```

frontend/src/components/ControlPanel.jsx
hathix: Oh, yeah, I saw that React / lint / something automated does that.

frontend/src/components/MareyChart.jsx
youngj: Fixed, thanks!

PR #610, opened 2020-03-26
Geolocation button

### LINKED ISSUE #186: Geolocation button in isochrone map
It would be useful to automatically draw the isochrones starting at the user's current location, if known

licarpen:
## Proposed changes

Add setInitialLocation method to Isochrone that gets the user's geolocation if available and sets to default lat/long if location is not available.  Isochrones are then generated in either case on page load.  Converts fetchRoutes action into promise using async await in order to setInitialLocation after routes have loaded. 
...
## Screenshot

On page load when user's location is successful:
Screen Shot 2020-03-25 at 7 54 28 PM

On page load when location is not successful:
Screen Shot 2020-03-25 at 7 55 05 PM


hathix:
Does this automatically request the user's location as soon as they open the page? We chose not to do that on the metrics dashboard to make it not too intrusive.

licarpen:
It does - see issue #186 for the conversation regarding this decision.

youngj:
Using the user's location on the dashboard is a bit of a different situation than on the isochrone map. On the dashboard, all routes are shown in the table when no location is provided, whereas only nearby routes are shown when a location is provided. In that case, it would be confusing to not show all of the routes by default.

It makes sense to me to use the user's location for the isochrone on page load so that it shows information without users needing to click something. Although they would still need to accept or reject the prompt to share their location the first time.

frontend/src/screens/Isochrone.jsx
```diff
@@ -150,11 +151,13 @@ class Isochrone extends React.Component {
 
   componentDidMount() {
     if (!this.props.routes) {
-      this.props.fetchRoutes();
-    }
+      this.props.fetchRoutes().then(() => this.setInitialLocation());
```
youngj: Previously we haven't been using actions as promises, because the component props will be updated anyway when the redux state changes. I don't think `.then(() => this.setInitialLocation())` is actually necessary here because componentDidUpdate is called after fetchRoutes completes. It would probably be simpler to remove the `.then` here and the async/await in fetchRoutes.

frontend/src/screens/Isochrone.jsx
```diff
@@ -150,11 +151,13 @@ class Isochrone extends React.Component {
...
+    } else this.setInitialLocation();
```
youngj: I'm not sure if we have a style rule for this, but my preference would be to always use curly brackets for if/else blocks with a single statement and put the statement on a separate line. The advantage of always using curly braces is that it would not require adding curly braces later if more than 1 statement was needed in the if/else block, and could potentially avoid bugs if someone added multiple statements without curly braces.

frontend/src/screens/Isochrone.jsx
```diff
@@ -150,11 +151,13 @@ class Isochrone extends React.Component {
...
   }
 
   componentDidUpdate(prevProps) {
+    // computeIsochrones requires loaded routes
+    if (prevProps.routes !== this.props.routes) this.setInitialLocation();
```
youngj: Same as above

frontend/src/screens/Isochrone.jsx
licarpen: Makes sense.  Thanks for clarifying the reasoning behind this preference.

frontend/src/screens/Isochrone.jsx
licarpen: Got it.  The componentDidUpdate will be called once the routes are fetched and state is updated, so the componentDidMount either needs to fetchRoutes or setInitialLocation but not both.

licarpen:
Changes have been made to reflect the feedback and suggestions above.

PR #615, opened 2020-03-26
Dependency version bump

akgupta89:
## Proposed changes

- Bumped everything up automatically with semver
- Should theoretically work fine, am unable to test due to terrible wifi adapter **(PLEASE TEST BEFORE APPROVING)**
- Should resolve some issues with vunerabilities


hathix:
Testing process:

* Pulling your branch
* Running `docker-compose up`
* Opening `localhost:3000`

It builds and runs totally fine for me.

youngj:
BTW, it looks like this PR broke development. Running `docker-compose up` is not sufficient to test changes to package.json -- it is necessary to run `docker-compose build` first. The breaking issues appear to be fixed in #629 but I'm not sure if there are any more subtle issues that bumping the versions of so many packages will cause.

Issue #609, opened 2020-03-26
Store cache files as CSV instead of JSON


hathix:
On #580, Jesse left this feedback:

> Good call on storing cache files as CSV to simplify appending chunks.
> It seems like we don't need to store the JSON format anymore:
>    trynapi.get_state could rename each temporary CSV file to a permanent cache path when it's done.
>    CachedState.get_for_route could return a pandas DataFrame similar to the DataFrame constructed from JSON by eclipses.produce_buses.
>    the route_state parameter in eclipses.py could be a DataFrame instead of a dict.

> This probably would simplify the code as well as improving performance. Do you want to work on this now or create an issue?

Eddy said:

>I've updated the PR message with some performance comparisons. Good point that using CSV's would simplify things in eclipses and improve performance there - I'll make an issue for it. I'll let you merge this as to ensure it works properly on the Kubernetes/GCloud environment.

_Originally posted by @EddyIonescu in https://github.com/trynmaps/metrics-mvp/pull/580#issuecomment-591233480_

PR #618, opened 2020-03-29
use CSV for storing per-route state instead of JSON for better performance

youngj:
When fetching the state from tryn-api to compute arrivals, profiling showed that a significant amount of running time was spent in read_temp_chunk_state, which loaded a cache file in CSV format into memory, converted the data structure, and wrote a file in JSON format. However, the JSON format was only used by eclipses.produce_buses to initialize a DataFrame, which could also be done from the original CSV format without using JSON. 

This PR removes the JSON format and updates CachedState.get_for_route to return the DataFrame that used to be returned by eclipses.produce_buses. eclipses.find_arrivals is updated to take the DataFrame as a parameter instead of a dict.

### DIFF
--- a/backend/models/trynapi.py
+++ b/backend/models/trynapi.py
@@ -17,6 +17,9 @@ class CachedState:
         self.cache_paths[route_id] = cache_path
 
     def get_for_route(self, route_id) -> pd.DataFrame:
+        if route_id not in self.cache_paths:
+            return None
+
         cache_path = self.cache_paths[route_id]
         print(f'loading state for route {route_id} from cache: {cache_path}')
         buses = pd.read_csv(
--- a/backend/models/trynapi.py
+++ b/backend/models/trynapi.py
@@ -16,12 +16,34 @@ class CachedState:
     def add(self, route_id, cache_path):
         self.cache_paths[route_id] = cache_path
 
-    def get_for_route(self, route_id):
+    def get_for_route(self, route_id) -> pd.DataFrame:
         cache_path = self.cache_paths[route_id]
         print(f'loading state for route {route_id} from cache: {cache_path}')
-        with open(cache_path, "r") as f:
-            return json.loads(f.read())
+        buses = pd.read_csv(
+                cache_path,
+                dtype={
+                    'vid': str,
+                    'did': str,
+                },
+                float_precision='high', # keep precision for rounding lat/lon
+            ) \
+            .rename(columns={
+                'lat': 'LAT',
+                'lon': 'LON',
+                'vid': 'VID',
+                'did': 'DID',
+                'secsSinceReport': 'AGE',
+                'timestamp': 'RAW_TIME'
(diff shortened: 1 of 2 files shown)


backend/models/trynapi.py
```diff
@@ -16,12 +16,34 @@ def __init__(self):
...
-
+        buses = pd.read_csv(
+                cache_path,
+                dtype={
+                    'vid': str,
+                    'did': str,
+                },
+                float_precision='high', # keep precision for rounding lat/lon
+            ) \
+            .rename(columns={
```
EddyIonescu: It doesn't look like the `\` is needed anymore.

backend/models/trynapi.py
```diff
@@ -92,17 +114,19 @@ def get_state(agency_id: str, d: date, start_time, end_time, route_ids) -> Cache
...
         cache_dir = Path(cache_path).parent
         if not cache_dir.exists():
             cache_dir.mkdir(parents = True, exist_ok = True)
 
-        with open(cache_path, "w") as f:
-            print(f'writing state for route {route_id} to cache: {cache_path}')
-            f.write(json.dumps(states))
+        temp_cache_path = get_route_temp_cache_path(agency_id, route_id)
+        if not os.path.exists(temp_cache_path):
+            return None
```
EddyIonescu: When does this case happen - when the file doesn't get downloaded? Also it goes against the type header.

youngj: Yeah this was a bug, changed to `continue`. It happens when no data for that route was returned by tryn-api.

backend/models/trynapi.py
youngj: Without it, there would be `IndentationError: unexpected indent`

youngj:
For a daily TriMet run, the runtime just to fetch and cache the state was was 6m43s for this version compared to 11m46s with the JSON version, so it saved about 5 minutes. (`time python get_state.py --agency=trimet --date=2020-03-25`) 

These .csv files are only cached locally within the Docker container and aren't saved to S3 (neither were the .json files) so there is no difference in storage costs. The .json.gz files containing the arrival times saved by compute_arrivals.py should be identical to the previous .json.gz files.

backend/models/trynapi.py
EddyIonescu: Makes sense, but previously it would've written an empty file? I assume you've checked that this wouldn't lead to errors elsewhere.

EddyIonescu:
I see, this just removes the JSON write/read that wasn't needed anymore - so no plans to change the format of anything outputted to S3. Thanks for doing https://github.com/trynmaps/metrics-mvp/issues/609!

backend/models/trynapi.py
EddyIonescu: I see, didn't notice the indentation was off, didn't know `\` could override that; I guess it makes it more readable.

PR #625, opened 2020-04-02
if no routes currently have active services, save all routes

youngj:
The current GTFS feed from SF Muni starts on April 4, which caused get_active_routes to save an empty list of routes. This PR updates the behavior so if there are no active services, all routes in the GTFS feed will be saved.

It also removes a muni custom direction which no longer exists in the GTFS feed.

### DIFF
--- a/backend/models/gtfs.py
+++ b/backend/models/gtfs.py
@@ -992,7 +992,7 @@ class GtfsScraper:
         for service_date in dates_map:
             if service_date <= d:
                 before_service_ids += dates_map[service_date]
-            elif service_date >= d:
+            if service_date >= d:
                 after_service_ids += dates_map[service_date]
         active_services = set.intersection(
             set(before_service_ids),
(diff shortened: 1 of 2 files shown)


backend/models/gtfs.py
```diff
@@ -992,7 +992,7 @@ def get_active_routes(self, routes_df, d):
         for service_date in dates_map:
             if service_date <= d:
                 before_service_ids += dates_map[service_date]
-            elif service_date >= d:
+            if service_date >= d:
```
EddyIonescu, line 995: Thanks for catching this!

youngj:
Thanks for explaining how agencies update the GTFS feed a few days ahead. On GKE there is a fresh /data directory each time save_routes.py runs, so the previous GTFS feed would not be stored there, but I suppose save_routes.py could just skip saving the routes in that case?

EddyIonescu:
@youngj I see, yes skipping it seems to be the better option. Sorry, I left resolving this issue (just skipping the update) for a future PR that I never ended up doing.

I ran save_routes.py using the previous GTFS that's valid for today with the --s3 option, so you only have to restart the backend as to reload the routes file for it to work.

So if the length of active routes is 0, then just return with an error and not update.

PR #627, opened 2020-04-02
Skip updating GTFS if no active routes on date being run

EddyIonescu:
Agencies normally update their GTFS a few days ahead so the one that gets downloaded ends up having no active routes. This could result in the wrong routes being shown on the dates just before the changes take effect, if for example a route is added or removed.

If there are no active routes on the provided date, save_routes will raise an exception and the routes will not be updated. The older version of routes will be used by the backend, which should result in the expected behaviour.

There could be the case where an agency schedules zero routes for the entire duration of their GTFS, but I don't see that happening.

### DIFF
--- a/backend/models/gtfs.py
+++ b/backend/models/gtfs.py
@@ -1062,9 +1062,13 @@ class GtfsScraper:
         agency = self.agency
         agency_id = agency.id
         routes_df = self.get_gtfs_routes()
-        active_routes_df = self.get_active_routes(routes_df, d)
-        if len(active_routes_df) > 0:
-            routes_df = active_routes_df
+        routes_df = self.get_active_routes(routes_df, d)
+        if len(routes_df) == 0:
+            self.errors.append((
+                f'Zero active routes for {agency_id}, the routes config was not updated. '
+                f'Ensure the GTFS is active for the given date {d}'
+            ))
+            return
 
         routes_data = [
             self.get_route_data(route)


backend/models/gtfs.py
```diff
@@ -1062,9 +1062,9 @@ def save_routes(self, save_to_s3, d):
         agency = self.agency
         agency_id = agency.id
         routes_df = self.get_gtfs_routes()
-        active_routes_df = self.get_active_routes(routes_df, d)
-        if len(active_routes_df) > 0:
-            routes_df = active_routes_df
+        routes_df = self.get_active_routes(routes_df, d)
+        if len(routes_df) == 0:
+            raise Exception('Zero active routes, ensure the GTFS is active for today.')
```
youngj, line 1067: save_routes.py may loop over multiple agencies, so an unhandled exception would block other agencies from being saved. To avoid this, one way would be to append an error message to the `self.errors` list, like how it works for custom directions. In that case, save_routes.py would raise an exception after processing all of the agencies.

EddyIonescu: @youngj didn't see that - fixed. Also it looks like there's no issue with timetables and scheduled_stats continuing as they use the GTFS start/end dates.

EddyIonescu: Also s3://opentransit-data/routes/v3a/routes_v3a_muni.json.gz is now valid but the backend is still using the old one without any routes. So on the April 3rd run it should keep using that file.

PR #628, opened 2020-04-06
detects when none of the days of week selected are in the date range …

Brian-Lee:
## Proposed changes
This detects when the user has not selected any days of the week that
are actually included in their date range, alerts them with a snackbar, and disables the 'Apply' button.

Without this functionality, the user would see an ugly error msg about arrays to concatenate.
...

## Demo
https://metrics-mvp-brians-branch.herokuapp.com/

## Screenshot

image


Brian-Lee:
The first commit has the functionality. The rest of the commits deal with lint errors. Many of those lint errors were in the file I changed (DateTimePopover.jsx), but I also had to fix lint errors in many files that I didn't change for some reason. I suspect this is due to changing lint requirements.

hathix:
Let me test your branch locally. One quick suggestion: I would use a Material Snackbar (https://material-ui.com/components/snackbars/) instead of a native JS alert. Using Material will feel less disruptive to the user, and more professional.

hathix:
Also, what happens when you dismiss the alert? Does the user's change not get applied?

Now that I think about it, I think the best UX would be to disable the "apply" button (and not apply any changes) if the user has chosen the wrong set of days of the week (like when none of the days of the week are in the date range). This would prevent this error from happening in the first place.

hathix:
OK, I tried it locally and it indeed lets you apply the change even after showing the alert, which leads to a screen with no data. Let's just disable the "apply" button when invalid information is entered in the dropdown, as I suggested in the last column.

frontend/src/components/DateTimePopover.jsx
```diff
@@ -89,12 +89,69 @@ function DateTimePopover(props) {
   const classes = useStyles();
   const maxDate = Moment(Date.now()).format('YYYY-MM-DD');
 
+  /**
+   * Returns an array of dates between the two dates.
+   */
+  function getDates(startDate, endDate) {
```
hathix, line 95: Make this function name little more descriptive? Like `getDatesBetween`?

frontend/src/components/DateTimePopover.jsx
```diff
@@ -89,12 +89,69 @@ function DateTimePopover(props) {
...
+    const addDays = function(days) {
+      const date = new Date();
+      date.setDate(date.getDate() + days);
+      return date;
+    };
+    while (currentDate <= endDate) {
+      dates.push(currentDate);
+      currentDate = addDays.call(currentDate, 1);
+    }
+    return dates;
```
hathix, line 107: Coding style suggestion: you might be able to replace this with a single for loop, like

```
for (date d = startDate; d <= endDate; d.setDate(d.getDate() + 1)) {
 ...
}
```

frontend/src/components/DateTimePopover.jsx
hathix: There might be a function in Moment that does this for you automatically

frontend/src/components/DateTimePopover.jsx
```diff
@@ -89,12 +89,69 @@ function DateTimePopover(props) {
...
     const newGraphParams = Object.assign({}, graphParams);
     newGraphParams[targetRange] = localDateRangeParams;
+    const year = newGraphParams.firstDateRange.startDate.substring(0, 4);
+    // month starts at 0 for January so subtract 1
+    const month = newGraphParams.firstDateRange.startDate.substring(5, 7) - 1;
+    const day = newGraphParams.firstDateRange.startDate.substring(8, 10);
+    const endYear = newGraphParams.firstDateRange.date.substring(0, 4);
+    // month starts at 0 for January so subtract 1
+    const endMonth = newGraphParams.firstDateRange.date.substring(5, 7) - 1;
+    const endDay = newGraphParams.firstDateRange.date.substring(8, 10);
```
hathix, line 126: You might be able to use the Moment library to parse the date strings for you, instead of having to do it manually like this.

frontend/src/components/DateTimePopover.jsx
```diff
@@ -89,12 +89,69 @@ function DateTimePopover(props) {
...
+
+    for (i = 0; i < dates.length; i++) {
+      dowsUsed[dates[i].getDay()] = true;
+    }
+
+    // If the combination of the daterange and the days of the week result in
+    // at least one day being selected, atLeastOneDaySelected = true
+    let atLeastOneDaySelected = false;
+
+    for (i = 0; i < 7; i++) {
```
hathix, line 141: the "7" should be replaced with `dowsUsed.length` just so it's extra-obvious where that number came from

frontend/src/components/DateTimePopover.jsx
Brian-Lee: That's way better. TY.

frontend/src/components/DateTimePopover.jsx
Brian-Lee: ty

frontend/src/components/DateTimePopover.jsx
Brian-Lee: Good suggestion. I managed it :)

frontend/src/components/DateTimePopover.jsx
Brian-Lee: I looked for a Moment function that does this. I don't think there is one just for that.

Brian-Lee:
> Let me test your branch locally. One quick suggestion: I would use a Material Snackbar (https://material-ui.com/components/snackbars/) instead of a native JS alert. Using Material will feel less disruptive to the user, and more professional.

I managed to change the native javacript alert to a snackbar.

Brian-Lee:
I have incorporated much of Neels great ideas/feedback. Please review.

frontend/src/components/DateRangeControl.jsx
```diff
@@ -109,6 +134,37 @@ function DateRangeControl(props) {
     dateRangeParams,
   );
 
+  let j;
+  const dowsUsed = [false, false, false, false, false, false, false];
+  const dates = enumerateDaysBetweenDates(
+    new Date(
+      Moment(localDateRangeParams.startDate, 'YYYY-MM-DD').year(),
+      Moment(localDateRangeParams.startDate, 'YYYY-MM-DD').month(),
+      Moment(localDateRangeParams.startDate, 'YYYY-MM-DD').date(),
```
hathix, line 143: I would create this Moment object before calling this function so that you don't have to make a new object for each bit of this function call

frontend/src/components/DateRangeControl.jsx
```diff
@@ -284,6 +340,7 @@ function DateRangeControl(props) {
 
   return (
     <>
+      <NoDaysSelectedAlert showAlert={!atLeastOneDaySelected} />
```
hathix, line 343: Very nice! Elegant solution.

frontend/src/components/DateRangeControl.jsx
```diff
@@ -109,6 +134,37 @@ function DateRangeControl(props) {
...
+  let j;
+  const dowsUsed = [false, false, false, false, false, false, false];
+  const dates = enumerateDaysBetweenDates(
+    new Date(
+      Moment(localDateRangeParams.startDate, 'YYYY-MM-DD').year(),
+      Moment(localDateRangeParams.startDate, 'YYYY-MM-DD').month(),
+      Moment(localDateRangeParams.startDate, 'YYYY-MM-DD').date(),
+    ),
+    new Date(
+      Moment(localDateRangeParams.date, 'YYYY-MM-DD').year(),
```
hathix, line 146: Same as above, we should pre-create the Moment object here

frontend/src/components/DateRangeControl.jsx
```diff
@@ -109,6 +134,37 @@ function DateRangeControl(props) {
     dateRangeParams,
   );
 
+  let j;
+  const dowsUsed = [false, false, false, false, false, false, false];
+  const dates = enumerateDaysBetweenDates(
```
hathix, line 139: Can't this function just take two Moment objects as input? It looks like we currently take Moment objects, turn them into Dates, pass them to this function, and turn them *back* into Moment objects. It would be simpler to just have that function take Moment objects as input and then pass Moment objects here.

frontend/src/components/DateRangeControl.jsx
Brian-Lee: I  addressed this in PR #634

frontend/src/components/DateRangeControl.jsx
Brian-Lee: I  addressed this in PR #634 
Thanks for the helpful suggestion!

frontend/src/components/DateRangeControl.jsx
Brian-Lee: Yes. I've addressed this in PR 634

frontend/src/components/DateRangeControl.jsx
Brian-Lee: TY

PR #629, opened 2020-04-08
Fix local docker-compose

rickpr:
- Match ESLint versions
- Use a TTY as it's required for React Development

## Proposed changes
Use `babel-eslint` 10.1.0 as it is required by `normalize.css` already, breaking the dependency tree if another version is used.

Use a TTY for local `docker-compose`. See https://github.com/facebook/create-react-app/issues/8688 for more information as to why this.

## Screenshot
https://sfbrigade.slack.com/archives/C7HUYBHAA/p1586304906043500
image

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -40,3 +40,4 @@ services:
     environment:
       NODE_ENV: development
       REACT_APP_METRICS_BASE_URL: http://localhost:5000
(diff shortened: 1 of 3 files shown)


Brian-Lee:
I ran this and it solved my "docker-compose up" npm ERR.

PR #638, opened 2020-04-24
Brians branch

### LINKED ISSUE #626: Allow running save_routes on historic GTFS files
Currently we only store the latest version of route-configs, but have versioning for timetables.

save_routes.py runs daily and updates these using the recentmost GTFS that is available.

Using TransitFeeds (https://transitfeeds.com/api/swagger/) we can run save_routes.py on older versions of the GTFS based on the start/end dates of each GTFS file. Then we'll be able to show scheduled times and the proper routes for dates in the past.

For the most part, the agency config can be changed while remaining compatible with previous versions of the agency's network. However, for cases like where a branch is completely changed or the routes are redesigned, a different version would be required here too (see https://github.com/trynmaps/metrics-mvp/issues/127), which can be done in another issue and PR.

Brian-Lee:
Fixes #626

## Proposed changes
Get the feed from transitfeeds (https://api.transitfeeds.com/v1/getFeedVersions?key={your-key}3&feed=sfmta%2F60&page=1&limit=10&err=1&warn=1)
Use get the last 10 (changeable) previous versions of the GTFS files.
Use those files in save_routes.py to save routes to subfolders in /backend/data.
Subfolders and files have date suffixs.

These older versions of route JSON files can then be used for historical queries.
...

## Screenshot
folder_screenshot

...

...

### DIFF
--- a/backend/models/gtfs.py
+++ b/backend/models/gtfs.py
@@ -1101,4 +1101,4 @@ class GtfsScraper:
 
         routes = [routeconfig.RouteConfig(agency_id, route_data) for route_data in routes_data]
 
-        routeconfig.save_routes(agency_id, routes, save_to_s3=False, use_versioning=True, version_date=d)
+        routeconfig.save_routes(agency_id, routes, save_to_s3=False, version_date=d)
--- a/backend/models/gtfs.py
+++ b/backend/models/gtfs.py
@@ -261,7 +261,8 @@ class GtfsScraper:
         agency_id = self.agency_id
 
         dates_map = self.get_services_by_date()
-
+        ##bri##  print('here\n\n\n',dates_map)
+        ##bri##  exit()
         #
         # Typically, many dates have identical scheduled timetables (with times relative to midnight on that date).
         # Instead of storing redundant timetables for each date, store one timetable per route for each unique set of service_ids.
@@ -1100,4 +1101,4 @@ class GtfsScraper:
 
         routes = [routeconfig.RouteConfig(agency_id, route_data) for route_data in routes_data]
 
-        routeconfig.save_routes(agency_id, routes, save_to_s3=False)
+        routeconfig.save_routes(agency_id, routes, save_to_s3=False, use_versioning=True, version_date=d)
--- a/backend/models/gtfs.py
+++ b/backend/models/gtfs.py
@@ -70,6 +70,7 @@ def download_gtfs_data(agency: config.Agency, gtfs_cache_dir, archiving_old=Fals
         # to reuse for the archiving passes
 
         if cache_dir.exists():	
+            #exit()
             shutil.rmtree(cache_dir)
             print('removed',cache_dir)
             os.remove(zip_path)
--- a/backend/models/gtfs.py
+++ b/backend/models/gtfs.py
@@ -50,12 +50,32 @@ def get_stop_geometry(stop_xy, shape_lines_xy, shape_cumulative_dist, start_inde
         'offset': int(best_offset) # distance in meters between this stop and the closest line segment of shape
     }
 
-def download_gtfs_data(agency: config.Agency, gtfs_cache_dir):
-    gtfs_url = agency.gtfs_url
+def download_gtfs_data(agency: config.Agency, gtfs_cache_dir, archiving_old=False):
+    cache_dir = Path(gtfs_cache_dir)
+    zip_path = f'{util.get_data_dir()}/gtfs-{agency.id}.zip'
+    import os
+    if archiving_old == False:
+        gtfs_url = agency.gtfs_url
+    else:
+        '''
+        get an old GFTS file from 2020-02-19
+        https://transitfeeds.com/p/sfmta/60/20200219/download	
+        '''
+        gtfs_url = "https://transitfeeds.com/p/sfmta/60/20200219/download"
+        # need to delete existing zip file and directory in order
+        # to reuse for the archiving passes
+
+        if cache_dir.exists():	
+            import shutil
+            shutil.rmtree(cache_dir)
+            print('removed',cache_dir)
+            os.remove(zip_path)
+            print('removed',zip_path)					
+	
     if gtfs_url is None:
         raise Exception(f'agency {agency.id} does not have gtfs_url in config')
 
-    cache_dir = Path(gtfs_cache_dir)
+
     if not cache_dir.exists():
         print(f'downloading gtfs data from {gtfs_url}')
         r = requests.get(gtfs_url)
@@ -63,6 +83,7 @@ def download_gtfs_data(agency: config.Agency, gtfs_cache_dir):
         if r.status_code != 200:
             raise Exception(f"Error fetching {gtfs_url}: HTTP {r.status_code}: {r.text}")
 
+        ##bri## should not redefine
         zip_path = f'{util.get_data_dir()}/gtfs-{agency.id}.zip'
 
         with open(zip_path, 'wb') as f:
@@ -139,13 +160,13 @@ def contains_excluded_stop(shape_stop_ids, excluded_stop_ids):
     return False
 
 class GtfsScraper:
-    def __init__(self, agency: config.Agency):
+    def __init__(self, agency: config.Agency, archiving_old=False):
         self.agency = agency
         self.agency_id = agency_id = agency.id
         gtfs_cache_dir = f'{util.get_data_dir()}/gtfs-{agency_id}'
 
-        #download_gtfs_data(agency, gtfs_cache_dir)
-        download_old_gtfs_data(agency, gtfs_cache_dir)
+        download_gtfs_data(agency, gtfs_cache_dir, archiving_old=archiving_old)
+        #download_old_gtfs_data(agency, gtfs_cache_dir)
 
         self.feed = ptg.load_geo_feed(gtfs_cache_dir, {})
 
--- a/backend/models/gtfs.py
+++ b/backend/models/gtfs.py
@@ -1079,3 +1079,25 @@ class GtfsScraper:
         routes = [routeconfig.RouteConfig(agency_id, route_data) for route_data in routes_data]
 
         routeconfig.save_routes(agency_id, routes, save_to_s3=save_to_s3)
+
+    def save_old_routes(self, save_to_s3, d):
+        agency = self.agency
+        agency_id = agency.id
+        routes_df = self.get_gtfs_routes()
+        routes_df = self.get_active_routes(routes_df, d)
+        if len(routes_df) == 0:
+            self.errors.append((
+                f'Zero active routes for {agency_id}, the routes config was not updated. '
+                f'Ensure the GTFS is active for the given date {d}'
+            ))
+            return
+
+        routes_data = [
+            self.get_route_data(route)
+            for route in routes_df.itertuples()
+        ]
+        routes_data = self.sort_routes(routes_data)
+
+        routes = [routeconfig.RouteConfig(agency_id, route_data) for route_data in routes_data]
+
+        routeconfig.save_routes(agency_id, routes, save_to_s3=save_to_s3)
--- a/backend/models/gtfs.py
+++ b/backend/models/gtfs.py
@@ -59,6 +59,8 @@ def download_gtfs_data(agency: config.Agency, gtfs_cache_dir, archiving_old=Fals
         gtfs_url = agency.gtfs_url
     else:
         '''
+		need to set up a system for properly getting URLs for archiving routes
+		
         get an old GFTS file from 2020-02-19
         https://transitfeeds.com/p/sfmta/60/20200219/download	
         '''
--- a/backend/models/gtfs.py
+++ b/backend/models/gtfs.py
@@ -52,7 +52,7 @@ def get_stop_geometry(stop_xy, shape_lines_xy, shape_cumulative_dist, start_inde
         'offset': int(best_offset) # distance in meters between this stop and the closest line segment of shape
     }
 
-def download_gtfs_data(agency: config.Agency, gtfs_cache_dir, archiving_old=False):
+def download_gtfs_data(agency: config.Agency, gtfs_cache_dir, archiving_old=False, archiving_url=None):
     cache_dir = Path(gtfs_cache_dir)
     zip_path = f'{util.get_data_dir()}/gtfs-{agency.id}.zip'
     if archiving_old == False:
@@ -64,7 +64,8 @@ def download_gtfs_data(agency: config.Agency, gtfs_cache_dir, archiving_old=Fals
         get an old GFTS file from 2020-02-19
         https://transitfeeds.com/p/sfmta/60/20200219/download	
         '''
-        gtfs_url = "https://transitfeeds.com/p/sfmta/60/20200219/download"
+        #####gtfs_url = "https://transitfeeds.com/p/sfmta/60/20200219/download"
+        gtfs_url = archiving_url
         # need to delete existing zip file and directory in order
         # to reuse for the archiving passes
 
@@ -84,9 +85,6 @@ def download_gtfs_data(agency: config.Agency, gtfs_cache_dir, archiving_old=Fals
 
         if r.status_code != 200:
             raise Exception(f"Error fetching {gtfs_url}: HTTP {r.status_code}: {r.text}")
-
-        ##bri## should not redefine
-        zip_path = f'{util.get_data_dir()}/gtfs-{agency.id}.zip'
 
         with open(zip_path, 'wb') as f:
             f.write(r.content)
@@ -133,12 +131,12 @@ def contains_excluded_stop(shape_stop_ids, excluded_stop_ids):
     return False
 
 class GtfsScraper:
-    def __init__(self, agency: config.Agency, archiving_old=False):
+    def __init__(self, agency: config.Agency, archiving_old=False, archiving_url=None):
         self.agency = agency
         self.agency_id = agency_id = agency.id
         gtfs_cache_dir = f'{util.get_data_dir()}/gtfs-{agency_id}'
 
-        download_gtfs_data(agency, gtfs_cache_dir, archiving_old=archiving_old)
+        download_gtfs_data(agency, gtfs_cache_dir, archiving_old=archiving_old, archiving_url=archiving_url)
         #download_old_gtfs_data(agency, gtfs_cache_dir)
 
         self.feed = ptg.load_geo_feed(gtfs_cache_dir, {})
--- a/backend/models/gtfs.py
+++ b/backend/models/gtfs.py
@@ -70,6 +70,37 @@ def download_gtfs_data(agency: config.Agency, gtfs_cache_dir):
 
         with zipfile.ZipFile(zip_path, 'r') as zip_ref:
             zip_ref.extractall(gtfs_cache_dir)
+			
+			
+			
+
+
+def download_old_gtfs_data(agency: config.Agency, gtfs_cache_dir):
+    '''
+    get an old GFTS file from 2020-02-19
+    https://transitfeeds.com/p/sfmta/60/20200219/download	
+    '''
+    #gtfs_url = agency.gtfs_url
+    gtfs_url = "https://transitfeeds.com/p/sfmta/60/20200219/download"
+    if gtfs_url is None:
+        raise Exception(f'agency {agency.id} does not have gtfs_url in config')
+
+    cache_dir = Path(gtfs_cache_dir)
+    if not cache_dir.exists():
+        print(f'downloading gtfs data from {gtfs_url}')
+        r = requests.get(gtfs_url)
+
+        if r.status_code != 200:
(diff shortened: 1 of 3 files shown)


backend/save_routes.py
```diff
@@ -52,12 +54,38 @@
...
+        to get old routes 
+        and cache them in date versioned folders
+		
+        '''
+
+        while(len(archiving_dates) > 0):
+            archiving_date = archiving_dates.pop()
+            archiving_url = archiving_urls.pop()
+            scraper_archiving = gtfs.GtfsScraper(agency, archiving_url=archiving_url)			
+            scraper_archiving.save_routes(False, d, version_date=archiving_date)
```
EddyIonescu, line 87: `archiving_date` should be passed in instead of `d`

backend/save_routes.py
```diff
@@ -52,12 +54,38 @@
 
     save_to_s3 = args.s3
     d = date.today()
-
+    archive_date = date.today()
     errors = []
 
+    limit = '10'
+    urls_feed = 'https://api.transitfeeds.com/v1/getFeedVersions?key=' + transitfeeds_api_key + '&feed=sfmta%2F60&page=1&limit=' + limit + '&err=1&warn=1'
```
EddyIonescu, line 61: Feed URL should be in the agency config file. Fetching from TransitFeeds is out-of-scope for issue #626 though, see my other comments.

EddyIonescu: Also it looks like their limit is 10 so providing that isn't necessary; getting more than 10 GTFS files requires passing in the page number to the transitfeed API.

backend/save_routes.py
```diff
@@ -52,12 +54,38 @@
 
     save_to_s3 = args.s3
     d = date.today()
-
+    archive_date = date.today()
```
EddyIonescu, line 57: This line doesn't look necessary.

backend/save_routes.py
Brian-Lee: TY. I had to convert archiving_date from a string to a date, so I put in datetime.strptime(archiving_date, "%Y-%m-%d").date()

backend/save_routes.py
Brian-Lee: Right. I removed it.

backend/save_routes.py
Brian-Lee: I got rid of the downloading of gtfs files from transitfeeds

Brian-Lee:
I made a bunch of changes and I think I addressed everything mentioned. 
However,  I am not sure that the Nextbus files are working correctly. Shouldn't they also be in date versioned folders?

backend/models/gtfs.py
```diff
@@ -49,28 +52,40 @@ def get_stop_geometry(stop_xy, shape_lines_xy, shape_cumulative_dist, start_inde
...
+            r = requests.get(gtfs_url)
 
-        zip_path = f'{util.get_data_dir()}/gtfs-{agency.id}.zip'
+            if r.status_code != 200:
+                raise Exception(f"Error fetching {gtfs_url}: HTTP {r.status_code}: {r.text}")
+
+            with open(zip_path, 'wb') as f:
+                f.write(r.content)
+
+            with zipfile.ZipFile(zip_path, 'r') as zip_ref:
```
EddyIonescu, line 78: These lines are the same as lines 85/86. I don't think they needed to be copy/pasted.

backend/save_routes.py
```diff
@@ -34,39 +37,70 @@
...
+            # check if this zip file exists
+            loops = 0
+            max_loops = 365
+            gtfs_date_to_use = gtfs_date
+            while Path(gtfs_path).is_file() == False and loops < max_loops:
+                # go back one day and re-represent date as a string
+                gtfs_date_to_use = (datetime.strptime(gtfs_date_to_use, '%Y-%m-%d') - timedelta(days=1)).strftime('%Y-%m-%d') 		
+                gtfs_path = f'{util.get_data_dir()}/gtfs-{agency.id}-{gtfs_date_to_use}.zip'
+                gtfs_cache_dir_to_use = f'{util.get_data_dir()}/gtfs-{agency.id}-{gtfs_date_to_use}.zip'
+                loops += 1
```
EddyIonescu, line 89: While this does work, what I was thinking was:
- get the list of files with the agency suffix (using os.listdir)
- then iterate through each filename, taking the recentmost one before date
- this removes the need for iterating for every day (which would also cause an issue if a GTFS hasn't been updated in over a year)
- does this seem simpler for you?

backend/save_routes.py
```diff
@@ -34,39 +37,70 @@
...
-#
+# UPDATE: We are now saving some older routes in versioned directories in metrics-mvp/backend/data
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(description='Save route configuration from GTFS and possibly Nextbus API')
     parser.add_argument('--agency', required=False, help='Agency ID')
     parser.add_argument('--s3', dest='s3', action='store_true', help='store in s3')
     parser.add_argument('--timetables', dest='timetables', action='store_true', help='also save timetables')
     parser.add_argument('--scheduled-stats', dest='scheduled_stats', action='store_true', help='also compute scheduled stats if the timetable has new dates (requires --timetables)')
+    parser.add_argument('--gtfs_date', required=False)
```
EddyIonescu, line 48: I think this could just be date.

backend/models/gtfs.py
Brian-Lee: Combined now.

backend/save_routes.py
Brian-Lee: I changed the command line argument to be called just "date", but I left the internal variables named gtfs_date. Let me know if you think those should be changed as well.

backend/save_routes.py
Brian-Lee: Good suggestion.

Brian-Lee: I rewrote that part. It may well be in need of improvement.

backend/save_routes.py
```diff
@@ -34,39 +38,79 @@
...
 
     save_to_s3 = args.s3
-    d = date.today()
+    gtfs_date = args.date 
 
     errors = []
-
+	
+	# should probably change things so we supply gtfs_path 
+	# instead of gtfs_date
```
EddyIonescu, line 65: Comments like "should" should link to an issue, for now I don't think a comment's needed here

backend/save_routes.py
```diff
@@ -34,39 +38,79 @@
...
+	
+        if gtfs_date is None:
+            # save the normal way, downloading the most recent GTFS file
+            # should probably not be using both date_to_use and gtfs_date_to_use
+            date_to_use=date.today()
+            gtfs_date_to_use=date.today()
+            gtfs_path = None
+        else:
+            # save with date suffix, using the GTFS file provided
+            save_to_s3=False
```
EddyIonescu, line 77: I don't think this is needed as save_to_s3 is False by default

backend/save_routes.py
```diff
@@ -34,39 +38,79 @@
...
+            date_to_use=datetime.strptime(gtfs_date, "%Y-%m-%d").date()	
+            gtfs_path = f'{util.get_data_dir()}/gtfs-{agency.id}-{gtfs_date}.zip'
+ 
+            # find most recent zip file before gtfs_date
+            best_candidate_zip_file = ""
```
EddyIonescu, line 82: What does best candidate mean? Does that variable contain a file or a path? Would something like `latest_past_gtfs_path` work better?

backend/save_routes.py
```diff
@@ -34,39 +38,79 @@
...
+            best_candidate_date = datetime.today()
```
EddyIonescu, line 83: Shouldn't this be the date that was passed in?

backend/models/gtfs.py
```diff
@@ -49,28 +52,40 @@ def get_stop_geometry(stop_xy, shape_lines_xy, shape_cumulative_dist, start_inde
...
 
-        with open(zip_path, 'wb') as f:
-            f.write(r.content)
+            #with zipfile.ZipFile(zip_path, 'r') as zip_ref:
+            #    zip_ref.extractall(gtfs_cache_dir)
 
-        with zipfile.ZipFile(zip_path, 'r') as zip_ref:
-            zip_ref.extractall(gtfs_cache_dir)
+    else:
+        cache_dir = Path(gtfs_cache_dir)
```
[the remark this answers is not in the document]
Brian-Lee: TY. I've combined those.

backend/models/gtfs.py
```diff
@@ -49,28 +52,40 @@ def get_stop_geometry(stop_xy, shape_lines_xy, shape_cumulative_dist, start_inde
...
+        zip_path = gtfs_path
```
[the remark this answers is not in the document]
Brian-Lee: Good suggestion. I've made that change.

backend/models/routeconfig.py
```diff
@@ -121,8 +122,17 @@ def get_directions_for_stop(self, stop_id):
...
         ]
 
-def get_cache_path(agency_id, version=DefaultVersion):
-    return f'{util.get_data_dir()}/routes_{version}_{agency_id}.json'
+def get_cache_path(agency_id, version=DefaultVersion, version_date=None):
+    # version_date is for saving old versions of routes
+    # It has nothing to do with version=DefaultVersion
+    if version_date == None:
+        return f'{util.get_data_dir()}/routes_{version}_{agency_id}.json'
+    else:
```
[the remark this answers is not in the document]
Brian-Lee: Ok. I removed it.

backend/save_routes.py
Brian-Lee: TY. I've removed those comments.

backend/save_routes.py
Brian-Lee: Removed.

backend/save_routes.py
Brian-Lee: There were two variables, best_candidate_zip_file (a file) and best_candidate_date (a date.)

In an effort to make this more clear, I added a comment and changed those variable names to:
recentmost_date_qualified_zip_file 
recentmost_date_qualified_date

Let me know if that works. Good names are a little challenging for this.

backend/save_routes.py
Brian-Lee: Yes. I've made that change.

backend/save_routes.py
```diff
@@ -34,39 +38,85 @@
...
+
+            recentmost_date_qualified_zip_file = ""
+            recentmost_date_qualified_date = gtfs_date
+            smallest_timedelta_so_far = timedelta.max
+            for candidate_zip_file in os.listdir(util.get_data_dir()):
+                if f'gtfs-{agency.id}-' in candidate_zip_file and '.zip' in candidate_zip_file:
+                    candidate_year = candidate_zip_file.split('-')[2]
+                    candidate_month = candidate_zip_file.split('-')[3]
+                    candidate_day = candidate_zip_file.split('-')[4]
+                    candidate_day = candidate_day.split(".zip")[0]
```
EddyIonescu, line 96: You could chain `.split('.zip')[0]` to the previous line

backend/models/gtfs.py
```diff
@@ -8,6 +8,9 @@
 import gzip
 import hashlib
 import zipfile
+import shutil
```
EddyIonescu, line 11: Is this being used anywhere?

backend/models/routeconfig.py
```diff
@@ -121,8 +122,15 @@ def get_directions_for_stop(self, stop_id):
             for s in direction['stops'] if s == stop_id
         ]
 
-def get_cache_path(agency_id, version=DefaultVersion):
-    return f'{util.get_data_dir()}/routes_{version}_{agency_id}.json'
+def get_cache_path(agency_id, version=DefaultVersion, version_date=None):
+    # version_date is for saving old versions of routes
+    # It has nothing to do with version=DefaultVersion
```
EddyIonescu, line 127: I think we need better naming so that this comment won't be needed.
Could it be renamed to something like `gtfs_date`? (in all instances of this PR, not just here)

backend/save_routes.py
```diff
@@ -1,7 +1,11 @@
-from models import gtfs, config
+from models import gtfs, config, util
 from compute_stats import compute_stats_for_dates
 import argparse
-from datetime import date
+from datetime import date, datetime, timedelta
+import requests
+from pathlib import Path
+from secrets import transitfeeds_api_key # you may have to create this
```
EddyIonescu, line 7: Not needed in this PR

backend/save_routes.py
```diff
@@ -34,39 +38,85 @@
 #
 # Currently the script just overwrites the one S3 path, but this process could be extended in the future to
 # store different paths for different dates, to allow fetching historical data for route configurations.
-#
+# UPDATE: We are now saving some older routes in versioned directories in metrics-mvp/backend/data
```
EddyIonescu, line 41: Adding it like this is pretty confusing, it's okay to edit the script description.
For example, something like
`When no date is provided the script just overwrites the one S3 path representing the recentmost GTFS that an agency has made available that is active. Providing the date adds _YYYY-MM-DD to the routes path, which would allow the backend to use versioned route files.`

backend/models/gtfs.py
```diff
@@ -1058,7 +1066,7 @@ def get_sort_key(route_data):
             return route_data['title']
         return sorted(routes_data, key=get_sort_key)
 
-    def save_routes(self, save_to_s3, d):
+    def save_routes(self, save_to_s3, d, version_date=None):
         agency = self.agency
```
EddyIonescu, line 1070: The gtfs_date should be d. Notice that we want the gtfs_date to be d in line 1073, otherwise it won't select any routes from the GTFS.

backend/save_routes.py
```diff
@@ -34,39 +38,85 @@
...
+	
     for agency in agencies:
-        scraper = gtfs.GtfsScraper(agency)
-        scraper.save_routes(save_to_s3, d)
+	
+        if gtfs_date is None:
+            # save the normal way, downloading the most recent GTFS file
+            # should probably not be using both date_to_use and gtfs_date_to_use
+            date_to_use=date.today()
+            gtfs_date_to_use=date.today()
```
EddyIonescu, line 70: Why is date_to_use needed? Would we be fine with just gtfs_date? (without any to_use, I think it's fine to change gtfs_date to the recentmost qualified date below)

backend/save_routes.py
```diff
@@ -34,39 +38,85 @@
...
+                        recentmost_date_qualified_date = candidate_date
+                        recentmost_date_qualified_zip_file = candidate_zip_file
 
+            gtfs_date_to_use = recentmost_date_qualified_date
+            gtfs_path = recentmost_date_qualified_zip_file
+            gtfs_path = f'{util.get_data_dir()}/{recentmost_date_qualified_zip_file}'
+
+        # save the routes
+        scraper = gtfs.GtfsScraper(agency, gtfs_path=gtfs_path)		
+        scraper.save_routes(save_to_s3, date_to_use, version_date=gtfs_date_to_use)
```
EddyIonescu, line 109: We should only have to pass in one date to save_routes no matter what?

backend/save_routes.py
Brian-Lee: Good idea. Done.

Brian-Lee:
> Hey @Brian-Lee I'm really sorry for the delay in reviewing this, I'll be faster in the follow-ups. Getting closer, I've added some comments. Request a review from me in the PR whenever you're ready. Thanks!

No worries! Thanks for helping me!

backend/models/gtfs.py
Brian-Lee: Apparently not. Removed now.

backend/models/routeconfig.py
Brian-Lee: That's better. I made that change in gtfs.py and routeconfig.py.

backend/save_routes.py
Brian-Lee: Looks like we need datetime, but I removed the imports for requests, pathlib and secrets.

backend/save_routes.py
Brian-Lee: Much better. I made that change.

backend/models/gtfs.py
Brian-Lee: TY. I made that change.

backend/save_routes.py
Brian-Lee: Right. That's been fixed.

backend/save_routes.py
Brian-Lee: Right. I've simplified that.

backend/save_routes.py
```diff
@@ -32,41 +33,85 @@
...
+                    candidate_date = datetime.strptime(candidate_date_string,"%Y-%m-%d").date()
+                    if candidate_date - gtfs_date <= smallest_timedelta_so_far and candidate_date <= gtfs_date:
+                        recentmost_date_qualified_date = candidate_date
+                        recentmost_date_qualified_zip_file = candidate_zip_file
+
+            gtfs_date = recentmost_date_qualified_date
+            gtfs_path = f'{util.get_data_dir()}/{recentmost_date_qualified_zip_file}'
+
+        # save the routes
+        scraper = gtfs.GtfsScraper(agency, gtfs_path=gtfs_path)
```
EddyIonescu, line 103: So an issue here is that if no qualified GTFS is found the script will use the path "", so it'll throw an error like this:
`IsADirectoryError: [Errno 21] Is a directory: '/Users/eddy/Documents/metrics-mvp/backend/data/'`
Could an error message saying something like "an active GTFS for this date was not found"?

Also, lines 74-97 look like they could be in their own function, where you return `candidate_date` and `candidate_zip_path` when they're found, or at the end of the function throw an exception with a descriptive error message.

backend/save_routes.py
```diff
@@ -32,41 +33,85 @@
 #}
 #
 #
-# Currently the script just overwrites the one S3 path, but this process could be extended in the future to
-# store different paths for different dates, to allow fetching historical data for route configurations.
-#
+# When no date is provided the script just overwrites the one S3 path 
+# representing the recentmost GTFS that an agency has made available that # is active. Providing the date adds _YYYY-MM-DD to the routes path, 
+# which would allow the backend to use versioned route files.
```
EddyIonescu, line 38: The header says _YYYY-MM-DD but you implemented -YYYY-MM-DD, make sure they're consistent.

backend/models/gtfs.py
```diff
@@ -108,15 +117,14 @@ def contains_excluded_stop(shape_stop_ids, excluded_stop_ids):
...
 
 class GtfsScraper:
-    def __init__(self, agency: config.Agency):
+    def __init__(self, agency: config.Agency, gtfs_path=None):
         self.agency = agency
         self.agency_id = agency_id = agency.id
         gtfs_cache_dir = f'{util.get_data_dir()}/gtfs-{agency_id}'
 
-        download_gtfs_data(agency, gtfs_cache_dir)
+        get_gtfs_data(agency, gtfs_cache_dir, gtfs_path=gtfs_path)
```
EddyIonescu, line 125: For say trimet, the GTFS directory ends up being:
/data/gtfs-trimet/gtfs-trimet-2020-02-22 instead of /data/gtfs-trimet_2020-02-22 (for a sample date).

Then, in line 127 the GTFS that's passed in ends up being the one in gtfs_cache_dir, so gtfs_path ends up not being used.

backend/save_routes.py
Brian-Lee: Thank you for pointing out the lack of error msg. I added one and moved code into a new function called get_recentmost_date_qualified_gtfs_path(gtfs_date)

backend/save_routes.py
Brian-Lee: Nice catch. It is now consistent as -YYYY-MM-DD

backend/models/gtfs.py
Brian-Lee: I added an if-else which should solve that. I think there are cleaner ways, and I haven't yet done a good job of testing this fix.

Brian-Lee: I am going to re-request a review at this point. Thank you so much for looking through this over and over again!

Brian-Lee:
> Just wondering, could you, in the PR description, write out a list of what you did to test your PR? A section on Tests is part of the PR template (not completely sure), so that's missing.
> 
> It should look something like this:
> 
> Tests:
> Command: python3 save_routes.py --agency trimet
> ...
> Command: python3 save_routes.py --agency trimet --date 2000-02-02
> (should give error as we don't have a GTFS from then)
> ...
> Command: python3 save_routes.py --agency trimet --date 2020-02-22
> ... (saves route_v3a_trimet_2020-02-22...)

Great suggestion. I put in a few hypothetical test cases. The actual results depend on the date and the contents of your directory.

backend/models/gtfs.py
hathix: Brian, can you formally re-request a review? This PR still looks like it's in the "review required" phase.

hathix:
Eddy, does this look good at this point?

Issue #647, opened 2020-06-05
Metrics for days after 5/29/20 aren't loading


hathix:
https://muni.opentransit.city/metrics?firstDateRange%5BstartDate%5D=2020-05-29&firstDateRange%5Bdate%5D=2020-06-01 shows data. https://muni.opentransit.city/metrics?firstDateRange%5BstartDate%5D=2020-05-30&firstDateRange%5Bdate%5D=2020-06-01 doesn't. 

Did we do something that broke data collection on 5/30?

EddyIonescu:
Hey data is still being collected so we haven't lost anything (you can confirm this in the old OpenTransit map here: http://opentransit-map.s3-website-us-west-2.amazonaws.com).

What happened is that Muni hasn't updated their GTFS (their recentmost one ended on May 29) so the API's returning null because of missing date keys (I think, I can look more into it this weekend, @youngj would know more about this). The arrivals are still getting computed with the old GTFS route configs and are in S3.

hathix:
I see... can we reuse the old GTFS (pre-5/29) if Muni hasn't given a new one? Or is the data useless unless we have the freshest GTFS?

EddyIonescu:
@hathix yep that's what https://github.com/trynmaps/metrics-mvp/pull/648 does, it is the freshest GTFS except that it's expired which is why the pre-computing scripts failed.

There could be a note (like a banner or something) saying that all schedule-based metrics (ie. scheduled times, scheduled headways, on-time performance, scheduled runtimes) for Muni are inaccurate since April as they've been running entirely on a headway-based system, so would be worth making into a ticket. Still it's the runtime data that's the most interesting - like making a more detailed and granular version of this graphic Muni made a few weeks back: https://www.masstransitmag.com/bus/press-release/21140228/san-francisco-municipal-transportation-agency-sfmta-shelterinplace-order-allows-sfmta-to-analyze-sources-of-delays

hathix:
Great! I'll look at that PR. I agree that we should give some notification that recent stats are off, I can make a PR to let us put notifications on screen

PR #648, opened 2020-06-11
Use recentmost date key if date is not active in GTFS

EddyIonescu:
Resolves https://github.com/trynmaps/metrics-mvp/issues/647

As to identify the GTFS date corresponding to a given
date (ie. the GTFS date containing the scheduled trips
for that date) a date_keys lookup is done.

The lookup is done on the date_keys file, which persists
the GTFS date for each date that was processed, as to allow
for re-running pre-computation on older GTFS's.

Since the recentmost GTFS file Muni put out expired on 2020-05-29,
the pre-computation as well as the API failed due to the date key
lookup failing.

This PR modifies the get_date_key function as to return the
recentmost date key corresponding to the same day of the week.
Note that it does not take holidays into account. It raises
an exception if it cannot find a suitable date_key to use.

It prints a warning if it succeeds in using another date_key, although a
future PR could send out an email so that we are aware the GTFS is outdated,
which the save_routes batch should be doing too.

### DIFF
--- a/backend/models/timetables.py
+++ b/backend/models/timetables.py
@@ -308,11 +308,33 @@ def get_s3_path(agency_id, route_id, date_key, version=DefaultVersion):
 
 def get_date_key(agency_id, d: date, version = DefaultVersion):
     date_keys = get_date_keys(agency_id, version)
+    date_key = date_keys.get(str(d), None)
+    if date_key is not None:
+        return date_key
 
-    try:
-        return date_keys[str(d)]
-    except KeyError as ex:
-        raise KeyError(f"Date key not found for {d} on {agency_id}")
+    # Return the recentmost date key that is on the same day of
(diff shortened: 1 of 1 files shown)


hathix:
How does the frontend find out if the get_date_key function still can't find a suitable date_key? Is it possible for that function to return `null` or something so the frontend can raise an error to the user instead of just showing an empty screen? Or is there another way to check this?

EddyIonescu:
@hathix so the frontend doesn't find out it just gets empty fields so it shows `-` for all the dashboard cells. It's the same behaviour as when you enter a date that we don't have data for (like say in 2015). If the fields are null the frontend could say something like "Sorry we don't have any data for this date" but that'd be a separate issue/PR.

hathix:
Got it, we can do that in a separate pr quite easily.

